PR #4, opened 2026-07-06
feat(webui): align model-profile rows and move activated profile to top

Grace-xyx:
- webui/{app.js,style.css}: render each configured-models row as a 4-column CSS grid (info | set-active | edit | delete) so the action buttons line up across rows; the active row inserts a spacer to preserve column alignment. Scoped to a new .prof-row class so the shared .cust-row flex layout used by other panels is untouched.
- gateway.py: POST /model-profiles/<id>/activate now moves the profile to the top of the list (others shift down); idempotent if already #1.
- CLAUDE.md: document webtools.py (keyless web_search/web_fetch) next to the egress fence and permissions module.


Nobody-Zhang:
Superseded by #6, which consolidates this UI work with #5 and preserves the original Grace-xyx authored commit in the combined branch history.

PR #5, opened 2026-07-06
fix(config): placeholder API key from .env.example falsely reported a…

Grace-xyx:
…s configured

setup.sh copies .env.example to .env, leaving OPENAI4S_LLM_API_KEY=your-api-key-here. On first launch that stub is seeded into every built-in Ark model profile and persisted to SQLite, so Customize>Models shows a configured key for all of them but requests 401. The stub was never recognized anywhere in the chain.

- config.py: add is_placeholder_api_key() + a placeholder set; LLMConfig.__post_init__ drops them so cfg.llm.api_key never carries a stub into has_api_key, profile seeding, or effective_api_key. test-key (offline suite) is deliberately preserved.
- gateway.py: _mask_profile rejects placeholders too, so the UI never reports "configured" for a stub regardless of where the value came from.

### DIFF
--- a/openai4s/config.py
+++ b/openai4s/config.py
@@ -47,6 +47,23 @@ def _load_dotenv() -> None:
 _load_dotenv()
 
 
+# Obvious template stubs copied verbatim from .env.example — never a real key.
+# Filtering these means cfg.llm.api_key (and everything derived from it:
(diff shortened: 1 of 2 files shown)


Nobody-Zhang:
Superseded by #6, which consolidates this config fix with #4 and preserves the original Grace-xyx authored commit in the combined branch history.

PR #40, opened 2026-07-20
Trust Foundation (0.1.x Trust Patch): P0 correctness, secrets, durability, contract v1

Nobody-Zhang:
Implements Phase 1 ("Trust Foundation") of `OpenAI4S-nextgen-consolidated-improvement-proposal.md` — what the proposal itself calls the **0.1.x Trust Patch** (§10): close the confirmed false-success, path/archive, secret, implicit-install, browser, and transport defects before anything else.

**All 10 P0 items, all 4 `known_bug` characterizations, and contract v1 are done.** Tests: 1595 → 1955.

## ⚠️ Read first: five behaviour changes that need a human decision

Each reverses something an existing test documented as intentional. I believe each is right, and in every case the repo had already recorded that it wanted the opposite — but a semantic change is yours to sign off, not mine.

| Commit | Reverses | Who could be affected |
|---|---|---|
| `957a394` | **Secrets `auto` no longer degrades to plaintext — it fails closed** | **A Linux server with no keychain and no `OPENAI4S_SECRET_*` env will now refuse to handle credentials.** This is the frozen product decision ("both, and neither may degrade"), but it *is* a breaking deployment change. |
| `9652c40` | A disabled MCP connector is zero-spawn; `host.mcp.tools()` no longer probes it | Flows relying on "disabled but still introspectable" |
| `8395932` | Malformed JSON body: `{}`+200 → explicit 400 | Clients relying on a bad body being tolerated |
| `8395932` | Only *leading* system messages become the initial system field | Anything asserting on Anthropic/Gemini payload shape |
| `36c1249` | `scp` local paths are confined to the workspace | Uses that read/write outside it |

For the first three, the golden trace already carried `known_bug: true` with the new behaviour as its stated `desired_contract`, and the proposal names the fix directly.

## What's in it

**Correctness.** The headline defect: **no ssh job had a reachable `failed` state**. `&` binds looser than `&&`, so `mkdir && cd && cat > run.sh && nohup … &` made the *whole* and-list async, and POSIX gives an async list `/dev/null` for stdin — `cat` read nothing, `run.sh` was written empty, `bash run.sh` exited 0, and a hardcoded `exit_code: 0` reported it as done. Every ssh job "succeeded" without executing. No mocked test could find it (they asserted the command *string*), so the regression runs a real shell; against the old code a `sleep 30` job returns `done` instantly.

**Secrets.** Credentials move behind opaque references (`secret://v1/...`) in the system keychain or, on servers, the process environment — covering `llm_api_key`, `tavily_api_key`, per-profile keys, and connector env. No obfuscated-file backend, and a test argues against adding one.

**Durability.** Remote jobs survive a daemon restart, with the provider receipt as evidence; `reconcile()` reports and deliberately never resubmits.

**Schema.** `PRAGMA user_version` + transactional migrations: a database is fully at version N or fully at N-1, never between.

Also: strict CSP with no `unsafe-inline` in `script-src`, compute rows built from DOM nodes, startup no longer installs packages, typed transport with bounded `Retry-After` retries, redacted diagnostic bundle, and standalone `openai4s verify-package`.

## How it was verified

Tests are the floor. Each fix was also driven end-to-end: real `bash` for the shell semantics, a real browser for CSP and XSS (hostile GPU names and markdown produce **zero** `` nodes and zero CSP violations — the markup is never generated, not merely blocked), the real macOS keychain for credential round-trips, a real cross-process restart for job recovery (`kill -0` on the recovered receipt), and real HTTP for the canary checks.

That caught four cases where **my own tests were passing for the wrong reason** — a backup-restore masking whether the transaction worked, the NUL guard being unreachable through a real tar, a comment asserting behaviour Python removed in 3.6. Each fix was then re-run against the old code to confirm it actually fails.

One fix I **completed and then reverted**: budgeting oversized observations with an absolute spill path made the golden trace depend on the data-dir path length and leaked `$HOME` into the model's context. The correct form (fixed-width workspace-relative ref) landed later in `d53d6a2`.

## Note on the diff

7 of the 33 commits shown are **not mine** — they come from your local `main`, which has an unpushed `pr36tmp` merge (BioProBench + conda discovery). They'll drop out of this diff once `main` is pushed. The branch is rebased on that local `main`, so the skill count is set to **34** (both new skills), derived from the loader rather than arithmetic.

## Not included

Phases 2–4 — Connector Hub, the 10-workflow/20-case benchmark, and the signed release chain — depend on §8 decisions that are still open (connector licensing and canary budget, claim/evidence semantics). Four §8 decisions were frozen during this work and are implemented; the release pipeline was explicitly deferred.


Nobody-Zhang:
Reopened as a correctly-named branch — `claude/…` violates the branch-naming policy in CONTRIBUTING.md (needs a feat|fix|docs|… prefix). Same commits, same content.

PR #57, opened 2026-07-30
fix(security): survive squash-rewritten fingerprints without widening a path rule

Nobody-Zhang:
**Draft — this is step 1 of a deliberate two-step verification. Do not merge yet.**

## The problem

`Scan Git history` fails on `main` itself. The four findings are already-triaged synthetic fixtures, and `.gitleaksignore` already contains their exact `<file>:<rule>:<line>` triples — but with a **stale commit SHA**:

```
.gitleaksignore : 8d715ebe…:tests/test_model_profile_readiness.py:generic-api-key:80
CI reports      : f2d8adb1…:tests/test_model_profile_readiness.py:generic-api-key:80
```

`protect-main` permits only squash and rebase merges, so every branch that adds a suppression has its commits rewritten on the way into `main`, and the fingerprint that was correct on the branch names a commit that no longer exists. When #52 landed as a squash, four suppressions silently expired and the scan went red on a tree nobody had changed. This is structural: **fingerprint suppression is incompatible with squash-only merging**, and it will recur on every future suppression.

## Why not a path allowlist

The obvious fix — allowlist `tests/` — is one this repository has already argued against, in writing. From `tests/test_governance.py`:

> Suppressed by fingerprint rather than by widening a path rule, so a real leak in the same file is still caught, and pinned by count here so a new suppression cannot be added without a reviewer seeing it.

and from `tests/test_retrieval_source.py`, on why the canary is not exempted by being in a test file:

> a scanner that made an exception for test files would be a scanner with a hole exactly where people paste real keys "just to check".

So this PR does not widen a path. It moves the suppression from *"this finding, in this commit"* to *"this exact string, which is not a secret"* — `regexTarget = "secret"` with anchored patterns. That survives any rewrite of history, and gives up strictly less than a path rule: a real credential in these same files still fails the scan.

## Why this is a draft, and what step 2 will be

A `.gitleaks.toml` that gets `[extend] useDefault` wrong **replaces** the default rule set instead of extending it. The scan then finds nothing and exits 0 — which is indistinguishable from a scan that works. Landing the config and the allowlist in one commit would mean publishing "the secret scanner is green" with no evidence the scanner still has any rules. That is the failure mode this repo calls *wrong rather than absent*.

So:

- **Step 1 (this commit)** — the config lands with `[extend] useDefault = true` and **no allowlist at all**. The Gitleaks job on this PR must still report `leaks found: 4`. That is the evidence the default rule set is loaded.
- **Step 2 (next commit)** — add the allowlist, drop the four now-dead fingerprints from `.gitleaksignore`, and update the count assertion and comment in `tests/test_governance.py`. The drop from 4 to 0 is then attributable to the allowlist rather than to a neutered scanner.

Both run logs stay in this PR's history as the argument.

`--config` is also now passed explicitly rather than relying on gitleaks' auto-discovery, so a renamed or missing config is a loud error instead of a silent fallback to defaults.

## Not a real leak

For the record, both flagged values are synthetic: `abc123def456ghi789` is an obvious placeholder in a model-profile dict, and `Zx9Qw3Er7Ty1Ui5Op2As6Df4Gh8Jk0Lm` is a redaction-test canary that has to be high-entropy (the redaction under test keys on entropy) while not being provider-shaped (`scripts/source_secret_scan.py` correctly refuses `sk-…`).


Nobody-Zhang:
Two-step verification complete — this is now ready to review.

| step | commit | commits scanned | result | run |
|---|---|---|---|---|
| 1 — config only, **no allowlist** | `ed77691` | 535 | `leaks found: 4` (failure) | [30568183935](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/30568183935) |
| 2 — allowlist added | `190b581` | 536 | `no leaks found` (success) | [30568401974](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/30568401974) |

Step 1 is the load-bearing one. It reports the same four findings `main` reports today, which is the evidence that `[extend] useDefault = true` extends the default rule set rather than replacing it. Without that run, a green step 2 would be equally consistent with a scanner that simply has no rules left — and "the secret scanner is green" is exactly the kind of claim that gets believed rather than checked.

`tests/test_governance.py` now asserts that failure mode directly: `useDefault` present, the permitted values exactly those two, and no `paths`/`files`/`commits`/`stopwords` key having crept in — because any of those would suppress findings this allowlist has not individually accounted for.

Nobody-Zhang:
**Correcting two things an adversarial review found in my own PR** (pushed as `1dd7714`). Both were in the part a reader would believe rather than the part CI checks, which is the worse half to be wrong in.

### 1. The four dropped `.gitleaksignore` rows were not dead

I claimed — in the PR body, the config header, a commit message and a test docstring — that commits `8d715ebe` and `3dcda11f` "stopped existing when #52 squashed". That is false. `gitleaks git` scans every ref the checkout fetched (`fetch-depth: 0`), and both commits are still reachable on `origin/next`. With the allowlist stripped, the scan reports **8** findings, not 4: the originals under those two SHAs *plus* the copies under `f2d8adb`.

The real mechanism is a better argument than the one I gave. Squashing does not destroy a suppressed commit — it **copies** the finding under a new SHA. The old fingerprint keeps suppressing the old commit and nothing suppresses the copy, so a fingerprint list can only ever grow by one row per merge. Dropping those four rows is still correct, because suppressing by *value* covers original and copy at once, which is exactly what a fingerprint cannot do. But they were live, not dead — they were merely insufficient.

### 2. The pinning test did not pin

`test_gitleaks_config_extends…` matched only `'''^…$'''` on its own line. Adding `".*",` in ordinary double quotes left all three assertions passing while gitleaks permitted everything — confirmed by execution, a genuine `api_key = "sk-live-…"` went from `leaks found: 1` to `no leaks found` under that config. A test whose stated purpose is *"a new suppression cannot be added without a reviewer seeing it"* has to survive the spelling an author would actually reach for.

`_allowlisted_regexes` now extracts string literals in all four TOML quoting styles and the assertion compares the whole list, so an extra entry fails however it is written. Also pinned: exactly one `[[allowlists]]` block, and both patterns anchored at both ends so a permitted value cannot decay into a prefix rule. Verified by injecting that exact bypass and watching the test fail (`Left contains one more item: '.*'`), then restoring.

### Also corrected

The docstring's inventory of the remaining 12 rows was wrong about two of them — `docs/refactor-plan.md` and an `NGC_API_KEY` fixture in `test_compute_nvidia.py` are not redaction canaries. It now accounts for all twelve.

### Known and not fixed here

This fixes `main`. `next` has no `.gitleaks.toml` and still carries the 16-line `.gitleaksignore`, so its scan stays red until this reaches it.

PR #62, opened 2026-07-31
fix(compute): fail closed in the BYOC confinement probe, and describe the boundary that now exists

Nobody-Zhang:
## Summary

`docs/security.md`'s "Remote-compute confinement" row claimed `enforce` refuses `byoc:*` ops "because no host-side boundary exists for the provider helper yet", and that `auto` "runs unconfined". Both stopped being true when `security/byoc_confinement.py` landed. Investigating that divergence surfaced a real fail-open in the helper's own probe.

Two separable commits. A third — a fix for the same `.git` worktree problem in `tests/test_byoc_confinement_scope.py` — was **dropped**, because #56 landed the same fix on `main` first and did it better: it resolves the `gitdir:` pointer outside the boundary and names the real metadata directory, so one uniform assertion covers both layouts instead of branching, and it also handles a worktree whose main repository sits outside `$HOME`. This branch is rebased onto that.

**1. `docs(security,compute)` — describe the boundary that now exists**

`compute/manager.py` wraps the helper in a Seatbelt profile (macOS) or bubblewrap invocation (Linux), and `posture()` reports `state: "active"`, `enforced: true` on a host that can establish one. Verified rather than assumed on macOS 26.x: `available()` → `(True, 'macOS Seatbelt')`; a canary planted in `$HOME` and read from inside a real `wrap()` raises `PermissionError`, as does `os.listdir($HOME)`; `security list-keychains` exits non-zero.

The divergence erred safe — the doc claimed *less* protection than existed — but it misled a reader deciding whether to enable BYOC and contradicted `posture()`'s own `detail` string. Two things the old row did not say and the new one does:

- **the network is NOT isolated** (`network_isolated: false`). Calling a provider's REST API is the helper's whole job, so egress is a separate capability and is not enabled. "The helper is confined" reads to most people as "the helper cannot phone home".
- `enforce` **does** still refuse — but only where no boundary can be established (no `bwrap`/`sandbox-exec`, a failing self-test, an unsupported platform), and on every op rather than only at submit. The old row was right for those cases and merely missing the qualifier.

`docs/compute.md` carried a stronger form of the same claim, and one of its mode bullets was a real behavioural divergence: `off` is no longer "same as `auto` today". `host.compute` keeps its Prototype marking, re-attributed to the reason in `docs/v02-decisions.md` (no named production-grade provider) rather than to a boundary that is no longer unbuilt.

**2. `fix(compute)` — fail closed when the probe verifies nothing**

`_probe_confined()`'s no-anchor netns fallback answered "I could not perform the check" with `True`. `run_oneshot` only calls it when the caller passed `expect_confined`, so that `True` let an unconfined helper go on to read the credential and call the provider having verified nothing. Neither `OSError` was evidence of confinement:

- `/proc/self/ns/net` missing — under the real bubblewrap boundary `/proc` is mounted (`--proc /proc`), so its absence says the process is somewhere *unexpected*, not somewhere confined;
- `/proc/1/ns/net` unreadable — the most reachable way in, since an unprivileged helper whose PID 1 belongs to root routinely cannot stat it.

A malformed `OPENAI4S_HOST_NETNS_INO` also now fails closed instead of raising `ValueError` out of the probe.

**What this does not change:** the macOS branch returns early and never reaches this fallback, and was already fail-closed. No host triggers these paths today, since the manager always supplies the anchor when it wraps. This is about which way the system slides when that stops being true.

`OPENAI4S_HOST_HOME_DEV` is the live anchor and nothing pinned it — the scrub test covered only the legacy netns var, so a `HOST_`-shaped addition to `BASELINE_SECRET_PREFIXES` would have scrubbed the anchor with every assertion still green, dropping the probe into exactly the fallback above. Now pinned.

## Area

- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Documentation
- [x] Security-related change

## Verification

Ran:

```text
uv run pytest tests/test_byoc_confinement.py tests/test_byoc_confinement_scope.py tests/test_compute_nvidia.py
    -> 57 passed, 1 skipped
uv run mypy                        -> Success: no issues found in 6 source files
uv run pre-commit run --all-files  -> all hooks passed
    -> 100 maintained directories, complete bilingual coverage
uv run pytest                      -> see "Not run" below

Manual, on macOS 26.x (measured, not inferred):
  bc.available()        -> (True, 'macOS Seatbelt')
  bc.posture('enforce') -> enforced=True, state='active', network_isolated=False
  canary planted in $HOME, read from inside a real bc.wrap() -> PermissionError
  os.listdir($HOME) from inside                              -> PermissionError
  security list-keychains from inside                        -> exit 1
  mutation test: reverting the three probe returns makes the new test fail with
    "unverifiable must not report confined"

Re-run after rebasing onto main (which upgraded mypy 1.20->2.3, isort 5->8 and
ruff): targeted tests 57 passed / 1 skipped, mypy clean, pre-commit --all-files
clean under the new pins.
```

Not run:

```text
uv run pytest (full offline suite) — completes except for a pre-existing
  environment gap, unrelated to this change:
    tests/test_admet_genetic.py            -> collection error: No module named 'pandas'
    tests/test_catalyst_sar_screening.py x2 -> RuntimeError: need matplotlib
scripts/capture_response_schemas.py --check -> exits 2, blocked by the same
  pandas collection error ("the suite failed; the capture would be incomplete")
Linux / bubblewrap execution of the BYOC boundary
```

Reason:

```text
The local venv holds only the locked lightweight dependencies, without the
`science` extra, so pandas/matplotlib are absent. Those three failures are
present on the base commit and are not touched by this change; CI installs
what the suite needs. The schema gate is blocked by that same collection
error rather than by drift. (Worth noting for whoever runs that gate next: it
also used to be blocked inside any worktree by the `.git` probe that #56 has
now fixed on main.)

No Linux host, container runtime, or VM was available here, so the bubblewrap
half of the row is supported by code reading and branch simulation (mocked
platform + PATH) rather than by execution. The nightly linux-sandbox-smoke
job exercises OPENAI4S_KERNEL_SANDBOX, not OPENAI4S_COMPUTE_CONFINEMENT, so
the Linux BYOC boundary has no automated proof anywhere today. Flagged rather
than papered over; a harness/smoke/linux_byoc.py alongside the existing kernel
smoke would close it.
```

## Risk and Reviewer Focus

**Lowest confidence: the Linux claims in the `docs/security.md` row.** They are the one part not confirmed by execution — see above. Reviewers with a Linux box should sanity-check `bc.available()` and `bc.posture('enforce')` there.

`_resident.py:_probe_confined()` is the only behavioural change, and it moves a security decision from fail-open to fail-closed. The direction is safe, but it means a host that *cannot* verify confinement will now be refused under `enforce` where it previously proceeded. No current host reaches those paths (the manager always supplies the anchor when it wraps), so no behaviour change is expected in practice.

`compute/manager.py` is a hotspot file; the only edit there is a five-line comment correction, no logic.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched — `manager.py` receives a comment-only change.

### Tests

- [x] No new test requires live LLM, network, GPU, SSH, Docker, lab hardware, or secrets.
- [x] No tests were deleted without tracked replacements (`git ls-files --deleted` is clean).
- [x] The probe change ships focused coverage, including a positive control so the new assertions cannot pass on a probe that only ever returns `False`, plus a mutation check that the test fails against the old behaviour.

### Core Dependency Policy

- [x] Optional science imports are guarded — none added; `_resident.py` remains stdlib-only, as the sandboxed provider runtime requires.
- [x] No new dependencies.

- [x] No unpublished research roadmap, internal assignment, benchmark detail, or unreleased result is disclosed.
- [x] Security-sensitive changes include verification notes — the canary probe and mutation test above. The `$HOME` canary was synthetic and removed afterwards; the keychain check reads no secret, only whether securityd answers.
- [x] Docs do not overstate isolation, sandboxing, privacy, or security guarantees — this PR exists to correct an understatement, and adds the explicit `network_isolated: false` caveat plus the Linux-unproven note above.

### Docs

- [x] User-facing behavior changes are reflected in docs.
- [x] `README.md` / `README_zh.md` sync — not affected. `docs/security.md` and `docs/compute.md` have no per-file `_zh` siblings (only the `docs/README` pair is bilingual, and it already lists both files), so no translation is required; `check_directory_readmes.py` passes.

---


Nobody-Zhang:
Reviewed by a parallel agent team (posting as a comment — GitHub refuses self-approval, so this merges via the admin bypass like the rest of today's self-authored PRs).

The core claim was verified by **execution**, not by reading: `ByocResident._probe_confined` was exported at the merge base and at the head and driven through four scenarios.

| scenario | before | after |
|---|---|---|
| no `/proc/self/ns/net` | `True` | `False` |
| self readable, PID 1 not | `True` | `False` |
| malformed `OPENAI4S_HOST_NETNS_INO` | raises `ValueError` | `False` |
| positive control (mine ≠ anchor) | `True` | `True` |

So it genuinely fails closed on both `OSError` paths and on the malformed anchor, and the 80 new test lines are capable of failing — each assertion inverts against the pre-fix code, and the `ValueError` case would previously have errored rather than mis-asserted.

Also checked, since this is the one module where ordering is load-bearing: **the two-stage credential scrub is untouched.** `run_oneshot` still does `_prologue()` (re-scrubbing with the provider's declared `secret_env_prefixes`) → probe → `read_auth(fd=0)` at `_resident.py:114-120`; nothing is moved or late-bound relative to the credential read. Confirmed separately that no bundled provider declares a prefix matching the new `tests/test_compute_nvidia.py` anchors, so that pin is real coverage of a real gap rather than a restatement.

Merge is mechanically clean against current main: `git merge-tree --write-tree origin/main <head>` produces a tree with zero conflict output, even though main has since gained #58's 53-line `byoc_confinement.py` change and its rewrite of `tests/test_byoc_confinement_scope.py` — this PR touches neither.

### Non-blocking, worth a follow-up

- **`docs/compute.md` and `docs/security.md:34` say the boundary "replaces \`\$HOME\`" on macOS. It does not.** `byoc_confinement.py:191` emits `(deny file-read-data (subpath \$HOME))`, and the comment two lines above says so on purpose — *"metadata is what the loader needs; contents are what a credential is"*. `os.stat('~/.ssh/id_rsa')` still succeeds. "Replaced" is the Linux `--tmpfs \$HOME` mechanism's language applied to a platform that does something narrower, and overstating isolation is the one thing CONTRIBUTING singles out. Suggested: "\`\$HOME\` is replaced (Linux) / its contents are unreadable (macOS)".
- The security.md row says the helper "exits 71 without acting" — true only under `enforce`. Under `auto`, the default named in the same row, `manager.py:1774-1797` audits `compute_confinement_degraded` and re-spawns the op **unconfined**. Pre-existing and audited, but the row reads as unconditional.
- The netns fallback is unreachable from any host that has ever existed: `git log -S OPENAI4S_HOST_NETNS_INO --all -- openai4s/` returns nothing. Defensive code is fine; the "a host a release behind" framing implies a compatibility obligation with no referent.

### Ordering note

#65 (`fix/byoc-home-read-class`) is currently `DIRTY` and also touches `tests/test_byoc_confinement.py`, which this PR adds 80 lines to. Landing this first is the cheaper order; #65 needs a rebase either way.

PR #65, opened 2026-07-31
fix(security): deny the whole read class over $HOME in the BYOC profile

Nobody-Zhang:
## Summary

The BYOC helper's Seatbelt profile denied `file-read-data` over `$HOME`, so a
`stat()` on a guessed path still answered: `~/.ssh/id_ed25519`,
`~/.aws/credentials` and friends returned size, mtime, mode and owner to a
provider shim that has the network by design.

The in-source comment held that `file-read*` could not be used, because denying
the whole read class also denies the metadata reads `execvp`/dyld perform on the
interpreter — `sandbox-exec` then dies before the helper starts. That symptom is
real and reproduces on demand. The conclusion drawn from it was wrong: the loader
needs metadata on the *specific components it walks*, and those can be enumerated.

`traversal_metadata_paths()` walks the route the way the loader does — emit every
component, and at each component that is a symlink, rewrite the remainder onto the
target and walk that too. The load-bearing entries are the **unresolved** hops a
realpath'd allowance cannot name: a uv-managed venv reaches its interpreter through
`~/.local/share/uv/python/cpython-3.13-macos-aarch64-none`, an alias symlink that
appears nowhere in `sys.base_prefix`. Allowing `file-read*` on the resolved prefix
alone is **not** enough — measured separately from allowing nothing at all.

The walk is seeded from every allowed read root, not just `sys.executable`. On a
system-`python3` layout the prefix is outside `$HOME` entirely and the only
components under it lead to the helper's own package; seeding from the interpreter
alone produces a profile that fails to start on exactly that layout.

### Second defect, fixed here because it is the check that should have caught the first

The macOS self-test probed with `/bin/sh`. The shell lives outside `$HOME` and
starts under any profile this module can emit, so the probe stayed green on hosts
where the **interpreter** could not start — which is the single failure mode of a
metadata enumeration. `auto` would have reported an active boundary and the op
would then have died at exec. It now probes with `sys.executable`, and an execvp
denial is reported as "the metadata allowances do not cover every path component
the loader walks" rather than as "no boundary", which sends a reader to the wrong
place entirely.

### What this does not buy, stated because the docstring would otherwise imply it

**Existence is not closed.** Seatbelt answers a denied path `EPERM` and a missing
one `ENOENT`, so a shim can still distinguish "something is here I may not read"
from "nothing is here". SBPL cannot make a deny lie.

**Attribute *names* stay readable on the components the traversal allowance names.**
`listxattr` on a route component succeeds; `listxattr` on any other path under
`$HOME` is `EPERM`. Attribute **values** are closed everywhere under `$HOME`,
route components included — `getxattr` is `EPERM` throughout. The split is
Seatbelt's: names follow `file-read-metadata`, values follow the read class. Both
facts are measured and written into the source rather than left to be assumed away.

### Reconciliation with #58, which landed while this was in review

#58 is where this change was invited: its comment block says a `file-read*` denial
"does start, and closes `stat()` as well, once every ancestor path component of the
interpreter carries an explicit `(allow file-read-metadata (literal …))`", and that
lifting it "belongs in a change that can be tested across venv/uv/Homebrew/conda/
system layouts rather than assumed from one." This is that change, rebased onto it.

Two consequences, both resolved here:

**The dedicated `(deny file-read-xattr (subpath $HOME))` from #58 is dropped.**
`file-read*` is the whole read class and subsumes it. That is *measured with a
control*, not argued: building this profile with and without the line and probing a
file that carries its own bytes in an xattr gives byte-equal verdicts — `getxattr`
`EPERM` either way, including on the traversal components this PR names.

I first resolved the conflict the other way, on the theory that the metadata
literals re-open xattr on the components they name. The control refuted it. Keeping
the line would have been the safer-looking choice and the more misleading one: what
it does *not* close is `listxattr`, which is governed by `file-read-metadata` and so
follows the traversal allowance — an xattr deny placed after that allowance does not
change it either. A line that closes nothing, sitting beside a residual it appears to
close, is worse than no line. The restore condition (narrowing the denial back to
`file-read-data`) is recorded in the source.

The residual narrows rather than grows: attribute *names* were readable across the
whole home while metadata was, and are now readable only on the interpreter's route.

**`test_metadata_is_readable_under_home_on_purpose_and_contents_are_not` is flipped.**
That test arrived with #58 and deliberately pinned the old trade-off
(`stat_file == "ALLOWED"`). It now asserts `EPERM` for `stat_file`/`stat_dir`, and its
docstring says the trade-off was lifted rather than that it is intended. #58's other
assertions — `xattr_value == "EPERM"`, `xattr_names` tolerant of either answer — pass
unchanged, because they assert behaviour rather than profile text.

## Area

- [ ] Web UI / frontend
- [ ] Server / gateway / API
- [x] Tests / harness / CI
- [ ] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Test / harness
- [ ] Documentation
- [x] Security-related change
- [ ] Release / packaging

## Verification

Ran:

```text
uv run pytest                      # full offline suite, 0 failures
uv run mypy                        # clean
uv run pre-commit run --all-files  # clean, incl. the new ruff pin from #55
uv run pytest tests/test_byoc_confinement.py tests/test_byoc_confinement_scope.py

Re-verified after rebasing onto #58 (0276ca5): full suite 0 failures, lint clean,
and the layout matrix re-run on the new base.

Control for the dropped xattr line: same profile built with and without
`(deny file-read-xattr ...)`, probing a file whose bytes live in an xattr and a
traversal component carrying one. Verdicts byte-identical; getxattr EPERM both ways.

Ad-hoc profile matrix (macOS 26.5.2, arm64) across five interpreter layouts:
  - uv-managed venv (symlinked, via the version-alias hop)
  - python -m venv off Homebrew python3.14 (base prefix outside $HOME)
  - conda base install under $HOME
  - conda env under it
  - system /usr/bin/python3 (prefix outside $HOME entirely)
For each: the helper starts and imports its own package; listdir($HOME) EPERM;
stat(~/.zshrc) EPERM where it previously returned a stat result; listxattr on a
non-route file EPERM (ALLOWED on a route component, by the split described above);
TLS/ssl unaffected. Baselines C1/C2 from the report reproduced exactly.
```

Not run:

```text
A second macOS release.
Linux (this change is macOS/Seatbelt-only; build_bwrap_argv is untouched).
```

Reason:

```text
No second Mac available. This is the gap in the evidence and it is named in the
source comment rather than papered over. dyld's lookup behaviour is exactly the
kind of thing that can move between macOS majors, so the enumeration deserves a
run on Sonoma or Sequoia before this is relied on.

CI cannot substitute: CI is Linux and skips every Seatbelt test in this PR, so a
green CI here is not evidence about the profile change.
```

## Risk and Reviewer Focus

**The failure mode is asymmetric.** An over-broad enumeration weakens the
boundary; an *incomplete* one stops the helper starting on a user's machine,
which is worse than the disclosure being closed. That is why the self-test change
travels with it — it converts that failure from a mysterious mid-op `execvp`
death into a named, visible self-test failure that `auto` degrades on and
`enforce` refuses on.

Look first at:

- `traversal_metadata_paths()` — the walk, its termination bound, and whether the
  seed set covers every layout you can think of. Counter-examples are the most
  valuable review output here.
- `build_profile()` — `literal` vs `subpath` on the metadata block. A `subpath`
  where a `literal` belongs hands back everything beneath `$HOME`'s directories,
  which is the disclosure the deny exists to end.
- `_probe_argv()` — that the probe now execs the real interpreter.

**Overlaps with `fix/byoc-confinement-probe-and-docs`.** That branch fixes the
*helper-side* `_probe_confined()` fail-open in the Linux netns fallback; this one
fixes the *host-side* `self_test()` fail-open on macOS. Different files, different
processes, no contradiction — I trial-merged them and the combined
`test_byoc_confinement*.py` + `test_compute_nvidia.py` run is green. One caution:
that branch's message says "the macOS branch was already fail-closed", which is
true of `_resident._probe_confined` and **not** of the host-side self-test fixed
here. Easy to read as covering both.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched: none of the hotspot files
      are touched by this PR.

### Tests

- [x] No new test requires live LLM, network, GPU, SSH, Docker, lab hardware, or
      secrets in the default offline suite.
- [x] No tests were deleted without tracked replacements. One assertion in
      `test_the_self_test_probes_the_keychain_too` was rewritten to follow the
      probe's mechanism change (shell -> interpreter); the property it pins is
      unchanged and the reason is recorded inline.
- [x] Kernel / host-RPC / gateway changes include focused contract coverage or a
      documented smoke test. (No kernel/gateway change here; the boundary tests
      execute a real `sandbox-exec` rather than asserting on the profile string.)

### Core Dependency Policy

      stdlib web server.
      in-tree use site. (None added.)
- [x] New dependencies, if any, are documented and justified. (None.)

- [x] No unpublished research roadmap, internal assignment, benchmark detail, or
      unreleased experimental result is disclosed.
      verification notes.
      guarantees. The two limits this change does **not** close (existence via
      `EPERM`/`ENOENT`, and xattr on the named traversal components) are stated
      explicitly in the module docstring and the profile comment.

### Docs

      appropriate. (Module docstring section "Why an unreadable home is not enough
      on macOS" rewritten; no user-facing surface changed.)
      affected. (Not affected.)

### DIFF
--- a/openai4s/security/byoc_confinement.py
+++ b/openai4s/security/byoc_confinement.py
@@ -289,7 +289,11 @@ def build_profile(
 
     Order is load-bearing — SBPL takes the *last* matching rule — so the home
     denial comes after `allow default`, and the runtime read allowances come
-    after the denial.
+    after the denial. There are two of those, and they are not interchangeable:
+    `file-read*` subpaths for the trees the helper reads, and
+    `file-read-metadata` literals for the components the loader walks to reach
+    them. A subpath where a literal belongs would hand back everything beneath
+    `$HOME`'s directories, which is the disclosure the denial exists to end.
     """
     home_dir = _canonical(home if home is not None else Path.home())
     stage_dir = _canonical(stage)
--- a/openai4s/security/byoc_confinement.py
+++ b/openai4s/security/byoc_confinement.py
@@ -56,15 +56,22 @@ recurring lesson of this module -- a check that answers an easier question than
 the one it stands in for -- and it is why the host-side test asserts xattrs
 separately from data.
 
-That hole was first closed by a dedicated `(deny file-read-xattr …)` beside the
-data denial. The whole-class `file-read*` denial above subsumes it, so the
-dedicated line is gone and `getxattr` stays `EPERM` -- confirmed by building the
-profile with and without it and getting byte-equal verdicts. What neither form
-closes is `listxattr`: attribute *names* are governed by `file-read-metadata`,
-so they follow the traversal allowance and remain readable on the interpreter's
-own route. That residual is narrower than it was -- names were readable across
-the whole home while metadata was -- and it is stated here because a line that
-looks like it closes something is worse than no line at all.
+That hole was closed by a dedicated `(deny file-read-xattr …)` beside the data
+denial, and it stays. The whole-class `file-read*` denial above does subsume it
+*at that line* -- but the profile ends with `(allow file-read* …)` over the
+interpreter prefixes, the helper package and the stage, and a later whole-class
+allow re-opens `getxattr` on everything it names. An operation-specific deny is
+not re-opened by it. Measured, because an earlier revision of this change
+dropped the line on the strength of a control that probed a file directly under
+`$HOME` -- where it genuinely is a no-op -- and so answered an easier question
+than the one it stood in for: inside an allow-listed read root, without the line
+`getxattr` returns the bytes, with it `EPERM`.
+
+What neither form closes is `listxattr`: attribute *names* are governed by
+`file-read-metadata`, so they follow the traversal allowance and remain readable
+on the interpreter's own route. That residual is narrower than it was -- names
+were readable across the whole home while metadata was -- and it is stated here
+because a line that looks like it closes something is worse than no line at all.
 
 ## Why the profile is shaped the way it is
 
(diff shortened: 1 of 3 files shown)


Nobody-Zhang:
Reviewed by a parallel agent team, and **pushed one fix** (`b62d1c1`) before merging.

## The profile change is right, and does not over-deny — verified by execution

The feared failure of `(deny file-read* (subpath $HOME))` — a venv interpreter under $HOME can no longer start — does **not** occur. Checked across five layouts (uv venv via the alias hop, Xcode `/usr/bin/python3`, Homebrew 3.14, conda under $HOME, `python -m venv` under $HOME) with real `sandbox-exec`: the interpreter starts and imports ctypes/ssl/sqlite3/numpy, while `stat(~/.zshrc)` flips ALLOWED → EPERM, which is precisely the hole this PR exists to close.

The new self-test is load-bearing and works: with `traversal_metadata_paths` stubbed to `[]`, `self_test()` returns False naming the execvp denial, where the old `/bin/sh` probe returns rc 0 on the identical profile — i.e. it was blind. `wrap()` → `available()` → `self_test()`, so `enforce` fails closed and `auto` degrades visibly.

## The one defect: the deleted xattr deny was not redundant

The branch dropped `(deny file-read-xattr (subpath $HOME))`, with an in-source justification that it was *"measured with a control rather than argued … byte-equal verdicts, `getxattr` EPERM either way"*.

It does subsume it **at that line**. But this branch also widened the trailing allowance from `(allow file-read-data` to `(allow file-read*` over the interpreter prefixes, the helper package and the stage. A later whole-class *allow* re-opens `getxattr` on everything it names; an operation-specific *deny* is not re-opened by it. Those trees sit under $HOME on any uv or conda install, and the provider-shim directory among them is user-authored.

Measured on macOS with real `sandbox-exec`, a file carrying `com.apple.review.secret=INSIDE` inside `sys.base_prefix`:

| profile | `getxattr` inside an allowed root | `stat(~/.zshrc)` | interpreter |
|---|---|---|---|
| main | DENIED | ALLOWED ← the hole this PR fixes | 3.13.13 |
| this PR | **ALLOWED:INSIDE** | DENIED | 3.13.13 |
| after `b62d1c1` | **DENIED** | DENIED | 3.13.13 |

The original control probed a file *directly under $HOME*, where the line genuinely is a no-op — so it answered an easier question than the one it stood in for, which is this module's own recurring bug and the subject of its docstring. Honest blast radius: those trees' data forks are readable under both profiles, so what was re-opened is xattr-only content (resource forks, `kMDItemWhereFroms`) on files the helper may already `open()`. Not a credential-path bypass — but a measurable regression against the boundary #58 set, and a false "we measured this" in a security profile outlives whoever wrote it. The comment is corrected rather than deleted.

The tail stays `file-read*`; it needs the whole class for metadata deep inside the prefixes `traversal_metadata_paths` does not enumerate, and narrowing it back would reintroduce the startup failure this branch fixed.

## Tests, both checked against the regression

- `test_an_allowed_read_root_under_home_still_denies_attribute_values` — hermetic (home and allowed root both under `tmp_path`, so it touches neither the real $HOME nor the real interpreter): the data fork reads, `getxattr` is EPERM. Deleting the deny again makes it fail.
- `test_the_profile_denies_the_read_class_over_the_home` gains the string assertion and, more usefully, **loses its `@macos_only`**. `build_profile` has no platform branch — it emits the same SBPL on Linux — and every other assertion about this profile is macOS-gated, so a full revert of the denial used to pass the required Linux jobs with 21 green tests and nothing red. This is the one assertion that can travel, so now it does.

## Non-blocking, left alone

- `tests/test_byoc_confinement_scope.py:504` writes `~/.openai4s-confinement-probe` into the developer's **real** $HOME (conftest redirects `OPENAI4S_DATA_DIR`, never `HOME`) and skips itself if a leftover exists — so a kill between write and `finally` permanently silences this PR's headline test on that machine, as a skip rather than a failure.
- `tests/test_byoc_confinement.py:15-20` still states the conclusion this PR refutes ("`file-read-data` is the correct granularity").
- `test_the_walk_terminates_on_a_symlink_loop` cannot fail by assertion, only by hanging — mutating `_MAX_TRAVERSAL_HOPS` to 1 leaves it green.

## What CI still cannot verify

No CI run ever executes the `@macos_only` tests here: the required pytest jobs are `ubuntu-latest`, and the only macOS job is nightly-gated and runs `harness.smoke.macos_sandbox`, which contains zero occurrences of "byoc". Every SBPL semantic in this PR rests on hand-run macOS execution — mitigated by the new self-test failing closed under `enforce`, and now partly by the un-gated string assertion above.

PR #66, opened 2026-07-31
fix(contract): stop publishing a thread race as the example-session shape

Nobody-Zhang:
## Summary

`docs/response-schemas.json` froze `POST /example/session [ok].error` as
`string`. The field is `str | None`: `gateway.py` builds **one** dict literal for
both verbs and the value is `runner.example_seed.last_error()`. The sibling entry
`GET /example/session [ok]` already said `["null","string"]` — the same payload,
disagreeing with itself one key away.

The cause is a race, not a stale capture. `_ExampleSeedState.start()` holds
`self._lock` while it sets `_last_error = None` and calls `thread.start()`. The
seed thread fails immediately and blocks on that same lock; the handler then calls
`last_error()`, which also wants it. Whoever wins decides whether `error` is `null`
or a string. The suite observes this route's `[ok]` POST **exactly once**, right at
that moment.

So the frozen entry recorded a scheduler outcome and published it as an API
contract. On Linux/CI the thread wins and the gate is green; on macOS the caller
wins (measured 0/600 the other way, even at a 1µs switch interval) and every macOS
developer got a `BREAKING` report on a file nobody had touched.

### Fixed with a test, not by editing the artifact

The file's stated claim is that it was captured from real responses. Hand-editing
the entry would trade a wrong shape for a wrong provenance — the worse failure,
because provenance gets believed.

The new test drives the **already-seeded** POST path, which `docs/webapp-api.md`
documents ("Idempotent: already seeded → `{seeded: true, started: false}`") and
which nothing exercised. There `existing is not None` short-circuits before
`start()`, so `error` and `frame_id` pass through deterministically. Two
observations — one with an error recorded, one on a runner that never failed —
pin both halves of both fields regardless of who wins the race. The seeded state is
a **real store row** written the way the seeder writes it, not a patched lookup, so
the recorder stays on and the published shape stays honest.

Regenerated: 3 type changes; 796 shapes and 151/151 coverage unchanged.
`["null","string"]` is a superset of both race outcomes, so the entry now admits
either machine instead of calling one of them a breaking change.

### A survey, because this one is not alone

94 fields across 43 route entries are frozen as `null`-**only** — a type nothing
can legitimately have unless it is a constant null. `frame_id` on both
`/example/session` verbs was two of them (fixed here). The rest — e.g.
`GET /frames[].model`, `GET /frames/…/plan.plan_id`, `PATCH /me.email` — are latent
versions of the same thing, waiting for the first capture that sees the populated
state.

A note in `response_capture.py` records this and distinguishes it from
`_MACHINE_STATE_KEYS`. The tempting fix for the other 90 is to elide them there;
that would delete real guarantees to silence a coverage gap. They are
**under-observed**, not host-dependent, and each wants a test that reaches the
other state.

## Area

- [ ] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [ ] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Test / harness
- [ ] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

Ran:

```text
  -> no breaking change; 796 route/status shapes; coverage 151/151

uv run pytest                      # full offline suite via the --check run above
uv run pytest tests/test_gateway.py -k example_seed
uv run pre-commit run --all-files  # clean

Determinism: the capture was repeated 5x and POST /example/session [ok] came back
["null","string"] for both `error` and `frame_id` every time. Before the fix the
same repetition gave `null` 12/12 on macOS while CI froze `string`.

Attribution: the drift was confirmed pre-existing by checking out the base
commit's version of the changed files and re-running the gate with conditions
held constant — same failure, so it was not caused by any concurrent work.

Verified again after rebasing onto current main (which moved host_dispatch.py and
store.py underneath a *captured* artifact): --check still clean, 796 / 151-of-151.
```

Not run:

```text
The Linux/py3.12 capture, i.e. the machine whose scheduling produced the original
`string`.
```

Reason:

```text
No Linux host here. The claim that the union holds on both is structural rather
than observed: `check_compatible` passes whenever the observed type set is a
subset of the frozen one, and ["null","string"] is a superset of both race
outcomes. CI on this PR is the actual proof, and it is the one place that can be.
```

## Risk and Reviewer Focus

Low blast radius: one new test, one comment block, and three type widenings in a
generated artifact. No production code path changes.

Look first at:

- The artifact diff. It should be exactly three `"type"` entries on the two
  `/example/session` rows and nothing else. If it is broader, the regeneration
  picked up something host-specific and should not be merged.
- The new test's use of a real `store.new_frame` + `update_frame` rather than a
  patched `_example_session_frame`. That is deliberate: a stub would have to carry
  the `stubbed_backend` marker, which pauses the recorder, which would defeat the
  point of the test.
- Whether the widened entry is the contract you want published. It says `error`
  and `frame_id` are nullable on both verbs, which matches the source.

Note `fix/byoc-confinement-probe-and-docs` independently arrived at
`error: ["null","string"]` via a re-capture. Corroborating, but that branch's
`frame_id` is still `null` on both verbs and it has no test forcing the seeded
state — so it is another lucky capture rather than a fix. Merge order does not
matter; this PR is what keeps it from regressing.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched: none of the hotspot files
      are touched by this PR.

### Tests

- [x] No new test requires live LLM, network, GPU, SSH, Docker, lab hardware, or
      secrets in the default offline suite.
- [x] Kernel / host-RPC / gateway changes include focused contract coverage or a
      documented smoke test. (This PR *is* contract coverage for a gateway route.)

### Core Dependency Policy

      stdlib web server.
      in-tree use site. (None added.)
- [x] New dependencies, if any, are documented and justified. (None.)

- [x] No unpublished research roadmap, internal assignment, benchmark detail, or
      unreleased experimental result is disclosed.
      verification notes. (Not security-sensitive.)
      guarantees.

### Docs

- [x] Docs match actual code behavior. The published contract now matches what
      the route returns, which is the entire point of the change.
      appropriate. (No behavior change; `docs/webapp-api.md` already described the
      idempotent path this now tests.)
      affected. (Not affected.)

### DIFF
--- a/docs/response-schemas.json
+++ b/docs/response-schemas.json
@@ -5799,7 +5799,10 @@
             ]
           },
           "frame_id": {
-            "type": "null"
+            "type": [
+              "null",
+              "string"
+            ]
           },
           "project_id": {
             "type": "string"
@@ -17192,7 +17195,10 @@
             ]
           },
           "frame_id": {
-            "type": "null"
+            "type": [
+              "null",
+              "string"
+            ]
           },
           "project_id": {
             "type": "string"
--- a/openai4s/server/response_capture.py
+++ b/openai4s/server/response_capture.py
@@ -70,6 +70,26 @@ SCHEMA_VERSION = 1
 #: quietly drop two unrelated routes' contracts as well.
 _MACHINE_STATE_KEYS = frozenset({"sandbox", "default_host", "gpu_name", "cuda_version"})
 
+#: A different failure that looks like the one above, and must not be fixed the
+#: same way.
+#:
+#: `merge` widens across observations, so a nullable field ends up `["null",
+#: "string"]` *if the suite observes it both ways*. A field the suite only ever
+#: sees in one state freezes as that state alone -- and the resulting entry
+#: describes the run, not the API. `POST /example/session` was the first one
+#: caught: its `error` is `str | None` straight from `last_error()`, but the
+#: suite observed it exactly once, immediately after `start()` cleared the error
+#: and spawned a thread that may or may not have failed yet. Both threads then
+#: contend for one lock, so the field froze as `string` on Linux/CI and `null`
+#: on macOS, and the gate called the other machine a breaking change.
+#:
+#: The fix is a test that exercises the other state, never an entry here. These
+#: fields are not host-dependent -- they are *under-observed*, and eliding them
+#: would delete a real guarantee to silence a coverage gap. A survey at the time
+#: of writing found 94 fields frozen as null-only across 43 route entries
+#: (`frame_id` on both `/example/session` verbs among them); each is a latent
+#: version of this, waiting for the first capture that sees the populated state.
+
 #: How many incompatibilities to name per route before summarising the rest.
 #: One structural change can break dozens of nested fields, and a wall of them
 #: buries the first one, which is usually the cause of all the others.
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -186,6 +186,88 @@ def test_the_example_seed_route_reports_state_and_surfaces_its_error(
         runner.close()
 
 
+def test_the_example_seed_route_reports_the_already_seeded_state(tmp_path, monkeypatch):
+    """The idempotent path the API docs describe and nothing exercised.
+
+    `POST /example/session` on an install that already has the example must not
+    start anything: `existing is not None` skips `start()` entirely, so the
+    response reports the frame that is there and whatever error the last attempt
+    left behind. Worth a test on its own -- it is the difference between a
+    second click costing nothing and a second click running six cells.
+
+    It also closes a hole that surfaced somewhere unexpected. The frozen
+    `POST /example/session [ok]` shape in `docs/response-schemas.json` was built
+    from the single observation the suite happened to make, and that observation
+    sits immediately after `start()` has cleared `_last_error` and spawned a
+    thread that may or may not have failed yet. Both threads then contend for
+    the same lock, so `error` was captured as `string` on Linux/CI and `null` on
+    macOS -- a scheduler outcome published as a contract, in a file whose whole
+    claim is that it describes the API. The field is `str | None` on both verbs
+    (one dict literal, fed by `last_error()`), and so is `frame_id`
+    (`str` once seeded, `null` before). The two observations below pin both
+    halves of each, deterministically, without depending on which thread wins.
+    """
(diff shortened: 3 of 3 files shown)


Nobody-Zhang:
Reviewed by a parallel agent team. **Merging.**

### It does not fix the race — it stops the recorder having to publish one, which is the right call

`gateway.py` is untouched; `_ExampleSeedState.start()` still holds `_lock` across `thread.start()`. What this adds is a test that reaches the *other* legitimate state, so the captured union no longer depends on who wins the lock. Both halves now come from paths that never enter `start()` at all — a second POST on a seeded install (`existing is not None` short-circuits) and a fresh `SessionRunner` that never seeded. Two reviewers independently captured **5/5 identical `["null","string"]`** for both fields.

### The recorder is not masked — checked, because that was the thing to be suspicious of

`docs/response-schemas.json`'s entire claim is that it was captured from real responses, so a recorder that normalizes a racy field would be provenance that is wrong rather than absent. It doesn't: `git diff origin/main <head> -- openai4s/server/response_capture.py` with `#` lines filtered out leaves exactly one blank `+`. All 20 added lines are comment, and they argue *against* eliding these fields. `_MACHINE_STATE_KEYS` and `elide_machine_state()` are unchanged.

### #54's union must stay, and this PR is what earns it

Simulated rather than argued: with this PR's test in the suite, restoring `error: "string"` produces `POST /example/session [ok]: error.type widened from ['string'] to ['null','string']` — BREAKING, **deterministically**, taking the required `Offline tests (py3.12)` job red on every PR. #54 picked the right value and hand-asserted its provenance; this is the observation that justifies it. The frozen entry afterwards is exactly as strong as the route's guarantee — `error` is `last_error() -> str | None` and `frame_id` is `(existing or {}).get("frame_id") or …`, genuinely bivalent on both verbs.

### Two cosmetic notes, neither blocking

- **The PR body describes a base that no longer exists.** It claims "3 type changes" and that `error` is still frozen as `string` — both true only *before* #54 — and misattributes the surviving union to #62's branch. A reviewer applying the body's own acceptance criterion literally would reject a correct diff.
- The new comment's survey figure says "94 fields across **43** route entries"; the real count is 41 on main and 39 after this PR. Worth fixing given it sits in the one file whose subject is not publishing unmeasured numbers.

Merging via the admin bypass — GitHub refuses self-approval and `/openai4s/server/` has no second code owner who covers `docs/`.

PR #77, opened 2026-08-05
docs: refresh the root README and correct 28 directory READMEs against the code — plus the five source defects that audit found

Nobody-Zhang:
## Summary

Four commits: three documentation, and one that fixes the five `.py` defects the documentation pass found but was not authorized to touch. They are here rather than in a separate PR because they are the same finding — an audit that reads source to check what a README claims also finds where the source contradicts itself.

**`cae9c9b` — the root README described `v0.1.0`.** It now covers what landed since: read-only session sharing, seven source-attributed database connectors, a versioned `/api/v1`, environments as a transaction, a redacted support bundle, revocable telemetry, and a benchmark that runs against the real Store and kernels. Adds a "What ships today" capability map, splits the roadmap into Delivered / Next, and puts NVIDIA BioNeMo + Parabricks and the Windows/Linux package release on it.

Three items there were *wrong* rather than merely absent, and are corrected:

- The Linux and Windows install sections tell a reader to download a `tar.gz` and a `.zip` from the latest release. That release carries the macOS image only. A `> [!NOTE]` now says so.
- The roadmap listed a public workflow benchmark as future work, when 10 workflows / 20 cases already run against the real runtime. What is still ahead is publishing comparable results, so the entry says that instead.
- `README_zh.md` counted 33 bundled Skills. `skills/` holds 34.

**`1e8744d` — 28 directory READMEs, corrected against the code.** `scripts/check_directory_readmes.py` requires every direct file to be *named* in its README. The cheapest way to stay green is to append a filename and leave the prose alone, and five directories had done exactly that: `skills/`, `storage/`, `compute/`, `security/` and `server/` carried **thirteen files as dangling bullets below the closing line**, outside the Files table that claims to be the inventory. `server/` alone had seven. The gate was green throughout — it measures that a filename appears, not that a file is described.

The omissions were the smaller half. These claims had inverted:

| where | said | actually |
|---|---|---|
| `sdk/` | a compute API named `wait` | does not exist, and did not at `v0.1.0` |
| `kernel/` | background output is unbounded, grows daemon memory | capped; the concurrency slot is claimed before the spawn |
| `cli/` | the token gate applies only off loopback | required by default, loopback included |
| `openai4s_compute_provider/` | Linux probing compares network-namespace identity | compares the home directory's device id — and every unverifiable probe path used to answer `True` |
| `server/` + `webui/share/` | 3Dmol falls back to the `3Dmol.org` CDN | it does not, and `webui/` said so — two READMEs disagreeing about whether the page reaches a third-party host |
| `openai4s/` | the job registry lives only in memory; migrations live in `store.py` | both moved |
| `storage/` | a `category` projection on `memories.py` | no such projection; the real `update` was omitted |
| `docs/` + `scripts/` | `response-contract.json` is hand-written, "the side a human signs" | its own `note` says it is captured by driving every known route against a real handler |
| `envs/` | rdkit comes from pip | `python.yml` says the opposite, in a comment: `rdkit-pypi` is frozen at 2022.9.5 against the NumPy 1.x C ABI and segfaults the worker under 2.x |
| `.github/` | a contributor-wall workflow and a history secret scan | both deleted; the release job description predates its dispatch-only, draft-first design |
| `compute/` | the SSH path copies logs back | archives the whole workdir remotely, pulls it through the safe extractor |

**`8bced5a` — the five source defects the docs pass was not allowed to fix.** Each was re-verified against the code before being changed, not taken from the audit's word.

| where | said | actually |
|---|---|---|
| `cli/main.py:840` | "all nine answered with … 404", attached to `share` subcommands | the parser defines **eight** subcommands but the pre-fix body issues **nine** requests — `share create latest` resolves the session with a `GET /frames` of its own before it posts. Confirmed by counting the `/api/`-prefixed calls at `f2d8adb^`: exactly nine. The number was right; the noun it hung on was not, so the comment now says which it counts |
| `security/__init__.py` | "the three-layer security pipeline", four layers listed | six Cell-facing layers plus two boundaries that are not about a Cell, matching `security/README.md` as corrected in `1e8744d`. The blanket "every layer … fails open" was the worse error: `OPENAI4S_KERNEL_SANDBOX=enforce` and every SecretBroker mode fail **closed**, so failure behaviour is now stated per layer |
| `security/permissions.py` | the database "holds credentials in plaintext today" | predates `secret_broker.py` / `secret_migration.py`. Now records that the broker landed and that `auto` fails closed — without implying plaintext is gone, because it is not: `OPENAI4S_SECRET_STORE=plaintext` stays selectable by name, and an unmigrated or failed-migration value deliberately stays readable in place so the install keeps working (`migrate_connector_env`, `resolve_setting`) |
| `share_router.py:40` + `gateway.py:175` | `md_renderer.js`, in both static allowlists | no such file exists anywhere in the repo, and nothing generates one — the webui has no build step and is served straight from the working tree. `_load_share_assets` skipped it silently. Removed from both, and from the test fixture that mirrored them; a silently-skipped allowlist entry is the kind of thing that later gets "fixed" by creating the file |
| `execution/__init__.py` | — | did not re-export `QueueDepthExceeded`, although it is already in `coordinator.__all__` and `execution/README.md` claims the package re-exports the coordinator's tickets and errors. So the omission was the defect, not the claim: added, and the gateway now imports it from the package like its siblings instead of reaching into the module |

## Area

- [ ] Web UI / frontend
- [x] Server / gateway / API <sup>`gateway.py` + `share_router.py`: one dead allowlist entry, one import moved.</sup>
- [x] CLI / config / packaging <sup>`cli/main.py`: a comment only.</sup>
- [x] Tests / harness / CI <sup>`tests/test_share_router.py`: the fixture that mirrored the dead allowlist entry.</sup>
- [x] Docs

## Change Type

- [x] Bug fix <sup>The dead allowlist entry and the missing re-export; the other three are comment/docstring corrections.</sup>
- [ ] Feature
- [ ] Refactor
- [ ] Test / harness
- [x] Documentation
- [ ] Security-related change <sup>Left unchecked deliberately: two `openai4s/security/` **docstrings** change, no policy, threshold, or code path does.</sup>
- [ ] Release / packaging

## Verification

Ran, after `8bced5a`:

```text
  -> 9/9 hooks passed, mypy strict over the agent core + host dispatcher included

  -> contract up to date: 152/152 routes

  -> captured 819 route/status shapes
     coverage: 152/152 routes exercised by the offline suite
     no breaking change to the frozen response shapes

uv run pytest
  -> one failure, pre-existing on main and unrelated to this diff:
     test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained
     Verified by running it at base commit 98614e1 with none of this branch
     applied: same KeyError, same line. Filed as #78. The schema gate was
     therefore run with that one test deselected via PYTEST_ADDOPTS -- no file
     in the repo was changed to make a gate pass, and the test exercises no
     HTTP route, so route coverage is unaffected (152/152 either way).
```

Full numbers and the attribution evidence are in a [comment on this PR](#issuecomment-5187581185).

Ran for the documentation commits:

```text
  -> 101 maintained directories, 798 direct files/assets, complete bilingual coverage

  -> passed (1027 files)

Independent re-check over all 58 README files:
  - EN/ZH structural parity (heading sequence + table row counts) -> no mismatch
  - every relative Markdown link resolves -> none broken
  - local absolute paths in added lines -> 0
```

Not run:

```text
python -m harness.cli run --tier pr --offline
node tests/browser_smoke.mjs / browser_admission_fault.mjs / browser_matrix.mjs
uv build + scripts/verify_release_artifacts.py
```

Reason:

```text
`gateway.py` is touched, so both response gates apply and both are being run --
both have passed. No scenario, fault, trace, packaging or
resource file is touched, so the harness tier and the release-artifact checks
do not apply. The browser jobs need a live daemon; nothing here changes what a
page loads, since the one asset removed from the allowlists has no file behind
it and so was never served to begin with.
```

## Risk and Reviewer Focus

Runtime risk is small, but with `8bced5a` it is no longer nil. Three of that commit's five fixes are comment and docstring text with no runtime effect at all. The two that execute are: an allowlist entry removed for a file that has never existed in the repo — so nothing could ever have been served through it, and `_load_share_assets` was already skipping it — and a re-export plus the matching import move, which changes where a name is imported from, not what it is. Everything else in this PR is Markdown. So the real risk across all four commits remains that a *correction is itself wrong*, and review is about accuracy.

**Review depth, stated plainly.** The root README I wrote and verified claim by claim. The 28 directory READMEs were produced by seven agents working disjoint directory sets, each required to verify every claim against code before writing it and to report `file:line` or commit evidence for each change; I re-checked the gates, the bilingual parity, the links, the absolute-path scan and the secret scan, but did not re-read all 445 added lines against source myself. Reviewers should weight the security-adjacent rows accordingly.

**Look here first**, because these rows now make narrower or stronger claims about a boundary:

- `openai4s/security/README.md` — reframed from "six layers" to six Cell-facing layers plus two non-Cell boundaries; two matrix rows added stating *residuals* (existence under `$HOME` still observable via `EPERM` vs `ENOENT`; xattr names still readable on the traversal route; network never isolated; `plaintext` still exists as a mode).
- `openai4s_compute_provider/README.md` — the netns→device-id correction, and the fail-closed change. The name-based-heuristic caveat on secret scrubbing is deliberately kept explicit.
- `openai4s/cli/README.md` — the token-gate correction. This one previously described the gate as *weaker* than it is; confirm the new text against `gateway.py:8124-8148`.
- `openai4s/server/README.md` — the 3Dmol/CDN correction, and seven files promoted from dangling bullets into the Files table (text moved verbatim, no rewrite).

**Resolved in `bdb760b`, pushed after this PR was opened.** `openai4s/README.md`'s `server/` row said several UI and recovery workflows "remain Partial", which reads as work not yet done. That is not what the word means here: `server/README.md:41` and `docs/webapp.md:93-94` both use `Partial` for a recovery *outcome* — the verdict returned when an arbitrary historical namespace cannot be rebuilt and verified, which it deliberately never is. Verified by checking the UI call sites rather than the routes, as `CLAUDE.md` asks: `executeRecoveryAction` (`app.js:2917`), `forkSessionCheckpoint` (3025), `activateSessionBranch` (3017), `previewSessionRevert` (3026), `applySessionRevert` (3043, gated on `preview.can_apply`), `undoSessionRevert` (3006), `forkNotebookCell` (7872) and the four `NOTEBOOK_EXPORTS` forms (7502) are each bound to a real control and pass through `branchCapability()`/`branchCapabilityReason()`. At `app.js:7871` the Fork button is not rendered unless the cell carries a `fork_checkpoint_id` — so "Fork only from a record with a proven checkpoint mapping" is a control that does not appear, not a 409 a user hits. The row now says that, and the intro sentence names both remaining status kinds (`compute/`'s Prototype, and Partial-by-design).

**Resolved in `8bced5a`, pushed after this PR was opened.** This description previously listed one open decision: five defects inside `.py` sources, found by the audit and left alone as out of scope for a documentation-only pass. They are now fixed here, each re-verified against the code first — see the `8bced5a` table in the Summary for what each one turned out to be. One of the five did not survive re-verification as stated: the `cli/main.py` comment was reported as an off-by-one against eight `share` subcommands, and the count is in fact correct once you notice it counts requests, of which there are nine. That row corrects the wording rather than the number.

Reviewer note on that commit specifically: it is the one part of this PR I verified line by line myself rather than through agents, and every claim in its table is checkable from the diff plus `f2d8adb^`.

### Diff Readiness

- [x] I reviewed and understand the full diff. <sup>See "Review depth" above — the qualification is real.</sup>
- [ ] This PR is small and single-purpose. <sup>Single-purpose yes — one audit, its documentation corrections and its source corrections; small no — 65 files. Splitting the READMEs apart would leave the repo in a state where some directory READMEs contradict others, which is the condition being fixed, and splitting `8bced5a` out would separate five fixes from the audit rows that justify them.</sup>
- [x] Hotspot files were edited surgically if touched. <sup>Corrected: `gateway.py` **is** now touched, by exactly two lines — `md_renderer.js` dropped from `_SHARE_ASSET_NAMES`, and `QueueDepthExceeded` moved into the `openai4s.execution` import block that was already there. No routing, schema, or transport contract is altered. `host_dispatch.py`, `store.py`, `app.js`, `worker.py` and `manager.py` remain untouched.</sup>

### Tests

- [x] No new test requires live LLM, network, GPU, SSH, Docker, lab hardware, or secrets in the default offline suite. <sup>No test added.</sup>
- [x] Kernel / host-RPC / gateway changes include focused contract coverage or a documented smoke test. <sup>The gateway change is covered by `capture_response_contract.py --check` (152/152 routes, passed) and `capture_response_schemas.py --check`; `tests/test_share_router.py` continues to assert the static allowlist behaviour, with the fixture entry for the nonexistent file removed. No kernel or host-RPC change.</sup>

### Core Dependency Policy

- [x] Optional science imports are guarded by `try/except ImportError` at every in-tree use site. <sup>One import line changed, and it is stdlib-adjacent in the relevant sense: `QueueDepthExceeded` moves from `openai4s.execution.coordinator` to the `openai4s.execution` package. No third-party or optional-science import is involved.</sup>
- [x] New dependencies, if any, are documented and justified. <sup>None added.</sup>

- [x] No secrets, API keys, tokens, real credentials, private data, model weights, or local absolute paths are included. <sup>Verified: `source_secret_scan.py` passed over 1027 files; 0 local absolute paths among added lines.</sup>
- [x] Security-sensitive changes include appropriate synthetic tests or manual verification notes. <sup>Each security-adjacent correction cites the source line it was verified against; see "Look here first".</sup>
- [x] Docs do not overstate isolation, sandboxing, privacy, or security guarantees. <sup>This PR's main effect is the opposite: several boundaries were described as stronger, or as weaker, than they are.</sup>

### Docs

- [x] `README.md` and `README_zh.md` stay in sync where translated sections are affected. <sup>Verified structurally on every changed pair: identical heading sequence and table-row count.</sup>


Nobody-Zhang:
## Gate results for `8bced5a`

Both response gates now apply, because `gateway.py` is touched. Both pass.

```text
  -> 9/9 hooks passed (isort, black, ruff, check-json, mypy strict on the
     agent core + host dispatcher)

  -> contract up to date: 152/152 routes

  -> captured 819 route/status shapes
     coverage: 152/152 routes exercised by the offline suite
     no breaking change to the frozen response shapes
```

## The offline suite is red, and it is red on `main` too

One test fails:

```text
FAILED tests/test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained
>           assert usage["stdout_seen_bytes"] > 0, usage
E           KeyError: 'stdout_seen_bytes'
```

**It is not this PR.** Checked rather than assumed: the same test fails with the
same `KeyError` on the same line at base commit `98614e1`, with none of this
branch applied. This PR's `.py` diff touches `cli/`, `security/`, `server/` and
`execution/__init__.py`; the failing test drives `spawn_r_kernel` straight into
`kernel/` and `sink_drain.py` and reaches none of them.

It is a real defect, not a broken test — the R worker's *fallback* interrupt
path (`r_worker.R:668`) omits `sink_capture`, and the host gates both the
drained-output fill and the counter merge on that one flag (`manager.py:324-341`),
so an interrupt handled there discards output the host had already read. Filed
as #78, with the one step I did **not** confirm called out explicitly.

Worth knowing separately: **no workflow installs R**, and every R test is
`skipif(_REAL_R is None)`. So this never went red in CI and never will — the
whole R channel is skipped there. That half is in #78 too.

## How the schema gate was run, since it refuses on a red suite

`capture_response_schemas.py --check` deliberately refuses to compare when the
suite fails, so it was run with that one pre-existing failure deselected:

```bash
PYTEST_ADDOPTS="--deselect tests/test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained" \
```

Via `PYTEST_ADDOPTS` rather than by editing anything — no file in the repo was
modified to make a gate pass. The deselected test exercises no HTTP route, so it
cannot affect the route coverage the comparison is built on, and coverage came
back at the full 152/152 either way.

One correction on my own process, since it produced a result worth distrusting:
an earlier run of this gate failed on a *different* test
(`test_skill_write_authorization.py::test_a_legitimate_cwd_is_still_created`,
`OSError: Directory not empty: 'receipts'`) because I was running R kernel
reproductions concurrently with it. That file passes 10/10 on its own, and the
numbers above are from a clean re-run with nothing else executing. A gate result
produced under self-inflicted load is not evidence.

PR #90, opened 2026-08-14
fix(kernel): preserve R fallback capture and refresh contributors

### LINKED ISSUE #78: R kernel: an interrupt on the worker's fallback path drops the drained output and its counters (and no CI job runs any R test)
## Summary

When an R cell is interrupted on `r_worker.R`'s **fallback** path, the response
frame omits `sink_capture`. The host gates two separate things on that one flag,
so the cell loses both the output the host had already drained *and* the
counters that would have said so. `usage` comes back without
`stdout_seen_bytes` / `stdout_dropped_bytes`, which
[`docs/architecture.md:373`](../blob/main/docs/architecture.md) states the R
channel reports.

This is on `main` today. It is not caught by CI, for a reason worth its own
section below.

## Mechanism

`.oai4s_respond`'s signature defaults the flag to false
([`r_worker.R:111-113`](../blob/main/openai4s/kernel/r_worker.R)):

```r
.oai4s_respond <- function(id, stdout_txt, stderr_txt, error, interrupted,
                           ...,
                           sink_capture = FALSE) {
```

The normal execute path passes it explicitly (`r_worker.R:332-333`):

```r
  .oai4s_respond(id, "", "", err, interrupted, lineno, callname,
                 wall, cpu, .oai4s_rss_kb(), sink_capture = TRUE)
```

The top-level fallback does not (`r_worker.R:668-669`). This is the path for a
SIGINT that lands outside `.oai4s_run`'s own handlers — the comment above it
describes exactly that case, "a latched idle SIGINT firing at the next
checkpoint (before `.oai4s_run`'s own handlers arm)":

```r
      if (identical(outcome, "interrupted")) {
        .oai4s_respond(.oai4s_regex_id(line), "", "", "Interrupted", TRUE,
                       NULL, NULL, 0, 0, 0L)
```

Host side, both the output fill and the counter merge hang off that single flag
(`manager.py:324-341`):

```python
if capture is not None and frame.get("sink_capture"):
    frame["stdout"], frame["stderr"] = capture.finish()
    ...
    usage = frame.get("usage")
    if isinstance(usage, dict):
        usage.update(capture.counters())
```

So on the fallback path the host is holding real drained bytes in `capture`, and
hands back the worker's own empty `""` for `stdout`/`stderr` instead, with no
counter to indicate anything was dropped. The two failure modes the flag is
meant to distinguish — "the host read 300 MB and declined 299 of them" versus
"the worker quietly dropped them" — become indistinguishable again, which is the
thing `sink_drain.py`'s `counters()` docstring says it exists to prevent.

## Evidence

`tests/test_producer_output_budgets.py:787` asserts precisely this:

```
FAILED tests/test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained
>           assert usage["stdout_seen_bytes"] > 0, usage
E           KeyError: 'stdout_seen_bytes'
```

Reproduced on `main` at `98614e1` in a clean checkout, not only on a feature
branch:

```bash
uv run pytest tests/test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained -q
```

The test's own docstring states the contract it is defending: "what they had
already taken is what the cell reports."

**One step is not yet confirmed.** The mechanism above is read from the code and
is consistent with the observed `KeyError`, but I did not instrument which of the
two respond sites actually fired. The fallback passes `wall`, `cpu` and `rss` as
`0, 0, 0L`, so a `usage` carrying zeros in those three fields alongside the
missing counters would confirm it; a `usage` with a real wall time would mean the
gap is somewhere else and this diagnosis needs revisiting.

## Why CI does not catch this

Every R test is guarded by `@pytest.mark.skipif(_REAL_R is None, ...)`, and **no
workflow in `.github/workflows/` installs R** — `Rscript` does not appear
anywhere in that directory. So the entire R execution channel, which
`CLAUDE.md` and `docs/architecture.md` both describe as a first-class analysis
channel alongside Python, is skipped in CI on every run. `main` is green there
and red on any developer machine that has R installed.

That gap is arguably the more important half of this report: it is not that this
one test regressed unnoticed, it is that nothing in CI would notice if any R
kernel test regressed.

## Suggested fix

Pass `sink_capture = TRUE` at the fallback site too, since the host's fifos were
opened for that frame regardless of which path answers it. Worth checking the
other bare `.oai4s_respond` call sites (`r_worker.R:571`, `614`, `671`) against
the same question rather than fixing only the one the test happens to catch —
this repo has a documented pattern of a correct guard being wired to one call
site of several.

For the CI half: install R in the test job, or make the skip visible (a
collection-time warning, or a job that asserts the R tests actually ran) so a
silently-skipped channel cannot read as a passing one.

## Environment note, so a reader is not misled

On a machine whose R comes from a conda env built with OpenMP, this test also
fails intermittently — and under load, predominantly — with a *different* error
before it ever reaches the assertion:

```
RuntimeError: kernel worker exited unexpectedly: OMP: Warning #179: Function Can't set size of /tmp file failed:
Execution halted
```

That one looks environmental (OpenMP sizing a file under the kernel sandbox's
private temp) rather than a defect in this code path, but it masks the real
failure often enough to be worth knowing about when reproducing. It may deserve
its own issue.

## Not related to

Found while verifying #77, whose diff touches no kernel, R, or budget code. The
failure reproduces identically on the base commit with none of that PR applied.

Nobody-Zhang:
## Summary

Fixes #78.

The R worker's top-level fallback responses could omit `sink_capture = TRUE`
after the host had already drained the frame FIFOs. That discarded preserved
stdout/stderr and the host-derived seen, retained, dropped, and truncation
counters.

This change:

- marks both the interrupted and internal-error top-level R fallbacks as
  host-captured;
- leaves invalid-JSON and missing-sink refusals uncaptured because those paths
  do not own valid host-captured cell output;
- adds deterministic real-R coverage for both fallback outcomes and confirms
  the worker remains reusable;
- refreshes the bilingual Community contributors wall from GitHub's contributor
  graph, adding @ClarenceYC and @stau-7001;
- recognizes @EQSTLab from the accepted public credit on
  [GHSA-fm3g-2c7x-8qj8](https://github.com/PKU-YuanGroup/OpenAI4S/security/advisories/GHSA-fm3g-2c7x-8qj8);
- makes that non-commit recognition durable in the generator and preserves an
  existing current-contributor avatar if a transient refresh fails.

Only accepted, publicly visible recognition belongs in the maintained
non-commit list. Pending credits remain private until accepted.

## Area

- [ ] Web UI / frontend
- [ ] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

Ran:

```text
uv run pytest tests/test_r_kernel.py tests/test_producer_output_budgets.py::test_an_interrupted_cell_reports_what_the_host_had_drained tests/test_producer_output_budgets.py::test_r_top_level_fallback_preserves_host_capture -q
23 passed

uv run pytest tests/test_update_contributors.py -q
3 passed

uv run --with pillow python scripts/update_contributors.py
13 contributors; 13 circular PNGs; readmes updated: none

directory docs: 103 maintained directories, 820 direct files/assets, complete bilingual coverage

uv run python scripts/reaudit_crosswalk.py --check
crosswalk up to date: 48 closed rows, evidence unchanged

uv run pytest tests/test_plan_crosswalk.py -q
18 passed

uv run pre-commit run --files README.md README_zh.md .github/contributors/README.md .github/contributors/README_zh.md .github/contributors/EQSTLab.png scripts/README.md scripts/README_zh.md scripts/update_contributors.py tests/README.md tests/README_zh.md tests/test_update_contributors.py
All hooks passed

uv run pytest
4770 passed, 26 skipped, 1 deselected

git diff --check origin/main...HEAD
Clean
```

The real-R checks ran with R 4.5.3. The contributor generator was run twice;
the second run was idempotent.

Not run:

```text
node tests/browser_smoke.mjs
```

Reason:

```text
The browser suites do not exercise the R worker's top-level fallbacks or the
README contributor generator. Hosted PR CI runs the browser matrix.
```

## Risk and Reviewer Focus

For #78, review the two outer fallback responses in
`openai4s/kernel/r_worker.R`: `sink_capture = TRUE` must remain limited to
paths where the host owns the frame FIFOs. The focused regression writes known
bytes through the real FIFOs and raises outside the normal cell handler, then
checks the returned streams, all output counters, fallback identity, and worker
recovery.

For the contributor refresh, review `scripts/update_contributors.py`. Commit
contributors retain GitHub's contribution-count order; accepted public
non-commit recognition is appended once, case-insensitively de-duplicated, and
merged only after a non-empty API result. Avatar refresh failures keep an
existing current PNG, while departed identities and legacy SVGs are still
pruned.

### Diff Readiness

- [ ] This PR is small and single-purpose.
      `host_dispatch.py`, `store.py`, `app.js`, `worker.py`, `manager.py`.
      reformatted.

### Tests

      a documented smoke test.

### Core Dependency Policy

      stdlib web server.
      in-tree use site.

      verification notes.
      guarantees.

### Docs

      appropriate.
      affected.


Nobody-Zhang:
Community contributor acknowledgement 🎉

When this PR merges, @ClarenceYC, @stau-7001, and @EQSTLab will be added to the bilingual Community contributors wall. Thank you for your code and publicly recognized security contributions to OpenAI4S!

openai4s/kernel/r_worker.R
```diff
@@ -666,11 +666,11 @@ repeat {
     if (!.oai4s_responded) {
       if (identical(outcome, "interrupted")) {
         .oai4s_respond(.oai4s_regex_id(line), "", "", "Interrupted", TRUE,
-                       NULL, NULL, 0, 0, 0L)
+                       NULL, NULL, 0, 0, 0L, sink_capture = TRUE)
       } else {
         .oai4s_respond(.oai4s_regex_id(line), "", "",
                        paste0("openai4s r_worker ", outcome), FALSE,
-                       NULL, NULL, 0, 0, 0L)
+                       NULL, NULL, 0, 0, 0L, sink_capture = TRUE)
```
Nobody-Zhang: **A cell that shadows `.oai4s_respond` kills the kernel permanently.**

This fallback's `.oai4s_respond` call is outside any `tryCatch`, and `.oai4s_respond` / `.oai4s_write_frame` / `.oai4s_handle_line` / `.oai4s_responded` are ordinary unlocked globals in the same `globalenv()` user cells are evaluated in. `.oai4s_inspector` is `lockBinding`'d at :529-530 for exactly this reason — the guard was wired to one call site of several.

Reproduced:

```
kernel.execute(".oai4s_respond <- 1\ncat('shadowed\\n')")
-> RuntimeError: kernel worker exited unexpectedly
next cell        -> RuntimeError: kernel worker is not alive
is_alive()       -> False
```

`.oai4s_run`'s respond fails, the top-level `error=` handler fires, the fallback calls `.oai4s_respond` again, and an uncaught condition halts non-interactive Rscript. This breaks the invariant stated at :653-656 and the contract of `tests/test_r_kernel.py:341 test_real_r_worker_survives_hostile_cells` ("one frame may fail, the worker must survive them all"). `lockBinding()` on those four names is the matching fix.

scripts/update_contributors.py
```diff
@@ -113,25 +136,29 @@ def _circular_png(raw: bytes) -> bytes:
...
-        with open(os.path.join(AVATAR_DIR, f"{login}.png"), "wb") as f:
+        with open(path, "wb") as f:
             f.write(png)
-        ok.add(login)
-    # Drop anything (old SVGs, departed contributors) that is not a current png.
-    keep = {f"{login}.png" for login in ok}
+        have_png.add(login)
+    # Drop old identities and legacy SVGs, but keep a current contributor's
+    # committed PNG when a transient refresh fails.
+    current = {f"{person['login']}.png" for person in people}
```
Nobody-Zhang: **The prune deletes the avatar the new rescue path just claimed to keep, and the README then links the hole.**

`os.path.isfile` (line 149) answers case-insensitively on a case-preserving filesystem — the platform the module docstring names for "the daily local automation" — while this prune compares exact `os.listdir` names. Reproduced directly:

```
committed:      eqstlab.png        people: [{'login': 'EQSTLab'}]   (_get raising)
listdir after:  []
have_png:       {'EQSTLab'}
rendered:       <img src=".github/contributors/EQSTLab.png" ...>
file exists?:   False
```

A deleted committed binary plus a broken image on the front page of both READMEs. Pre-PR the same input left the login out of `ok`, so `render()` fell back to `https://github.com/EQSTLab.png`, which renders — this change converts a working remote fallback into a 404. The trigger goes live the moment EQSTLab lands a commit and the API returns canonical casing, since `RECOGNIZED_CONTRIBUTORS` is a hand-typed literal never reconciled with it.

Deriving `have_png` from a post-prune `os.listdir()` (exact match, spelled the way `render()` spells it) and comparing the prune case-insensitively fixes both halves and preserves the intended "keep the committed PNG on a transient failure" behaviour.

scripts/update_contributors.py
```diff
@@ -113,25 +136,29 @@ def _circular_png(raw: bytes) -> bytes:
 
 def write_avatars(people: list[dict], token: str | None) -> set[str]:
     os.makedirs(AVATAR_DIR, exist_ok=True)
-    ok: set[str] = set()
+    have_png: set[str] = set()
     for c in people:
         login = c["login"]
         url = c.get("avatar_url") or f"https://github.com/{login}.png"
```
Nobody-Zhang: **The new synthesized entry makes this fallback live, and it hands the repository token to a host that never asked for it.**

Every row from `fetch_contributors` already carries `avatar_url` on `avatars.githubusercontent.com`, so the `or f"https://github.com/{login}.png"` branch never fired in practice. The dict built at line 116 has no `avatar_url`, so it does now — and `_get` attaches `Authorization: Bearer <token>`.

Verified:

```
$ curl -D - 'https://github.com/EQSTLab.png?s=256'
HTTP/2 302
location: https://avatars.githubusercontent.com/u/148991397?v=4
```

CPython's `HTTPRedirectHandler.redirect_request` strips only `CONTENT_HEADERS = ("content-length", "content-type")` and copies every other header, so `GITHUB_TOKEN` is forwarded across that cross-host redirect once per recognized login on every daily run. The redirect also drops `?s=256`, so the 256px crop is lost.

```suggestion
        avatar_url = c.get("avatar_url")
        url = avatar_url or f"https://github.com/{login}.png"
```
…and fetch with `token if avatar_url else None`.

openai4s/kernel/r_worker.R
```diff
@@ -666,11 +666,11 @@ repeat {
     if (!.oai4s_responded) {
       if (identical(outcome, "interrupted")) {
         .oai4s_respond(.oai4s_regex_id(line), "", "", "Interrupted", TRUE,
```
Nobody-Zhang, line 675: **The fallback never closes the fifos `.oai4s_run` opened, so the host's drain never sees EOF.**

`.oai4s_unwind_sinks()` at :665 pops sink levels; `out_con`/`msg_con` are locals of the `.oai4s_run` frame the condition unwound, so the explicit `close()` pair at :323-324 — written precisely so the host's readers end on EOF rather than the grace period — never runs.

Measured on the production path (an `interrupt` signalled from a print method escapes the autoprint `tryCatch` at :300, which handles `error` but not `interrupt`):

```
normal cell      : 0.0007s
fallback cell #1 : 4.33s
fallback cell #2 : 4.32s
fallback cell #3 : 4.53s
fallback cell #4 : 0.22s   <- gc finally reclaimed the leaked connections
```

`SinkCapture.finish()` hits `BlockingIOError` with a live writer and burns `_FINISH_GRACE_SECONDS` on `_out` then again on `_err`, so `Kernel.execute` holds the FIFO execution ticket ~4.3s after a user clicks Stop. Worse: gc closes the connection later under `options(warn = 1)` (:556), so a **subsequent, unrelated** cell's captured stderr comes back as `closing unused connection 6 (/private/var/folders/.../out-N.fifo)` — the host's private `SinkDirectory` path leaking into a cell's output, the execution log, the model observation, and any shared session snapshot.

The latency is pre-existing (measured identical on `origin/main`), but the missing `close()` is squarely the subject of a PR whose stated purpose is making this fallback's capture correct.

openai4s/kernel/r_worker.R
```diff
@@ -666,11 +666,11 @@ repeat {
...
-                       NULL, NULL, 0, 0, 0L)
+                       NULL, NULL, 0, 0, 0L, sink_capture = TRUE)
```
Nobody-Zhang: **`0L` here fabricates a measurement this file explicitly refuses to fabricate 450 lines earlier.**

`.oai4s_rss_kb()` at :206-215 says it outright:

> `NULL`, not `0L`. `0` is a measurement — "this cell used no memory" — and it is one this worker cannot make … the column is nullable precisely so it can be given.

The sibling refusal at :617 passes `NULL`. These two do not. It matters more now than before, because `sink_capture = TRUE` makes `manager.py:339-341` merge *genuinely measured* byte counters into the same `usage` dict — so real and fabricated numbers sit side by side and a consumer cannot tell them apart. Measured on this machine:

```
normal cell : peak_rss_kb=None
fallback    : wall_s=0.0 cpu_s=0.0 peak_rss_kb=0   (alongside real stdout + real stdout_seen_bytes)
```

Those three survive verbatim into `execution_log.wall_s/cpu_s/peak_rss_kb` (`storage/frames.py:857-859`) and into `[usage wall=0.0s …]` (`agent/runtime.py:578`), so a ten-minute cell stopped during `print()` is recorded as having used no time and no memory. `tests/test_producer_output_budgets.py:855,864` now freezes the fabrication into the contract.

```suggestion
                       NULL, NULL, 0, 0, NULL, sink_capture = TRUE)
```

(`wall_s`/`cpu_s` cannot say "absent" yet — `.oai4s_num` renders `NULL` as `0` — so they remain the one place this frame still asserts a measurement it did not take.)

tests/test_update_contributors.py
```diff
@@ -0,0 +1,61 @@
...
+    merged = update_contributors.include_recognized_contributors(commit_people)
+
+    assert [person["login"] for person in merged] == ["MostCommits", "eqstlab"]
+
+    without_recognized = update_contributors.include_recognized_contributors(
+        commit_people[:1]
+    )
+    assert [person["login"] for person in without_recognized] == [
+        "MostCommits",
+        *update_contributors.RECOGNIZED_CONTRIBUTORS,
```
Nobody-Zhang: **This test passes with `RECOGNIZED_CONTRIBUTORS = ()` — it cannot fail if the feature the PR exists to add is deleted.**

The expectation splats `*update_contributors.RECOGNIZED_CONTRIBUTORS`, re-deriving the expected value from the code under test. Reproduced:

```
$ RECOGNIZED_CONTRIBUTORS = ()  # then re-run both assertions
BOTH ASSERTIONS STILL PASS      # ["MostCommits"] == ["MostCommits", *()]
```

The contributor wall can silently lose EQSTLab with a green suite. The first assertion (line 12) also tests *non*-appending (case-insensitive de-dup), the opposite of what the function name claims — two separate tests jammed into one.

```suggestion
    assert [person["login"] for person in merged] == ["MostCommits", "EQSTLab"]
```

…with the de-dup case split into its own test.

tests/test_producer_output_budgets.py
```diff
@@ -796,6 +796,89 @@ def test_an_interrupted_cell_reports_what_the_host_had_drained(tmp_path):
...
+.oai4s_saved_handle_line <- .oai4s_handle_line
+.oai4s_handle_line <- function(line) {{
+  assign('.oai4s_handle_line', .oai4s_saved_handle_line, envir = globalenv())
+  frame <- jsonlite::fromJSON(line, simplifyVector = TRUE)
+  out_con <- fifo(as.character(frame$sink_out), open = 'wb', blocking = TRUE)
+  err_con <- fifo(as.character(frame$sink_err), open = 'wb', blocking = TRUE)
+  writeBin(charToRaw('fallback stdout\\n'), out_con)
+  writeBin(charToRaw('fallback stderr\\n'), err_con)
+  flush(out_con); flush(err_con)
+  close(out_con); close(err_con)
```
Nobody-Zhang, line 849: **The fixture produces the one fallback shape production never produces.**

`flush()` then `close()` on both fifos before raising means the host's readers end on EOF. The real fallback cannot do that: the connections are locals of the unwound `.oai4s_run` frame, and `.oai4s_unwind_sinks()` pops sink levels without closing them.

Measured: this shape returns in **0.21s**; the production shape returns in **4.33s**. The stall, the leaked connection, and any flush/close-ordering regression are all invisible here — the test is green either way.

A deterministic variant needs no extra machinery and no timing:

```r
print.oai4sprobe <- function(x, ...) {
  cat('written through the live sink\n')
  signalCondition(structure(list(message = 'ctrl-c'),
                            class = c('interrupt', 'condition')))
}
```

then autoprint a value of that class — it escapes the `tryCatch` at `r_worker.R:300` (which handles `error` but not `interrupt`) with the sinks still live, which is exactly the path this PR fixes.

scripts/update_contributors.py
```diff
@@ -177,6 +204,7 @@ def main() -> int:
     if not people:
```
Nobody-Zhang, line 225: **This guard only catches the zero case, and the prune key widened to make the non-zero case destructive.**

The module docstring names this exact failure mode for the tool being replaced — contrib.rocks "gets rate-limited for this repo, and rendered only a single avatar". A truncated `/contributors` page passes a truthiness check.

With the prune key moved from "written this run" to "every login in `people`" (line 157), an API returning 1 of 12 contributors leaves `current` holding 2 filenames and deletes the other 10 committed PNGs, while both READMEs are rewritten to a 2-avatar wall.

`tests/README.md:310` claims the new test proves it "still fails closed on an empty API result" — true only for the zero case, which is not the case that happens. A floor (refuse to prune when the fetch returns implausibly fewer people than there are committed avatars), or making the prune opt-in, would cover the real one.

tests/test_producer_output_budgets.py
```diff
@@ -796,6 +796,89 @@ def test_an_interrupted_cell_reports_what_the_host_had_drained(tmp_path):
...
+  {raise_condition}
+}}
+"""
+        primed = kernel.execute(setup)
+        assert primed["error"] is None, primed
```
Nobody-Zhang, line 854: **This assertion sits between installing the dispatcher patch and its self-restore, so a failure here leaks the patch into teardown.**

The patch restores itself on its *first* call, which is `kernel.execute("ignored")` at line 848. Any failure at this line — or a `KeyboardInterrupt`/pytest timeout in that window — drops into `finally: kernel.shutdown()`, whose `{"type":"shutdown"}` frame the still-patched handler intercepts. A shutdown frame has no `sink_out`, so `fifo(as.character(NULL), ...)` errors, the top-level fallback answers with a bogus `response`, and the `repeat` loop never breaks.

Reproduced: `shutdown()` burns its full **5.00s** `wait(timeout=5)` then SIGKILLs, leaving the worker unreaped — per parametrization, on top of the real assertion failure.

Cheapest fix is to make the patch pass non-execute frames through:

```r
  if (is.null(frame$sink_out)) {
    return(.oai4s_saved_handle_line(line))
  }
```

which takes teardown from 5.00s to 0.02s (measured).

scripts/update_contributors.py
```diff
@@ -95,6 +100,24 @@ def fetch_contributors(token: str | None) -> list[dict]:
...
+def include_recognized_contributors(people: list[dict]) -> list[dict]:
+    """Append public non-commit contributors without duplicating API users."""
+
+    merged = list(people)
+    seen = {
+        str(person.get("login", "")).casefold()
+        for person in merged
+        if person.get("login")
+    }
+    for login in RECOGNIZED_CONTRIBUTORS:
```
Nobody-Zhang: **The recognition list skips both guards `fetch_contributors` applies, and uses a different identity rule.**

Line 94 rejects `type != "User"` and any login in `EXCLUDE`; the append at line 116 fabricates `"type": "User"` and is never tested against `EXCLUDE`. A bot identity or an Organization account added here renders on the wall inside a list the fetcher deliberately restricts to human users.

The same script also answers "is this the same person?" two ways in one run — exact match at line 94, `casefold()` here — and that divergence is the root of the filename bug at line 157.

```suggestion
    for login in RECOGNIZED_CONTRIBUTORS:
        if login.casefold() in seen or login in EXCLUDE:
```

scripts/update_contributors.py
```diff
@@ -43,6 +45,9 @@
 # unlinked grey avatar in the sidebar) and the /contributors API omits it, so
 # it is naturally excluded here too.
 EXCLUDE = {"github-actions[bot]", "dependabot[bot]", "actions-user"}
+# Publicly accepted contributions that are not represented in the commit graph.
+# Keep this list limited to GitHub logins whose recognition is already public.
+RECOGNIZED_CONTRIBUTORS = ("EQSTLab",)
```
Nobody-Zhang, line 50: **The recognition entry pins a mutable login with no account id, in the one path that re-fetches and re-commits an image daily.**

Every API row is re-derived from the commit graph each run and carries an immutable `id`, so a renamed or deleted account self-corrects — its PNG falls out of `current` and is pruned. This pins nothing but the string `"EQSTLab"`, and the synthesized dict at line 116 deliberately drops `id`.

If GitHub releases the login for re-registration after a rename or deletion, the next run fetches whatever `https://github.com/EQSTLab.png` now resolves to, overwrites `.github/contributors/EQSTLab.png` with a stranger's avatar, and commits it — rendered on the project's front-page contributor wall under a security-credit line. `write_avatars` only asks whether the download succeeded, so nothing notices.

Recording the id (`{"login": "EQSTLab", "id": 148991397}`) and failing the entry when `GET /users/EQSTLab` no longer returns it would close this.

Nobody-Zhang:
Four more findings whose lines fall outside the diff hunks, so GitHub will not take them inline. The other 11 are posted as inline comments.

---

### `openai4s/kernel/r_worker.R:571` — `sink_capture` is hand-typed at 5 sites and wired to 3

The real invariant is "the host opened fifos for the frame I am answering," established exactly once by the sink check at :608-620, right next to a `.oai4s_responded` flag the loop already resets per frame at :657. Instead five response sites restate it by hand — and the two this PR touched are the ones that were wrong.

The invalid-JSON / no-jsonlite response at :571-576 is reached from an execute frame the manager **already created fifos for**, and it still says `FALSE`. So on an R install without `jsonlite` — an explicitly supported configuration per the comment at :573-574 — every cell returns a `usage` dict missing all eight `stdout_*`/`stderr_*` counters that every other R response on the same kernel carries. Low blast radius today (no consumer indexes those keys), but it is a per-frame schema difference for a distinction the worker has no reason to make.

Deeper fix: declare `.oai4s_frame_sinks` beside `.oai4s_responded` (:87), set it `TRUE` once the sink check passes, reset it at :657, make it the default of the `sink_capture` formal at :113, and delete the explicit argument from all five sites. Every present and future responder then inherits the right answer instead of remembering it — the 6th call site is otherwise a repeat of the bug this PR just fixed.

---

### `openai4s/kernel/r_worker.R:156` — `.oai4s_responded` is set before the frame is written

```r
  .oai4s_responded <<- TRUE
  .oai4s_write_frame(json)
```

`.oai4s_write_frame` returns a success flag (:108) that the caller discards. A cell runs `closeAllConnections()` and the fd-3 reopen at :98 also fails: `.oai4s_run`'s response is silently dropped, `.oai4s_responded` is already `TRUE`, the top-level fallback declines to retry, and `Kernel.execute` blocks on an id that never arrives until the watchdog kills the worker — the exact hang the oversized-frame replacement at :130-155 was written to prevent, reached by a different door. Setting the flag from the return value is a one-line change.

Related and latent: `Kernel.execute` (`manager.py:324-352`) returns on the first `type == "response"` frame without ever comparing `frame["id"]` to the `cell_id` it sent, while `inspect_variables` does check (`manager.py:431-437`). The fallback's id comes from a regex over the raw line and is `"unknown"` when absent (:548-551).

---

### `openai4s/kernel/r_worker.R:666-675` — the fallback is frame-type-blind

It fires for whatever `.oai4s_handle_line` was handling, and now unconditionally claims host capture. Of the frame types that reach it, the claim is true-in-fact for one (`execute` with sinks). For `inspect_variables` no capture is ever opened: `.oai4s_inspector$respond(...)` at :587/:597 writes its frame and only then sets `.oai4s_responded <<- TRUE` at :601, so a condition in that window leaves the flag `FALSE` and the fallback emits a **second** frame — `{"type":"response","id":"variables-<uuid>","sink_capture":true}` — for a request that had no fifos. The host raises `kernel protocol desynchronized during variable inspection` and the stray frame is left for the next `execute()` to consume.

Bounded today by that desync check, but `sink_capture` is a claim about host state the worker cannot observe. A cheap guard is `grepl('"sink_out"', line)` on the raw line, or the tracked flag above.

---

### `scripts/update_contributors.py:214` — a totally failed run prints a healthy count and exits 0

`have_png` changed meaning from "written this run" to "a file with that name exists", so `circular pngs: {len(have_png)}` no longer distinguishes a full refresh from a total failure.

Run the script without Pillow — the documented hard requirement — and `_circular_png` raises `ImportError` for all 13 contributors; each hits the `except`, each has a committed PNG, so `have_png` is full, the README block is byte-identical, and the output reads `circular pngs: 13; readmes updated: none` with exit 0. Nothing re-fetched a single avatar. Same for a total CDN outage. A cron log scraped for that summary line reads a fully-failed refresh as a fully-successful one, and the "how many did I actually write" number that `ok` used to carry is not re-established anywhere.

---

<sub>Two notes on things that checked out: the `sink_capture = TRUE` fix itself is genuine — reverting both hunks fails both parametrized cases, and on the real escape path `stdout` goes from `''` to the preserved bytes. And no `CLAUDE.md` rule is violated by this diff; `check_directory_readmes.py`, `reaudit_crosswalk.py --check`, `source_secret_scan.py`, `pre-commit` on the changed files, and the full offline suite are all green at head. Separately worth knowing: no workflow under `.github/workflows/` installs R, so the R test gating this fix is skipped in CI and the two `closed` crosswalk rows re-recorded here are backed by a test that has not executed since the PR landed.</sub>

PR #96, opened 2026-08-18
fix(doubao,mcp): a fully-read response died on its own closed socket

Nobody-Zhang:
## Summary

Doubao Search failed with `Doubao search request failed (OSError)` *after* the
upstream answered HTTP 200 and the reader had consumed every result byte.

`HTTPResponse.read1` retires the transport on the same call that returns the
final byte of a known `Content-Length` (CPython 3.11+; only the 3.10 floor waits
for one further empty read), and urllib closed the connection socket back when
the headers arrived, so that last read drops the remaining `makefile` I/O
reference and the descriptor goes away. Both capped body readers re-armed the
socket read timeout at the top of every pass, so arming it once more after the
last chunk raised `OSError` (EBADF), and the controlled boundary projected a
finished exchange as a failed request. `mcp_http.py::_read_body` is the same
loop and failed the same way for any MCP HTTP reply carrying a `Content-Length`.

This was not a race: it was every such response on every interpreter above the
3.10 floor. It stayed invisible because the development venv is 3.10 — the one
version that waits — and every offline test used a `BytesIO` fake with no socket
underneath, and nothing under those refuses a `settimeout`.

A reader now stops when the body is exhausted, and the timeout setter skips a
socket whose descriptor is already gone while still raising for a live one. A
closed socket cannot block, so it needs no bound; a live one must keep its
bound, so the slow-drip protection is unchanged.

## Area

- [ ] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [ ] Docs

No box names it exactly: the change is in the shared outbound stdlib HTTP
transport (`http_deadline.py`) and its two consumers, the Doubao Search client
and the MCP HTTP client. No route, serializer, or response shape was touched.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Test / harness
- [ ] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

Ran:

```text
uv run pytest tests/test_doubao_search.py tests/test_mcp_client.py tests/test_egress_surface.py   # 66 passed
uv run pytest tests/test_doubao_search.py tests/test_mcp_client.py   # on 3.12 and on 3.13: 56 passed each
uv run mypy                                            # clean
uv run python scripts/check_directory_readmes.py       # complete bilingual coverage
uv run python -m harness.cli run --tier pr --offline   # 15/15
pre-commit isort / black / ruff / mypy-core            # all Passed (commit hook)
uv run pytest                                          # see note below
each new test, re-run with the source change reverted   # every one fails, on 3.10 and on 3.13
one live request to the provider through a local daemon's own /doubao-search/search route
  # available=true, source=doubao, count=8 (before the change: the OSError above)
```

The full offline suite reports 25 failures in this environment, byte-identical
to a clean tree: no local R runtime, and Seatbelt plus outbound TLS restricted.
None are in the touched modules.

Not run:

```text
scripts/capture_response_schemas.py --check
scripts/capture_response_contract.py --check
scripts/container_smoke.sh
tests/browser_smoke.mjs / browser_admission_fault.mjs / browser_matrix.mjs
```

Reason:

```text
No gateway route, serializer, or response envelope changed, so the two frozen
response contracts cannot move. Docker and Playwright were unavailable in the
environment this was prepared in; CI runs all four jobs.
```

## Risk and Reviewer Focus

Both changes bear on whether a body read stays bounded, so read them together:

1. `http_deadline.socket_timeout_setter` now returns a wrapper rather than the
   bound `settimeout`. `_socket_retired` deliberately tests `fileno() < 0` only.
   It must **not** consult `_closed`: urllib calls `sock.close()` while
   `makefile` still holds an I/O reference, so `_closed` is true for the whole
   body and treating it as retired would silently drop the read bound for every
   response. The `except OSError` path re-raises unless the socket is retired,
   so a live transport that refuses a timeout is still a hard failure.
2. `response_body_exhausted` returns on `isclosed()` or a remaining `length` of
   zero. Both are end-of-body, not error states; the second is what the 3.10
   floor reports one read earlier. A reader stopping there returns the same
   bytes and skips only a read that would have returned `b""`.

The new deterministic fake in `tests/test_doubao_search.py` is the contract
worth reviewing: it models the stdlib behaviour so the 3.10 CI job fails on this
too, rather than passing because its interpreter is the immune one.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched: none of `gateway.py`,
      `host_dispatch.py`, `store.py`, `app.js`, `worker.py`, `manager.py` is
      touched.
      reformatted.

### Tests

      or secrets in the default offline suite. The new socket test uses a local
      `socketpair`, and the network-policy calls around it are patched out so no
      DNS lookup or connection leaves the process.
      a documented smoke test.

### Core Dependency Policy

      stdlib web server.
      in-tree use site.
- [x] New dependencies, if any, are documented and justified: none added.

      verification notes.
      guarantees.

### Docs

- [x] Docs match actual code behavior: no documented behaviour changes, the
      reasoning lives in the touched docstrings.
      appropriate.
      affected: no translated section is affected.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/openai4s/doubao_search.py
+++ b/openai4s/doubao_search.py
@@ -18,7 +18,11 @@ from collections.abc import Callable, Mapping
 from typing import Any, Protocol
 
 from openai4s import datapro, egress, webtools
-from openai4s.http_deadline import HTTPExchangeDeadline, socket_timeout_setter
+from openai4s.http_deadline import (
+    HTTPExchangeDeadline,
+    response_body_exhausted,
+    socket_timeout_setter,
+)
 from openai4s.mcp_protocol import redact_reflected_secret
 
 ENDPOINT = "https://open.feedcoopapi.com/search_api/web_search"
@@ -159,6 +163,12 @@ def _read_capped(
                 f"Doubao search response exceeded the {limit}-byte limit"
             )
         chunks.append(chunk)
+        if response_body_exhausted(response):
+            # The last content byte and the transport's retirement arrive on the
+            # same read.  Looping to prove it with one more read asked a closed
+            # socket to re-bound a read that cannot happen, which reported a
+            # complete HTTP 200 as a failed request.
+            return b"".join(chunks)
 
 
 def _safe_error_code(value: Any) -> str:
--- a/openai4s/doubao_search.py
+++ b/openai4s/doubao_search.py
@@ -120,7 +120,7 @@ def _read_capped(
     response: Any,
     *,
     limit: int,
-    deadline: float,
+    exchange: HTTPExchangeDeadline,
     require_bound: bool = True,
 ) -> bytes:
     """This service's failure vocabulary over the shared bounded reader.
@@ -140,7 +140,7 @@ def _read_capped(
     return read_body_capped(
         response,
         limit=limit,
-        deadline=deadline,
+        exchange=exchange,
         on_timeout=lambda: DoubaoSearchError("Doubao search request timed out"),
         on_oversize=_oversize,
         on_truncated=lambda: DoubaoSearchError(
@@ -443,7 +443,7 @@ class DoubaoSearchService:
                 raw = _read_capped(
                     response,
                     limit=MAX_RESPONSE_BYTES,
-                    deadline=exchange.deadline,
+                    exchange=exchange,
                     require_bound=self._opener is None,
                 )
         except urllib.error.HTTPError as error:
--- a/openai4s/doubao_search.py
+++ b/openai4s/doubao_search.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import json
 import math
 import socket
-import time
 import urllib.error
 import urllib.parse
 import urllib.request
@@ -18,11 +17,7 @@ from collections.abc import Callable, Mapping
 from typing import Any, Protocol
 
 from openai4s import datapro, egress, webtools
-from openai4s.http_deadline import (
-    HTTPExchangeDeadline,
-    response_body_exhausted,
-    socket_timeout_setter,
-)
+from openai4s.http_deadline import HTTPExchangeDeadline, read_body_capped
 from openai4s.mcp_protocol import redact_reflected_secret
 
 ENDPOINT = "https://open.feedcoopapi.com/search_api/web_search"
@@ -126,49 +121,44 @@ def _read_capped(
     *,
     limit: int,
     deadline: float,
-    timeout_setter: Callable[[float], Any] | None = None,
+    require_bound: bool = True,
 ) -> bytes:
-    raw_length = response.headers.get("Content-Length")
-    if raw_length:
-        try:
-            if int(raw_length) > limit:
-                raise DoubaoSearchResponseError(
-                    f"Doubao search response exceeded the {limit}-byte limit"
-                )
-        except ValueError:
-            pass
+    """This service's failure vocabulary over the shared bounded reader.
 
-    chunks: list[bytes] = []
-    total = 0
-    read_once = getattr(response, "read1", None)
-    if not callable(read_once):
-        read_once = response.read
-    while True:
-        remaining = deadline - time.monotonic()
-        if remaining <= 0:
-            raise DoubaoSearchError("Doubao search request timed out")
-        if timeout_setter is not None:
-            timeout_setter(remaining)
-        # ``BufferedReader.read(n)`` may issue multiple recv calls while trying
-        # to fill ``n``; a slow-drip peer can therefore refresh the same
-        # relative socket timeout indefinitely. ``read1`` performs at most one
-        # raw read, returning here so the absolute deadline and remaining
-        # socket timeout are recomputed for every chunk.
-        chunk = read_once(min(65_536, limit + 1 - total))
-        if not chunk:
-            return b"".join(chunks)
-        total += len(chunk)
-        if total > limit:
-            raise DoubaoSearchResponseError(
-                f"Doubao search response exceeded the {limit}-byte limit"
+    Every property of the loop -- ``read1`` over buffered ``read``, the
+    absolute deadline, the byte cap, stopping at end-of-body, refusing a
+    truncated body -- is stated once in ``http_deadline`` and shared with the
+    MCP transport, which had to be given each of them separately while the two
+    loops were maintained side by side.
+    """
+
+    def _oversize() -> BaseException:
+        return DoubaoSearchResponseError(
+            f"Doubao search response exceeded the {limit}-byte limit"
+        )
+
+    return read_body_capped(
+        response,
+        limit=limit,
+        deadline=deadline,
+        on_timeout=lambda: DoubaoSearchError("Doubao search request timed out"),
+        on_oversize=_oversize,
+        on_truncated=lambda: DoubaoSearchError(
+            "Doubao search response ended before its declared length"
+        ),
+        # An injected opener is a test transport with no socket beneath it.  The
+        # daemon's own opener always has one, and a body read with no bound
+        # there is the slow-drip exposure this module exists to close.
+        on_unbounded=(
+            (
+                lambda: DoubaoSearchError(
+                    "Doubao search response has no bounded read transport"
+                )
             )
-        chunks.append(chunk)
-        if response_body_exhausted(response):
-            # The last content byte and the transport's retirement arrive on the
-            # same read.  Looping to prove it with one more read asked a closed
-            # socket to re-bound a read that cannot happen, which reported a
-            # complete HTTP 200 as a failed request.
-            return b"".join(chunks)
+            if require_bound
+            else None
+        ),
+    )
 
 
 def _safe_error_code(value: Any) -> str:
@@ -450,16 +440,11 @@ class DoubaoSearchService:
                     raise DoubaoSearchError(
                         f"Doubao search request failed with HTTP {int(status)}"
                     )
-                timeout_setter = socket_timeout_setter(response)
-                if self._opener is None and timeout_setter is None:
-                    raise DoubaoSearchError(
-                        "Doubao search response has no bounded read transport"
-                    )
                 raw = _read_capped(
                     response,
                     limit=MAX_RESPONSE_BYTES,
                     deadline=exchange.deadline,
-                    timeout_setter=timeout_setter,
+                    require_bound=self._opener is None,
                 )
         except urllib.error.HTTPError as error:
             status = int(error.code)
--- a/openai4s/http_deadline.py
+++ b/openai4s/http_deadline.py
@@ -31,8 +31,36 @@ class HTTPExchangeTimeout(TimeoutError):
     """One HTTP exchange exceeded its absolute wall-clock deadline."""
 
 
+def _socket_retired(sock: Any) -> bool:
+    """Whether this transport's descriptor is already gone.
+
+    A closed socket cannot block, so it needs no read timeout -- and touching
+    one raises ``OSError`` (EBADF).  ``socket.close()`` alone is not this state:
+    urllib closes the connection socket while ``makefile`` still holds an I/O
+    reference, and the descriptor stays readable for the whole response body.
+    """
+
+    fileno = getattr(sock, "fileno", None)
+    if not callable(fileno):
+        return False
+    try:
+        return int(fileno()) < 0
+    except OSError:
+        return True
+    except Exception:  # noqa: BLE001 - a transport that cannot say is treated as live
+        return False
+
+
 def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
-    """Find a live response socket's timeout setter through urllib wrappers."""
+    """Bound-read-timeout callable for the socket backing a urllib response.
+
+    The callable becomes a no-op once that socket is retired.  ``http.client``
+    retires it on the *same* ``read1`` that returns the final content byte once
+    a known ``Content-Length`` reaches zero (CPython 3.11+; only the 3.10 floor
+    waits for one further empty read), so a loop that re-bounds the next read
+    before checking for more data touched a closed descriptor and turned a
+    complete, successful response into ``OSError``.
+    """
 
     pending = [response]
     seen: set[int] = set()
@@ -44,12 +72,49 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
         seen.add(identity)
         setter = getattr(candidate, "settimeout", None)
         if callable(setter):
-            return setter
+            sock = candidate
+
+            def set_read_timeout(value: float, _sock: Any = sock) -> None:
+                if _socket_retired(_sock):
+                    return
+                try:
+                    _sock.settimeout(value)
+                except OSError:
+                    # Lost the race with a close (the watchdog's, or the
+                    # stdlib's end-of-body one).  Only a still-live transport
+                    # can leave a read unbounded, so re-raise for that case
+                    # alone rather than reporting a finished exchange as a
+                    # network failure.
+                    if _socket_retired(_sock):
+                        return
+                    raise
+
+            return set_read_timeout
         for attribute in ("fp", "raw", "_sock"):
             child = getattr(candidate, attribute, None)
             if child is not None:
                 pending.append(child)
     return None
+
+
+def response_body_exhausted(response: Any) -> bool:
+    """Whether this response can no longer yield body bytes.
+
+    Both signals are end-of-body, not error states: ``isclosed()`` reports the
+    body reader ``http.client`` retires once the last content byte is delivered,
+    and a remaining ``length`` of zero is the same fact on the 3.10 floor, which
+    retires one read later.  A reader that stops on either never asks a retired
+    transport for more.
+    """
+
+    is_closed = getattr(response, "isclosed", None)
+    if callable(is_closed):
+        try:
+            if bool(is_closed()):
+                return True
+        except Exception:  # noqa: BLE001 - an unknown reader is treated as open
+            return False
+    return getattr(response, "length", None) == 0
 
 
 def _response_socket(response: Any) -> socket.socket | None:
@@ -407,5 +472,6 @@ class _DeadlineHTTPSHandler(urllib.request.HTTPSHandler):
 __all__ = [
     "HTTPExchangeDeadline",
     "HTTPExchangeTimeout",
+    "response_body_exhausted",
     "socket_timeout_setter",
 ]
--- a/openai4s/http_deadline.py
+++ b/openai4s/http_deadline.py
@@ -23,7 +23,7 @@ import socket
 import threading
 import time
 import urllib.request
-from collections.abc import Iterable
+from collections.abc import Iterable, Iterator
 from typing import Any, Callable
 
 
@@ -38,6 +38,11 @@ def _socket_retired(sock: Any) -> bool:
     one raises ``OSError`` (EBADF).  ``socket.close()`` alone is not this state:
     urllib closes the connection socket while ``makefile`` still holds an I/O
     reference, and the descriptor stays readable for the whole response body.
+
+    Only consulted *after* an arming attempt already failed, so "cannot say"
+    must mean "not provably retired": a transport that exposes no descriptor,
+    or whose ``fileno()`` fails for any reason other than a dead descriptor,
+    keeps its failure rather than silently losing its read bound.
     """
 
     fileno = getattr(sock, "fileno", None)
@@ -51,15 +56,44 @@ def _socket_retired(sock: Any) -> bool:
         return False
 
 
-def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
-    """Bound-read-timeout callable for the socket backing a urllib response.
+def _arm_read_timeout(sock: Any, value: float) -> None:
+    """Bound one read on ``sock``, tolerating a transport already retired.
 
-    The callable becomes a no-op once that socket is retired.  ``http.client``
-    retires it on the *same* ``read1`` that returns the final content byte once
-    a known ``Content-Length`` reaches zero (CPython 3.11+; only the 3.10 floor
-    waits for one further empty read), so a loop that re-bounds the next read
-    before checking for more data touched a closed descriptor and turned a
-    complete, successful response into ``OSError``.
+    One caller: :meth:`HTTPExchangeDeadline.register_response`, where the
+    watchdog can close this socket in the gap after ``remaining()`` proved the
+    budget intact.  The body readers do not need this and do not use it --
+    ``read_body_capped`` stops at end-of-body *before* it would re-arm, which
+    is the whole of that fix; a second guard on the same path would only be a
+    place for the two to disagree.
+
+    The retirement test runs after ``settimeout`` has refused, never before:
+    asked first it answers about a moment already past by the time the call
+    runs, and it disarms every transport that bounds reads perfectly well but
+    cannot report a descriptor.  A still-live socket refusing a timeout stays a
+    hard failure -- that is the only case that can leave a read unbounded.
+    """
+
+    try:
+        sock.settimeout(value)
+    except OSError:
+        # Lost the race with a close (the watchdog's, or the stdlib's
+        # end-of-body one).  A finished exchange is not a network failure.
+        if _socket_retired(sock):
+            return
+        raise
+
+
+def _walk_response_transports(response: Any) -> Iterator[Any]:
+    """Yield ``response`` and every object urllib nests a transport under.
+
+    One traversal, two consumers: the read bound wants whatever carries
+    ``settimeout``, the watchdog wants the real ``socket.socket``.  Written out
+    twice they could answer with *different* objects from one response, and the
+    bound would then be armed on a transport the watchdog is not tracking.
+
+    Lazy on purpose.  A caller that stops at its first match never descends past
+    it, which is the search order both consumers had while they each open-coded
+    this walk.
     """
 
     pending = [response]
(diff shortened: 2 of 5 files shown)


openai4s/doubao_search.py
```diff
@@ -159,6 +163,12 @@ def _read_capped(
                 f"Doubao search response exceeded the {limit}-byte limit"
             )
         chunks.append(chunk)
+        if response_body_exhausted(response):
```
Nobody-Zhang: **A body cut short of its own `Content-Length` still reads as complete.**

This exit is right for a *finished* body. The sibling exit eight lines up — `if not chunk: return b"".join(chunks)` (line 158, and `mcp_http.py:368`) — still treats an empty read as end-of-body without asking whether more was expected, and `response_body_exhausted` is exactly the signal that separates the two.

Reproduced through this reader over a real socket, peer declares `Content-Length: 99`, sends 59 bytes, closes:

    bytes actually returned : 59
    response.length left    : 40
    reader reported         : COMPLETE (no error raised)

The boundary then diagnoses a cut connection as a provider defect — `DoubaoSearchResponseError("Doubao search returned invalid JSON")`, and for MCP an invalid-JSON / "did not contain the requested JSON-RPC id" error. `isinstance(response.length, int) and response.length > 0` at that exit distinguishes them exactly.

Not applied in the review branch: it adds a new failure path, which is a behaviour change rather than a review fix.

openai4s/http_deadline.py
```diff
@@ -31,8 +31,36 @@ class HTTPExchangeTimeout(TimeoutError):
...
+        return False
+    try:
+        return int(fileno()) < 0
+    except OSError:
+        return True
+    except Exception:  # noqa: BLE001 - a transport that cannot say is treated as live
+        return False
+
+
 def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
```
Nobody-Zhang: **`register_response` arms the same response socket, unguarded — the one arming site the fix skipped.** (anchored here because that method is unchanged and outside the diff; it is `sock.settimeout(self.remaining())` further down this file.)

`remaining()` proving the budget is intact does not keep it intact. The watchdog can fire and `_close_socket` the very socket in the window between the argument evaluating and `settimeout` running. Reproduced by forcing that window:

    register_response raised bare OSError: [Errno 9] Bad file descriptor

That `OSError` escapes `exchange.open()`, so `__exit__` sees `exc_type is not None` and never promotes it to `HTTPExchangeTimeout` — the caller gets `Doubao search request failed (OSError)` / `MCP HTTP transport failed (OSError)`, the exact wrong projection this PR removes, for a deadline the caller could otherwise recognise.

Fixed in the review branch by extracting the guarded arming into `_arm_read_timeout(sock, value)` and using it from both `set_read_timeout` and `register_response`. Same probe after the change: `register_response: no error`, and the exchange surfaces `HTTPExchangeTimeout`.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
+    retires one read later.  A reader that stops on either never asks a retired
+    transport for more.
+    """
+
+    is_closed = getattr(response, "isclosed", None)
+    if callable(is_closed):
+        try:
+            if bool(is_closed()):
+                return True
+        except Exception:  # noqa: BLE001 - an unknown reader is treated as open
```
Nobody-Zhang: **`except Exception: return False` also discards the `length` signal — in exactly the state this helper exists to detect.**

When `isclosed()` merely *returns* `False` the function still consults `length`; when it *raises* — strictly less information — the second signal is thrown away too. And a wrapper proxying `isclosed` to an already-closed file object raises `ValueError: I/O operation on closed file` precisely once the transport is retired.

Verified against this module: a response with `length = 0` whose `isclosed()` raises returns `False`, so the reader takes one more pass and re-arms `settimeout` on the retired socket — the bug this PR fixes, reintroduced through the helper that is supposed to prevent it.

Fixed in the review branch: `except Exception: pass`, falling through to the independent signal. Same case now returns `True`.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
+    body reader ``http.client`` retires once the last content byte is delivered,
+    and a remaining ``length`` of zero is the same fact on the 3.10 floor, which
+    retires one read later.  A reader that stops on either never asks a retired
+    transport for more.
+    """
+
+    is_closed = getattr(response, "isclosed", None)
+    if callable(is_closed):
+        try:
+            if bool(is_closed()):
```
Nobody-Zhang: **`bool(is_closed())` calls any truthy object end-of-body.**

Verified against this module:

    response_body_exhausted(unittest.mock.Mock())      -> True
    response_body_exhausted(unittest.mock.MagicMock()) -> True

`Mock().isclosed()` returns a truthy `Mock`, so any Mock-shaped double handed to `_read_capped` / `_read_body` makes the reader return after its first chunk and a truncated body is reported as a complete HTTP 200 — surfacing as "returned invalid JSON", or as silently partial results when the first chunk happens to parse.

`isclosed()` on a real `HTTPResponse` returns a genuine `bool`, so `is_closed() is True` costs nothing and admits only an answer that means what it says. Fixed that way in the review branch; both Mock cases now return `False`.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
+                return True
+        except Exception:  # noqa: BLE001 - an unknown reader is treated as open
+            return False
+    return getattr(response, "length", None) == 0
```
Nobody-Zhang: **`length == 0` is equally true of `False`, and not true of `"0"`.**

Verified against this module: a response exposing `length = False` (a boolean "does it have a length?" flag) returns `True` → the reader stops after one chunk and returns a truncated body as a successful 200. A wrapper exposing the raw header value `length = "0"` returns `False` → on the 3.10 floor, where this branch is the only working signal, that reader keeps the pre-PR extra pass.

`isinstance(remaining, int) and not isinstance(remaining, bool) and remaining == 0` is exact. Fixed that way in the review branch.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
         seen.add(identity)
         setter = getattr(candidate, "settimeout", None)
         if callable(setter):
-            return setter
+            sock = candidate
+
+            def set_read_timeout(value: float, _sock: Any = sock) -> None:
+                if _socket_retired(_sock):
```
Nobody-Zhang: **The pre-check disarms transports that bound reads perfectly well, and cannot see the race it is paired with.**

Deciding retirement *before* the call answers about a moment that has already passed by the time `settimeout` runs, so the `except OSError` path below is what actually covers the watchdog-close race — the pre-check adds nothing there. What it does add is a failure mode of its own: verified with a transport whose `fileno()` raises `OSError` but whose `settimeout` works, `socket_timeout_setter` returns a **non-`None`** callable that records **zero** `settimeout` calls. Because it is non-`None`, `doubao_search.py:454` (`"Doubao search response has no bounded read transport"`) cannot fire, so every raw read runs unbounded while the guard reports the opposite.

Measured cost as well: 258.2 ns/call shipped vs 188.1 ns for the `except`-only form (175.5 ns for the pre-PR bound method) — ~37 ms over a 4 MiB slow-drip body.

Fixed in the review branch by dropping the pre-check and keeping the post-failure check, now in the shared `_arm_read_timeout`. A new test pins it: with the pre-check restored, the opaque-`fileno` case fails `assert opaque.timeouts == [2.5]`.

tests/test_doubao_search.py
```diff
@@ -513,6 +614,89 @@ def read(self, *_args, **_kwargs):  # pragma: no cover - must stay unused
...
+    assert response.socket.fileno() == -1
+    # Every raw read was still bounded while the socket was live, and the
+    # budget only ever shrank.
+    assert response.read1_calls > 1
+    assert len(response.socket.timeouts) == response.read1_calls
+    assert all(0 < value <= 15.0 for value in response.socket.timeouts)
+    assert response.socket.timeouts == sorted(response.socket.timeouts, reverse=True)
+
+
+def test_a_real_socket_exchange_survives_the_stdlib_end_of_body_close(
```
Nobody-Zhang, line 767: **This test passes against the unfixed reader on Python 3.10, so the floor CI entry gets no signal from it.**

Reran the PR's tests against the pre-fix source on 3.10.20:

    FAILED test_a_completely_read_body_is_not_reported_as_a_failed_request
    (test_a_real_socket_exchange_survives_the_stdlib_end_of_body_close passed)

Structural, not flaky: 3.10's `read1` has no `if not self.length: self._close_conn()`, so the descriptor is still live at `length == 0`, the next `read1` gets `fp.read1(0) -> b""`, and the loop leaves through `if not chunk`. 3.10 is `requires-python`, one of the three `ci.yml` offline matrix entries, and the venv the PR description says development runs on — so the commit message's "each new test … every one fails, on 3.10 and on 3.13" holds for the two modelled fakes but not for this one.

Two further limits worth stating rather than leaving to be discovered: `connect` is overridden wholesale (no `create_connection`, no `wrap_tls`, no `SSLSocket`) and `build_opener` is monkeypatched with a version that substitutes `ProxyHandler({})` for urllib's proxy discovery and omits `http_handler()` — so the comment "this exercises the deadline-aware transport the daemon actually uses" claims more than it covers.

Fixed in the review branch by stating both limits in the docstring; the coverage itself is unchanged.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
+                    # Lost the race with a close (the watchdog's, or the
+                    # stdlib's end-of-body one).  Only a still-live transport
+                    # can leave a read unbounded, so re-raise for that case
+                    # alone rather than reporting a finished exchange as a
+                    # network failure.
+                    if _socket_retired(_sock):
+                        return
+                    raise
+
+            return set_read_timeout
```
Nobody-Zhang: **Nothing pins the retired-socket half of the fix.**

Mutation-checked three ways: collapsing this whole closure back to `return setter` (and deleting `_socket_retired`) leaves `tests/test_doubao_search.py` + `tests/test_mcp_client.py` at 56 passed, and instrumenting the module shows `_socket_retired` entered 27 times across both files and returning `True` **zero** times.

That is structural: with the `response_body_exhausted` early return in place, neither reader ever re-arms after the last chunk, so `assert len(response.socket.timeouts) == response.read1_calls` holds identically with and without this branch. Only the reader-side half of the diff is covered; the watchdog-close case this exists for can be broken and ship green from every matrix entry.

Fixed in the review branch by adding `test_the_read_bound_is_dropped_only_for_a_transport_that_is_really_gone`, which exercises all three states directly: retired → no-op, live-but-refusing → raises, opaque-`fileno` → still bounded.

tests/test_doubao_search.py
```diff
@@ -75,9 +176,9 @@ def open(self, request, timeout=None):  # noqa: ANN001
         reply = next(self.replies)
         if isinstance(reply, BaseException):
             raise reply
-        if isinstance(reply, _Response):
-            return reply
-        return _Response(reply)
+        if isinstance(reply, (bytes, dict, list)):
```
Nobody-Zhang, line 192: **The dispatch was inverted from "wrap unless it is already a response" to "wrap only bytes/dict/list", so every other reply type now reaches the service raw.**

No current test changes path (all 18 `_RecordingOpener` call sites pass `dict`, `bytes`, `HTTPError`, or a response-shaped fake), but the guarantee is gone. Verified: adding `"not-json"` beside the existing `b"not-json"` in `test_malformed_or_unproven_success_is_rejected`'s parametrize table — the natural place for it — lets the `str` reach `_search_with_key`, where `getattr(response, "status", None) or response.getcode()` raises `AttributeError` and the boundary reports `DoubaoSearchError("Doubao search request failed (AttributeError)")`. The `pytest.raises(DoubaoSearchResponseError)` then fails while the test still reads as though it exercised the malformed-payload path.

Fixed in the review branch by keeping the dispatch total: bytes/dict/list are encoded, a response-shaped object passes through, anything else raises `TypeError` naming the type.

openai4s/mcp_http.py
```diff
@@ -367,6 +371,12 @@ def _read_body(self, response: Any, deadline: float) -> bytes:
             if total > _MAX_FRAME_BYTES:
                 raise MCPOversizedResponse("MCP HTTP response exceeded the 4 MiB limit")
             chunks.append(chunk)
+            if response_body_exhausted(response):
```
Nobody-Zhang: **This reader accepts a `None` timeout setter in silence; its twin refuses that exact state.**

`doubao_search.py:454` raises `"Doubao search response has no bounded read transport"` when `socket_timeout_setter` returns `None`. `_read_body` (line 352 in this file) has no equivalent: it just skips `timeout_setter(remaining)` on every pass and reads the whole body with only whatever idle timeout `register_response` armed at header time — no error, no log. One of the two files is wrong about whether an unbounded body read is acceptable, and a property only one of two identical consumers enforces is a property of that file, not of the mechanism.

Not applied in the review branch: making it a hard failure changes MCP behaviour. Worth deciding deliberately, ideally by moving the check into a shared reader (see the note on `http_deadline.py`).

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
         for attribute in ("fp", "raw", "_sock"):
             child = getattr(candidate, attribute, None)
             if child is not None:
                 pending.append(child)
     return None
 
 
+def response_body_exhausted(response: Any) -> bool:
```
Nobody-Zhang, line 172: **The right altitude for this fix is a shared bounded reader, not a predicate two loops must remember to call.**

`http_deadline.py` already owns the deadline, the socket registry and the watchdog. The read loop is the only part of the exchange it does not own — and it is the only part that has now been fixed twice. After normalising indentation, `doubao_search._read_capped` and `mcp_http._read_body` differ only in the cap constant, the exception classes, and whether `timeout_setter` is injected or computed; everything else (the `Content-Length` pre-check, the `read1`-or-`read` probe, `min(65_536, CAP + 1 - total)`, the oversize check, and now the exhaustion check) is identical. The earlier `read1`/slow-drip fix had to be written into both by hand too, and the prose explaining it has already drifted apart between the two copies.

Two more readers are visible today and got neither hardening: `webtools._read_capped` (the `host.web_fetch` path, buffer-filling `read()` with only an idle timeout) and `share/fetch.py`'s bundle reader. A `read_body_capped(response, *, limit, deadline, timeout_setter, on_oversize, on_timeout)` here would make the next fix a one-place change and would take the "no unbounded read" check with it.

Not applied in the review branch: it is a cross-module refactor, well beyond a single-purpose bug fix.

openai4s/http_deadline.py
```diff
@@ -44,14 +72,51 @@ def socket_timeout_setter(response: Any) -> Callable[[float], Any] | None:
...
+    """Whether this response can no longer yield body bytes.
+
+    Both signals are end-of-body, not error states: ``isclosed()`` reports the
+    body reader ``http.client`` retires once the last content byte is delivered,
+    and a remaining ``length`` of zero is the same fact on the 3.10 floor, which
+    retires one read later.  A reader that stops on either never asks a retired
+    transport for more.
+    """
+
+    is_closed = getattr(response, "isclosed", None)
```
Nobody-Zhang: **The `isclosed` accessor is re-resolved on every chunk, in loops that already demonstrate the hoisting pattern.**

Measured against a real `HTTPResponse` paused mid-body: `response_body_exhausted(resp)` = 92.9 ns/call, vs 36.1 ns for the hoisted `is_closed() or resp.length == 0` (26.0 ns for `resp.isclosed()` alone) — 61% of the call spent re-deriving a bound method that cannot change between chunks. Both call sites already hoist the sibling accessor once (`read_once = getattr(response, "read1", None)`), so the cheaper form is the one the files already use. Cost avoided: ~3.6 µs per 4 MiB/64 KiB response, ~30 ms per 4 MiB slow-drip response.

Not applied in the review branch: it changes the shape of the public helper this PR just introduced, which is the author's call.

openai4s/http_deadline.py
```diff
@@ -31,8 +31,36 @@ class HTTPExchangeTimeout(TimeoutError):
...
+
+    The callable becomes a no-op once that socket is retired.  ``http.client``
+    retires it on the *same* ``read1`` that returns the final content byte once
+    a known ``Content-Length`` reaches zero (CPython 3.11+; only the 3.10 floor
+    waits for one further empty read), so a loop that re-bounds the next read
+    before checking for more data touched a closed descriptor and turned a
+    complete, successful response into ``OSError``.
+    """
 
     pending = [response]
```
Nobody-Zhang, line 100: **This `fp`/`raw`/`_sock` walk is a verbatim duplicate of `_response_socket`, 55 lines below in the same file.**

Eleven identical lines — `pending = [response]`, `seen: set[int] = set()`, the `id()` dedupe, and the `("fp", "raw", "_sock")` push — differing only in the acceptance predicate (`callable(getattr(candidate, "settimeout", None))` vs `isinstance(candidate, socket.socket)`). They are the only two occurrences of that attribute tuple in `openai4s/`, and they have already diverged in kind: one accepts any duck-typed object, the other only a real socket, so in principle the watchdog and the read bound can end up tracking different objects. (For urllib's real chain they agree — `_sock` is the `SSLSocket`, which is a `socket.socket` — so this is duplication rather than a live bug.)

One `_walk_response_transports(response)` generator with both callers expressed as `next((c for c in walk if pred(c)), None)` removes it. Not applied in the review branch: `_response_socket` is untouched by this PR.

tests/test_mcp_client.py
```diff
@@ -275,6 +276,82 @@ def read(self, _size=-1):
...
+            self.fp = _FP(self.socket)
+            self._body = body
+            self._offset = 0
+            self.length = len(body)
+            self.read1_calls = 0
+
+        def isclosed(self):
+            return self.fp is None
+
+        def read1(self, size):
```
Nobody-Zhang: **This double's `read1` diverges from the stdlib signature its sibling models correctly.**

`http.client.HTTPResponse.read1(n=-1)` has a default and treats a negative `n` as "up to `length`"; the doubao double added in the same commit guards both (`def read1(self, size=-1)` … `if size is None or size < 0: size = self.length`). Here, `min(size, 8)` with a negative `n` slices `self._body[offset:offset-1]` and returns the wrong (usually empty) chunk, and a no-argument `read_once()` raises `TypeError`. If `_read_body`'s call ever moves to the stdlib-default form, this double returns `b""` on the first pass and the test asserts a *successful* zero-length read instead of failing.

Fixed in the review branch by giving it the stdlib signature and the same guard as its sibling.

tests/test_doubao_search.py
```diff
@@ -513,6 +614,89 @@ def read(self, *_args, **_kwargs):  # pragma: no cover - must stay unused
...
+    Nothing below ``socket`` is faked here, so ``http.client`` performs its own
+    end-of-body ``_close_conn()`` and the descriptor is genuinely gone.  This is
+    the shape that failed against the live provider while every fake-transport
+    test passed.
+    """
+
+    from openai4s import egress, webtools
+    from openai4s.http_deadline import HTTPExchangeDeadline, _DeadlineHTTPSConnection
+
+    client, server = socket.socketpair()
```
Nobody-Zhang: **Both descriptors are opened outside the block that closes them.**

`_store(tmp_path)`, `save_agent_plan_key`, and `DoubaoSearchService(store)` all run between this line and the `try:`, so any failure in them leaks the socketpair for the rest of the session. `store` has the mirror problem in the other direction — it is closed in `finally` but created outside the `try`.

Fixed in the review branch by moving the socketpair and the store construction inside the `try`, with `store = None` so the `finally` stays safe.

Nobody-Zhang:
**Review follow-up: the structural pass these comments were asking for**

The 15 comments above were written against this diff as it stands. I applied them locally and then put the result through a second, independent structural review. Five of the "not applied in the review branch" notes are now stale, one names a test that no longer exists, and one describes a fix that has since been superseded by a bigger cut. Replies are on the individual threads; this is the shape they converge on.

**One reader instead of two.** `doubao_search._read_capped` and `mcp_http._read_body` were the same ~40 lines. They are now `http_deadline.read_body_capped(response, *, limit, deadline, on_timeout, on_oversize, on_truncated, on_unbounded=None)`, and each consumer is only its own failure vocabulary. Everything the loop enforces -- `read1` over buffered `read`, the absolute budget, the byte cap, end-of-body, a body cut short of its declared length, refusing to read one at all with no bound -- is stated once.

**One stop condition, at the top of the loop.**

    while not exhausted():
        remaining = deadline - time.monotonic()
        if remaining <= 0:
            raise on_timeout()
        if arm is not None:
            arm(remaining)
        chunk = read_once(min(65_536, limit + 1 - total))
        if not chunk:
            if (_remaining_body_bytes(response) or 0) > 0:
                raise on_truncated()
            break
        total += len(chunk)
        if total > limit:
            raise on_oversize()
        chunks.append(chunk)
    return b"".join(chunks)

Testing exhaustion *before* the arming is what makes the retired socket unreachable, rather than something a tolerant `settimeout` has to absorb.

**The `settimeout` wrapper is gone.** `socket_timeout_setter` returns the bound `settimeout` again. Ablation on the result: remove that single stop condition and all three of this PR's new tests go red -- there is no second mechanism left to remove. Two sufficient defenses for one bug was the thing worth cutting, and cutting it also retires the `_closed` trap from the read path entirely.

**What survives, and why it is not the second defense.** `_socket_retired` plus a small `_arm_read_timeout` remain, with exactly one caller: `HTTPExchangeDeadline.register_response`. `remaining()` proving the budget intact does not keep it intact -- the watchdog can close that socket in the gap before `settimeout` runs, and the bare `OSError` that escapes leaves `__exit__` with `exc_type` set, so it never becomes the `HTTPExchangeTimeout` a caller can recognise. Reproduced: consumers reported `... failed (OSError)` for an abort. Restoring the bare `sock.settimeout(self.remaining())` there turns the new test red with `OSError: [Errno 9]`.

**The `_closed` invariant has a test now rather than a comment.** Over a real socketpair: `makefile` holds an I/O reference, `close()` leaves `_closed = True` with `fileno() >= 0`, `_socket_retired` is `False`, and the read bound is still armed. Making `_socket_retired` consult `_closed` turns it red -- which is exactly what the warning in this diff could not do on its own.

**Also folded in** (details on the threads): a body shorter than its own `Content-Length` is now a transport failure instead of "invalid JSON"; `_read_body` no longer accepts an unbounded read in silence where its twin called it fatal; the `fp`/`raw`/`_sock` walk is one lazy generator shared by the setter and the watchdog lookup; the end-of-body accessors are resolved once per response instead of per chunk; and both exhaustion signals are matched exactly (`is True`, and an `int` that is not a `bool`) -- `Mock()` and `length = False` both read as "exhausted" before.

**Verification.** Full offline suite green (5146 collected, 0 failures); the touched modules green on the 3.10 floor as well; `mypy`, `pre-commit`, and `harness --tier pr --offline` 15/15; plus one live request each through the real Doubao Search endpoint and the real DataPro Streamable-HTTP MCP endpoint -- both fail with `(OSError)` on this PR's parent commit and succeed on it.

**Left alone, deliberately.** `tests/test_mcp_client.py`'s `_EndOfBodyClosingResponse` re-models the same 3.11+ retirement as the double in `tests/test_doubao_search.py`. With the reader shared, one socketpair test on the helper covers that contract and this double is surplus -- but deleting a test the PR added is the author's call, not a reviewer's.

The changes are in a local worktree, not pushed. Happy to open them as a branch or a follow-up PR if that is easier to read than the prose.

openai4s/doubao_search.py
Nobody-Zhang: Applied after all, in the shared reader rather than at this exit. `if not chunk:` now raises `on_truncated()` when `_remaining_body_bytes(response)` is still positive; Doubao maps it to `DoubaoSearchError("Doubao search response ended before its declared length")` and MCP to the matching `MCPError`. Re-verified over a real socket -- peer declares `Content-Length: 99`, sends 59, closes: a transport failure now, instead of 59 bytes handed back as a complete body.

openai4s/http_deadline.py
Nobody-Zhang: Still the fix, and it is now the *only* caller of the guarded helper -- the wrapper on the read path is gone (see the pre-check thread). It has a test: restoring `sock.settimeout(self.remaining())` here fails with `OSError: [Errno 9]` instead of the `HTTPExchangeTimeout` the exchange is supposed to report.

openai4s/http_deadline.py
Nobody-Zhang: Superseded by a bigger cut. `socket_timeout_setter` no longer returns a wrapper at all, so there is neither a pre-check nor a post-check on the read path: with the exhaustion test at the top of the loop the reader never re-arms a retired socket. Ablation confirms it is load-bearing on its own -- remove that single stop condition and all three of this PR's new tests go red, with nothing left to cover for it. `_socket_retired` survives only for `register_response`.

openai4s/http_deadline.py
Nobody-Zhang: Correction to my own note: the test I named no longer exists. Removing the wrapper made its three synthetic transports moot. What replaced it is two tests over real sockets -- `test_a_socket_urllib_already_closed_still_takes_its_read_bound`, which pins the `_closed` trap this diff currently warns about only in a comment (`makefile` holds an I/O reference, `close()` leaves `_closed = True` with `fileno() >= 0`, and the bound must still be armed), and `test_a_watchdog_close_is_reported_as_the_deadline_not_as_a_bare_oserror`. Each fails against the corresponding mutation.

openai4s/mcp_http.py
Nobody-Zhang: Applied. `read_body_capped` takes an `on_unbounded` factory and the MCP transport passes one, so an unbounded body read is now fatal on both consumers rather than fatal on one; Doubao keeps its existing escape hatch for an injected test opener (`require_bound=self._opener is None`). The shared MCP response double gained the transport it was missing -- a fake with no socket beneath it is not a cheaper HTTP response, it is a state the real transport never produces.

openai4s/http_deadline.py
Nobody-Zhang: Applied. `read_body_capped(response, *, limit, deadline, on_timeout, on_oversize, on_truncated, on_unbounded=None)` lives in `http_deadline.py`; both consumers are now just their own failure vocabulary, with no loop of their own. One consequence worth knowing: `import time` became dead in both modules, and the slow-drip test's `monkeypatch.setattr(mcp_http.time, ...)` moved to `http_deadline.time`, which is where the budget is actually consulted now.

openai4s/http_deadline.py
Nobody-Zhang: Applied. `_body_exhausted_probe(response)` resolves the accessors once and the reader hoists it next to `read_once`, the way both loops already treated `read1`. `response_body_exhausted` stays as the public one-shot predicate, now defined as `_body_exhausted_probe(response)()`, so the rule itself is stated in one place.

openai4s/http_deadline.py
Nobody-Zhang: Applied. `_walk_response_transports(response)` is a generator, and `socket_timeout_setter` and `_response_socket` both consume it and stop at their own predicate. Lazy on purpose: a caller that stops at its first match never descends past it, which preserves the search order both open-coded walks had. Verified the two still select the same object for urllib's real chain.

Nobody-Zhang:
Pushed as #105, based on this branch rather than on `next`, so the diff under review there is only the structural pass described in the follow-up above — the ablation, the removed `settimeout` wrapper, the single stop condition at the top of the loop, the shared `read_body_capped`, and the three failures consolidation surfaced (a truncated body read as complete, an unbounded MCP read accepted in silence, and the loose exhaustion signals).

Merge it into this branch and #96 carries both, or close it and take the notes — either works. All of this PR's own tests are retained and still fail against its parent.

PR #99, opened 2026-08-19
feat(skills): integrate pinned bioSkills collection

Linmj-Judy:
## Summary

Vendor 561 GPTomics/bioSkills recipes from commit d91ed3d563019e649dc854c56ccd62551359488a with MIT provenance, a reproducible importer, and verified payload hashes. Add the RFdiffusion workflow skill and structural-analysis dependencies for Biopython, FreeSASA, and DSSP.

Keep the large collection discoverable without flooding the resident prompt: expose collection provenance through loader, Host, and Web catalogs; aggregate bioSkills in progressive disclosure; and direct bioinformatics tasks to search relevant English methods, tools, and data types before loading recipes.

Package imported scripts and references in wheels and sdists, exclude byte-exact vendored payloads from format hooks, and add bilingual documentation, release checks, response-schema updates, and regression coverage for catalog readiness, requirements, and retrosynthesis planning.

## Area

- [ ] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [ ] Security-related change
- [x] Release / packaging

## Verification

Ran:

```text
uv run pytest
uv build && uv run python scripts/verify_release_artifacts.py dist
```

Not run:

```text
uv run mypy
python3 -m harness.cli run --tier pr --offline
node tests/browser_smoke.mjs
```

Reason:

No files in the mypy-core boundary (agent/{actions,engine,finalize,models,ports}.py,
host_dispatch.py) or harness scenario code changed; no UI streaming or Docker image
changes. These gates still run as separate CI jobs on the PR.

Risk and Reviewer Focus

- The vendored tree (skills/bioskills/, ~1,900 files) is byte-exact from the pinned upstream commit and is not for line review — provenance lives at the boundary (skills/bioskills/README*, LICENSE, MANIFEST.json with per-file hashes), and scripts/import_bioskills.py pins UPSTREAM_COMMIT and EXPECTED_SKILLS = 561. Review the importer and spot-check the boundary docs instead.
- openai4s/skills_loader/loader.py — bundled_roots() adds a second bundled root under skills/bioskills/ with a collection tag; collision precedence (bundled wins) is unchanged, but the system_context(only=...) filter now matches on skill.name. Focus on the progressive-disclosure aggregation path and the scoped-prompt behavior.
- openai4s/tools/skills.py — ListSkillsTool.output_limit raised 10k → 50k so explicit enumeration can surface all 561 collection names; please sanity-check archive/token implications of explicit full listings.
- Packaging — pyproject.toml package-data globs and MANIFEST.in were extended (.R, .sh, .wdl, .nf, .cwl, fasta/alignment fixtures, snakefiles) so imported recipes resolve from wheel and sdist; scripts/verify_release_artifacts.py and scripts/bundle_contract.py were updated in the same commit.
- envs/struct.yml adds biopython/freesasa/dssp to the struct env. RFdiffusion is deliberately not installed (its pinned CUDA-specific PyTorch/DGL stack conflicts with the portable Python 3.13 CPU env) — the skill documents running it in a dedicated env/container. Nothing here installs packages at startup.
- None of the six surgical-edit hotspots (gateway.py, host_dispatch.py, store.py, app.js, worker.py, manager.py) were touched; the server/skills.py change is a 3-line metadata passthrough.

Checklist

Diff Readiness

- [x] Hotspot files were edited surgically if touched: gateway.py,
  host_dispatch.py, store.py, app.js, worker.py, manager.py.
- [x] openai4s/server/webui/vendor/ and tests/fixtures/ were not
  reformatted.

Tests

  a documented smoke test.

Core Dependency Policy

  stdlib web server.
- [x] Optional science imports are guarded by try/except ImportError at every
  in-tree use site.

Public Disclosure and Security

  verification notes.
  guarantees.

Docs

  appropriate.
- [x] README.md and README_zh.md stay in sync where translated sections are
  affected.


openai4s/skills_loader/loader.py
```diff
@@ -767,7 +790,7 @@ def bootstrap_code(self, *, allowed: frozenset[str] | None = None) -> str:
                 for entry in (manifest.get("entries") or [])
                 if entry.get("directory") and entry.get("name") not in allowed
             )
-        roots = [str(self.skills_dir)]
+        roots = [str(root) for root, _collection in self.bundled_roots()]
```
Nobody-Zhang: **correctness — the `bioskills.` prefix walks around the in-kernel sidecar deny gate**

`skills/bioskills` is a package *under* `skills/`, which is already on `sys.path`, so every recipe has a second importable spelling. The generated gate keys on the leaf directory name, so `find_spec` sees `top == 'bioskills'`, finds no entry, returns `None`, and ordinary `PathFinder` imports the file.

Reproduced:

```python
exec(compile(loader.bootstrap_code(allowed=frozenset({"example_stats"})), "<b>", "exec"), {})
importlib.import_module("bio-alignment-multiple-alignment.scripts.codon_alignment")
# ModuleNotFoundError: skill sidecar '...' is not available to this agent   <- correct
importlib.import_module("bioskills.bio-alignment-multiple-alignment.scripts.codon_alignment")
# <module ...>  __file__ = .../skills/bioskills/.../codon_alignment.py      <- executed
```

A delegated Specialist whose allowlist excludes the collection, or a user who toggled a Skill off in Customize, can import and run any of the 408 vendored `.py` files with one line. Second order: the `sys.modules` purge loop above also keys on `_o4s_skill_entries`, so `bioskills.*` modules survive a re-bootstrap with a changed manifest.

Fixed in the review branch: `bootstrap_code` now passes the collection root names to `_bootstrap_runtime_code`, and the gate normalises `bioskills.<dir>...` to `<dir>` *before* the deny/disabled decision, so both spellings resolve to the same skill identity. Unrestricted imports and sidecar tracking are unchanged.

openai4s/tools/skills.py
```diff
@@ -24,15 +24,18 @@ class ListSkillsTool(Tool):
     )
     parameters = {"properties": {}, "required": []}
     requires_approval = False
-    output_limit = 10_000
+    # Explicit all-Skills audits must be able to see the complete catalog.  The
+    # pinned bioSkills collection adds 561 names, so 10k would archive/truncate
+    # the very result this tool promises to enumerate.
+    output_limit = 50_000
```
Nobody-Zhang: **correctness — the limit was raised on the one tool that did not need it, and left on the two that do**

Measured on this branch:

| tool | real payload | limit | result |
|---|---|---|---|
| `list_skills` | 22,112 chars | 50,000 (raised here) | fits — would also have fit in 30k |
| `load_skill` | up to 42,452 chars | 20,000 (inherited) | **210 of 597** documents truncated |
| `search_skills` (`limit=5`) | 101,701 chars | 20,000 (inherited) | **4 of 5 hits invisible**, cut mid-JSON |

The imported documents are ~3x the curated median (18k vs 5k chars), so `format_tool_result` now cuts the two tools that actually carry recipes. Concretely, `search_skills("rna seq differential expression deseq2")` returns 5 hits and only `bio-differential-expression-deseq2-basics` survives into the observation — the other four names never reach the model. And `load_skill("bio-workflows-liquid-biopsy-pipeline")` hands back the first 20,000 chars of a clinical protocol with its validation and caveat tail gone behind a `… [truncated]` marker.

That is the exact path the new aggregate prompt line makes primary ("For ANY bioinformatics task, search this collection before writing the pipeline") and that `prompts.py:94` tells the agent to fan out over.

Fixed in the review branch: explicit `output_limit = 50_000` on both `SearchSkillsTool` and `LoadSkillTool` (50k covers every bundled document). A deeper fix — bounding `doc` per hit in `loader.search()` so a wide search degrades by detail rather than by dropped hits — is still worth doing separately.

openai4s/skills_loader/loader.py
```diff
@@ -897,8 +926,25 @@ def system_context(self, *, only: frozenset[str] | None = None) -> str:
...
         ]
-        for s in skills.values():
-            lines.append(s.summary_line())
+        collections: dict[str, list[Skill]] = {}
+        for skill in skills.values():
+            if skill.collection and only is None:
+                collections.setdefault(skill.collection, []).append(skill)
+            else:
+                lines.append(skill.summary_line())
+        bioskills = collections.get("bioskills", [])
```
Nobody-Zhang: **correctness — any collection that is not literally `"bioskills"` disappears from the system prompt**

Line 931 buckets *every* truthy `skill.collection` into `collections`, but only one key is ever read back. Members of any other collection take the `if` branch (so they get no `summary_line()`) and are then never emitted.

Reproduced by re-tagging five discovered skills to `collection="chemskills"`: `"chemskills" in ctx.lower()` is `False`, none of the five names appear, and the prompt line count is unchanged at 40. Not one error anywhere.

This matters because the surrounding code invites it: `Skill.collection` is typed `str | None` with a docstring about "Large third-party collections" (plural), and `bundled_roots()` is written generically to take more roots. The next collection added through that door is discovered, enabled, searchable and loadable while being completely undisclosed to the model.

Fixed in the review branch: the loop now drains `sorted(collections.items())` through a `_COLLECTION_SUMMARIES` table, with a generic fallback line for a collection that has no bespoke blurb yet. Output for the bioskills-only case is byte-identical (8,902 bytes).

openai4s/server/skills.py
Nobody-Zhang, line 1: **correctness — a user Skill can still be created into a name that `discover()` will silently swallow**

The directory half of this guard only looks at `skills_dir/<slug>`, and its sibling `bundled_name_collision` only knows *declared* names. 143 of the 561 imported skills declare a name different from their directory (e.g. directory `bio-alignment-alignment-io` declares `bio-alignment-io`), so those 143 directory names pass both checks.

Reproduced with a temp data dir:

```
bundled_name_collision('bio-alignment-alignment-io')  -> None
(skills_dir / slug).is_dir()                          -> False        # create accepted
# ... user-skills/bio-alignment-alignment-io/SKILL.md written, declaring "my private helper"
loader.get('my private helper')                       -> None         # gone
```

`discover()` drops it at `loader.py:620` (`is_writable and child.name in discovered` — `discovered` is keyed by bundled *directory* name) with a bare `continue`. The user gets a success response and a Skill they can never see, list, or edit, and the name they chose now resolves to the GPTomics recipe.

Fixed in the review branch: the directory check iterates `loader.bundled_roots()` instead of just `skills_dir`, with a `getattr` fallback so loader test doubles behave as before.

scripts/bundle_contract.py
```diff
@@ -128,6 +128,7 @@ def check_sources(src: Path) -> int:
     if missing:
         raise BundleCheckError("source tree is missing: " + ", ".join(missing))
     skills = sorted(src.glob("skills/*/SKILL.md"))
+    skills.extend(sorted(src.glob("skills/bioskills/*/SKILL.md")))
```
Nobody-Zhang: **test-coverage — this makes the packaging floor unfalsifiable for the curated catalog**

The glob grew by 561 entries and `MIN_SKILLS` stayed at 20 (line 59; `release_import_smoke.py:61` hardcodes the same `< 20`). The count is now 597, so both floors clear by 28x on bioSkills alone.

A wheel, DMG or container that ships `skills/bioskills/` and drops **all 36** curated `skills/<name>/SKILL.md` directories still passes `check_sources`, `release_import_smoke`, the container smoke and the DMG verifier. Only `verify_release_artifacts.py`'s three hardcoded paths would notice, and none of them is a curated Skill other than `example_stats`. This is the same silent-pass shape `pyproject.toml` records for the missing `workflows*` package — "reported zero failures across zero workflows with exit code 0".

Fixed in the review branch: two separate floors here and in `release_import_smoke.py` — `MIN_SKILLS` against the curated glob, and a new `MIN_COLLECTION_SKILLS = 561` against the collection glob when it is present at all.

scripts/import_bioskills.py
```diff
@@ -0,0 +1,236 @@
...
+                "python module/command snippets use python3",
+                "silent curl snippets use fail-fast flags",
+                "Nextflow install-from-pipe snippets use the bioconda package",
+            ],
+        },
+        "skill_count": len(skills),
+        "skills": skills,
+        "files": files,
+    }
+    (destination / "MANIFEST.json").write_text(
```
Nobody-Zhang, line 293: **verification — 505 KB of SHA-256 that nothing ever rechecks**

I grepped every consumer of `skills/bioskills/MANIFEST.json` across `openai4s/`, `tests/`, `scripts/`, `.github/` and `harness/`. The only non-doc reference is `verify_release_artifacts.py:34`, which asserts the file **exists** in the wheel. There is no `--check` mode, and `import_collection` refuses a non-empty destination, so the tree could not be re-derived for comparison without deleting all 1,965 files first.

Meanwhile this same PR removes the tree from pre-commit and from `check_directory_readmes.py`, and `skills/bioskills/README.md:56` says the pair, LICENSE and manifest "document and **verify** their owning boundary" while the PR body claims "verified payload hashes". I checked all 1,962 rows: they match today. There is simply nothing that keeps them matching. After the first typo fix, bad merge, or line-ending rewrite of any of these 1,962 agent-instruction and executable files, the manifest asserts a digest the tree no longer has and every gate stays green.

Fixed in the review branch: `verify_collection()` + a `--check` flag that re-derives every hash and size, cross-checks the pinned commit and skill count, and flags untracked files under the collection; plus `test_the_pinned_collection_still_matches_its_manifest` in `tests/test_skills.py` so it runs in the offline suite. The README pair now names the mechanism instead of asserting the property.

scripts/import_bioskills.py
```diff
@@ -0,0 +1,236 @@
...
+
+def _frontmatter_value(lines: list[str], key: str) -> str:
+    prefix = f"{key}:"
+    for line in lines:
+        if line.startswith(prefix):
+            return line.partition(":")[2].strip()
+    return ""
+
+
+def _compatibility_rewrites(text: str) -> str:
```
Nobody-Zhang, line 64: **correctness — the rewrite filter claims more normalization than it performs, and its patterns miss the common spellings**

`text.replace("curl -s ", ...)` needs a trailing space, so `curl -sSL https://get.nextflow.io | bash` — the most common spelling of that flag trio — matches neither curl rule and would then reach the Nextflow replace with a prefix it cannot match. Same for `curl -fsS` (still present at `bio-database-access-sra-data/scripts/download_batch.sh:31`), `curl -O`, `curl -LO`, `curl -L`, and `wget -qO- … | bash`. `python -m ` also matches inside `ipython -m `.

Across all 561 skills these four rules changed **3 lines**. What ships untouched: `bio-causal-genomics-heritability-partitioning` instructs `wget .../LDAK/main/ldak6.3.linux && chmod +x` (an unpinned branch binary, download-to-execute), plus 84 `wget` lines, 140 `install_github(...)`, 13 `pip install git+https://`, and 2 `sudo apt install`. A reviewer reading `MANIFEST.json`'s `compatibility_rewrites` block concludes download-to-execute was removed; it was removed from one URL.

The blind `str.replace` also corrupted prose it did match — `bio-workflow-management-nextflow-pipelines/references/usage-guide.md:10-14` now reads:

```
# Install Nextflow (requires a Java 17+ JVM)
conda install -c bioconda nextflow

# Or via conda
conda install -c bioconda nextflow
```

Fixed in the review branch: word-boundary regexes covering `-sSL/-sL/-fsS/-sS/-s` and guarding `python` against `ipython`, plus README/docstring wording that says this is spelling normalization and not an audit. The vendored text itself is byte-exact and was not touched.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Preserve byte-exact vendored bioSkills payloads from the pinned upstream commit.
+skills/bioskills/** -whitespace
```
Nobody-Zhang: **correctness — `-whitespace` does not preserve a single byte**

```
$ git check-attr text eol -- skills/bioskills/LICENSE
skills/bioskills/LICENSE: text: unspecified
skills/bioskills/LICENSE: eol: unspecified
```

`whitespace` only controls git's whitespace *error* checks in `diff`/`apply`. `text`/`eol` stay unspecified, and this is the repo's only `.gitattributes` (no `* text=auto`), so `core.autocrlf` still applies. On a clone made by the Git-for-Windows installer (`core.autocrlf=true` by default), all 1,962 currently-LF files check out with CRLF and **every** `sha256` and `size` in the file `skills/bioskills/README.md` calls "the authoritative inventory" is wrong for the tree on disk — and nothing detects it, because nothing reads those hashes.

Fixed in the review branch: `skills/bioskills/** -text -whitespace`, which is the attribute that does what the comment promises.

scripts/import_bioskills.py
```diff
@@ -0,0 +1,236 @@
...
+                "directory": directory,
+                "name": declared_name,
+                "upstream_path": str(source_doc.parent.relative_to(source)),
+            }
+        )
+
+    # Examples and usage guides are upstream text assets too. Apply the same
+    # narrow relay/shell-safety rewrite without assuming every future asset is
+    # UTF-8 (binary fixtures, if added upstream, remain byte-identical).
+    for path in destination.rglob("*"):
```
Nobody-Zhang, line 236: **correctness — the importer's output and its manifest are platform-dependent**

Three independent sources of drift, all in the artifact whose whole job is reproducible provenance:

1. `path.read_text("utf-8")` applies universal newlines (CRLF→LF) and `path.write_text(...)` uses `newline=None` (LF→`os.linesep`). Any file containing one of the four rewrite tokens has its **entire** content re-line-ended — a whole-file change attributed to a four-token safety rewrite — and on Windows every matched file becomes CRLF.
2. `raw.splitlines()` + `"\n".join(...)` at lines 64/97 also splits on `\x0b \x0c \x1c-\x1e \x85 \u2028 \u2029` and turns each into `\n`.
3. `sorted(destination.rglob("*"))` sorts `Path` objects — `PurePath.__lt__` compares `_str_normcase`, which lower-cases and uses `\` on Windows — and `str(path.relative_to(destination))` emits OS separators, so the `files` array is both re-ordered and backslash-pathed there.

So the same pinned commit produces a different `MANIFEST.json` depending on who ran the importer.

Separately, `source.glob("*/*/SKILL.md")` is case-insensitive on macOS and case-sensitive on Linux (verified on this box: a file literally named `skill.md` is returned, and the returned `Path` reports its name as `SKILL.md`). A mis-cased upstream document is silently imported *and renamed* on a Mac, and trips `expected 561 skills … found 560` on the CI runner.

Fixed in the review branch: `newline=""` on both sides of the rewrite round-trip, and `key=lambda p: p.relative_to(destination).as_posix()` + `.as_posix()` for the manifest. The glob case-sensitivity is not fixed — flagging it for the next re-pin.

openai4s/skills_loader/loader.py
```diff
@@ -897,8 +926,25 @@ def system_context(self, *, only: frozenset[str] | None = None) -> str:
             "`glob_files` for Skill retrieval.",
             "",
         ]
-        for s in skills.values():
-            lines.append(s.summary_line())
+        collections: dict[str, list[Skill]] = {}
+        for skill in skills.values():
+            if skill.collection and only is None:
```
Nobody-Zhang: **efficiency — the collection collapse is switched off exactly on the delegation path**

`only is None` in this condition means the prompt-size fix does not apply whenever an allowlist is present — and an allowlist is present precisely on the delegated-Specialist path, which is the only reason `only=` exists (`host/skills.py:151`).

Measured:

| call | bytes |
|---|---|
| `system_context()` | 8,902 |
| `system_context(only=<561 bio names>)` | 134,759 |
| `system_context(only=<all 597 names>)` | 142,385 (~35,600 tokens) |

It scales linearly with the allowlist, so a specialist scoped to a few hundred genomics recipes gets tens of KB of summaries in its system prompt.

I did **not** change this: `skills/bioskills/README.md` documents "An explicitly scoped specialist receives the individual summaries it is allowed to load" as intended, and `test_bioskills_exact_load_search_and_scoped_prompt` pins it. Flagging it as a policy call — filtering by `only` first and then grouping unconditionally (emitting the collapsed line with the post-filter count) would keep small allowlists readable and stop large ones blowing up.

openai4s/skills_loader/loader.py
Nobody-Zhang, line 1: **efficiency — 266 KB of generated source, executed per kernel and persisted per generation, to gate 561 entries with nothing to gate**

`f"__openai4s_skill_bootstrap_manifest__ = {manifest!r}"` inlines all 597 entries, though the gate below needs only `directory -> enabled` plus the denied set.

Measured: `bootstrap_code()` **20,072 → 265,790 bytes**; host-side `compile()` 1.4 → 17.6 ms. That source is sent as a cell over the kernel JSON-per-line protocol on every start, restart, env switch and watchdog recovery (`gateway.py:3783/4210/4215/6633`); `recovery_runtime.py:629` stores it verbatim as `init_hooks`; `session_domain.py:714` copies the whole bootstrap mapping into `generation_refs` of **every cursor checkpoint** — one per cell — so a 100-cell session persists ~27 MB of byte-identical text. `bootstrap_manifest(persist=True)` additionally INSERTs a 242.9 KB row into `capability_manifests` per bootstrap, with no pruning anywhere.

And it buys nothing here: there are **0** `kernel.py` files under `skills/bioskills/`, so all 561 new entries feed a sidecar-import gate that can never fire for them.

I did **not** change this — it moves `bootstrap_code_sha256` and the durable recovery bytes, so it deserves its own PR. Embedding a `frozenset` of disabled + denied directory names plus a manifest id, and leaving the full manifest in SQLite where recovery already reads it, would cut this back to roughly its previous size.

openai4s/skills_loader/loader.py
```diff
@@ -855,6 +878,10 @@ def catalog(self, *, include_disabled: bool = False) -> list[dict]:
                 "version": s.version,
```
Nobody-Zhang, line 1075: **correctness — `collection` reaches the wire and the frozen schema, and no client reads it**

`server/skills.py:398` adds it to both catalog rows and `docs/response-schemas.json` marks it `required`, but `grep -n "collection" openai4s/server/webui/app.js` returns **nothing**.

Meanwhile `custSkills` (`app.js:9962`) still does a flat `skills.forEach(...)`: no grouping, no filter, no pagination, no virtualization, no bulk toggle. Measured, `/skills/catalog` grew **23,005 → 380,891 bytes** and the panel now builds 597 rows (~2,000 DOM nodes, 2–4 buttons each) in one synchronous loop. A user opening Customize → Skills scrolls past 561 read-only `bio-*` rows — visually identical to their own, each with a live enable toggle — to reach their own Skills, and disabling the collection means 561 individual clicks.

The field that would fix this is already on the wire. I did **not** change `app.js` (it is one of the six surgical-edit hotspots and this is feature work), but the wiring is what makes the new field worth having.

openai4s/skills_loader/loader.py
Nobody-Zhang, line 1: **correctness — `search_skills` gives the model a third-party recipe with no signal that it is third-party**

The importer stamps `origin: openai4s` on all 561 documents — the value `docs/security.md:99` and `skills_loader/versions.py:212` define as the *trusted first-party* origin — and nests the real provenance under `metadata.third_party`. But `_parse_frontmatter` deliberately ignores nested keys, so that block parses to `{"metadata": ""}` and nothing in `openai4s/` ever reads it.

And this result dict is `{name, origin, description, import, score, doc, sidecar_gate}`. `doc` is `s.doc` — the frontmatter-stripped body — so the upstream repository, commit and license never reach the model either. `catalog()` gained `collection` in this PR; `search()` did not, on the exact surface the new prompt line makes primary.

Fixed in the review branch: `"collection": s.collection` added to each search hit, so the one field that says "this recipe was downloaded, not written here" travels with the recipe. (`origin` itself is left alone — `skills/bioskills/README.md` documents it as the distribution boundary rather than authorship.)

tests/test_skills.py
```diff
@@ -348,5 +350,37 @@ def test_identifier_skill_uses_import_star_hint():
     assert s.import_hint.startswith("from example_stats.kernel import *")
 
 
+def test_bioskills_collection_is_discovered_but_prompt_is_compact():
```
Nobody-Zhang, line 353: **test-coverage — the repo's phantom-API gate never sees these 561 recipes, and 7 of them already fail it**

`tests/test_skill_recipes_are_callable.py:113` still globs `_SKILLS.glob("*/SKILL.md")`. The collection lives one directory lower, so the gate whose entire purpose is "a name the recipe calls must exist by the time the agent runs it" now covers **36 of 597** Skills — and the 94% it cannot see is the newest 94%.

I ran the gate's own `_unresolved()` over `skills/bioskills/*/`. Seven fail:

| skill | unresolved names |
|---|---|
| `bio-workflows-liquid-biopsy-pipeline` | `call_variants`, `preprocess_with_fgbio`, `run_ichorcna` |
| `bio-chip-seq-chip-deep-learning` | `encode`, `encode_dna_one_hot`, `one_hot_encode` |
| `bio-clinical-biostatistics-subgroup-analysis` | `GradientBoostingRegressor`, `LogisticRegression` |
| `bio-atac-seq-allele-specific-accessibility` | `map_snps_to_peaks` |
| `bio-atac-seq-deep-learning-atac` | `load_chrombpnet_model` |
| `bio-chemoinformatics-shape-similarity` | `ecfp_tanimoto` |
| `bio-sequence-io-paired-end-fastq` | `process_pair` |

An agent that follows the liquid-biopsy recipe dies with `NameError: call_variants` at exactly the step the recipe was written to perform. Running the gate over all 561 costs 0.29s, so cost was not the reason.

Extending it also surfaced 8 ```python blocks that do not parse: 6 are Snakemake DSL (`rule x:`, `configfile:`, `include:`) mis-fenced as python, and **2 are genuinely broken code** — `bio-crispr-screens-perturb-seq-analysis` (`'[' was never closed`) and `bio-data-visualization-circos-plots` (`':' expected after dictionary key`).

Fixed in the review branch: `_bundled_skills()` now returns curated + collection; the 7 failures and the 8 unparseable blocks are pinned in `KNOWN_COLLECTION_GAPS` / `KNOWN_UNPARSEABLE_BLOCKS` (the vendored tree is byte-exact, so they have to be carried upstream at the next re-pin rather than patched here), and a new `test_the_known_gaps_are_still_real` fails if an exemption stops applying, so a re-pin cannot inherit a stale allowance.

pyproject.toml
Nobody-Zhang, line 1: **correctness — `uv run pytest` is red on this branch, and one of the two failures is the egress guard**

I ran the offline suite on a pristine `git archive` of this PR's head (no review edits) to be sure these are yours and not mine. Both reproduce:

```
FAILED tests/test_egress_surface.py::test_no_undeclared_skill_reaches_the_network
FAILED tests/test_plan_crosswalk.py::test_no_closed_row_rests_on_evidence_that_has_since_changed
```

**1. `test_no_undeclared_skill_reaches_the_network`** — `_SKILL_EGRESS` is currently `{}`: before this PR, *zero* bundled Skills opened a socket. The import adds two that do, and neither was declared:

* `skills/bioskills/bio-database-access-geo-data/scripts/search_geo.py:32` — `urllib.request.urlopen(f'https://ftp.ncbi.nlm.nih.gov/geo/series/{prefix}/{gse}/soft/{gse}_family.soft.gz')`
* `skills/bioskills/bio-structural-biology-structure-io/scripts/download_structure.py:20` — `urllib.request.urlretrieve(f'https://files.rcsb.org/download/{pdb_id}-assembly1.cif.gz', assembly_gz)`

Both interpolate a caller-supplied identifier into the URL and go straight to `urllib`, so neither passes through `host.web_fetch`'s egress allowlist or its SSRF guard, and the second writes to a relative path from the URL. That test exists precisely so "a fourth cannot appear without anyone deciding to allow it" — and 561 recipes arrived without that decision being made.

I did **not** fix this. Adding two `_SKILL_EGRESS` entries would turn CI green in one line, but that is a security-posture decision about third-party code and it is yours to make: either declare them with the reason, or route them through `host.web_fetch`.

**2. `test_no_closed_row_rests_on_evidence_that_has_since_changed`** — rows `R1/O4S-06` and `R3/P1-04` both digest `tests/test_skill_catalog_readiness_route.py`, which this PR edits (the new `collection` assertions and the `bio-example` fixture). The digests were not re-recorded.

Fixed in the review branch: I read the change to that test first — it is additive (a fourth fixture Skill plus three `collection` assertions), so the evidence still supports both closed rows — then ran `python scripts/reaudit_crosswalk.py`, which rewrote exactly those two `evidence_digest` values and nothing else.

Worth double-checking the PR description's Verification block: it lists `uv run pytest` as run, and `capture_response_schemas.py --check` refuses to run on a failed suite, so that gate cannot have been green either.

Linmj-Judy:
已合并修改，我重新开一个pr

PR #117, opened 2026-08-24
feat(model-backend): formalize bring-up & admission; ship protein-design MCP; fix CI test bugs

Linmj-Judy:
## Summary

This PR formalizes **model backend bring-up and admission** as a framework-level contract, and ships the **first concrete adopter: an atomic protein-design MCP connector**.

Previously, "is the backend ready?" was a single confused question covering accelerator visibility, environment installation, checkpoint presence, model load, and inference correctness. This PR separates them into a five-state machine: **discovered → route selected → staged → admitted → formal**, with each transition individually verifiable.

The host surface gains two new methods:

- `host.accelerator_status()` — probes local NVIDIA hardware via `nvidia-smi` and reports configured SSH GPU routes independently. Hardware visibility no longer implies model readiness.
- `host.stage_model_asset(source_path, ...)` — an approval-gated, exact-path import that hashes bytes into the workspace, refuses symlinks and secret paths, and reports the SHA-256.

A staged asset is **not** admitted. Admission lives in a per-process `ModelAdmissionLedger` whose identity binds **namespace + operation + backend revision + checkpoint SHA-256 + execution target**. Restarting the connector invalidates the ledger; a new process must run a fresh canary with `run_mode="canary"` using the same adapter, backend revision, checkpoint digest, and execution target as the intended formal call.

The first concrete adopter is the new atomic protein-design MCP connector (`openai4s.mcp_servers.protein_design`): a stdio server with four JSON-schema tools, an isolated scientific worker for heavy dependencies, and a `run_mode` field that distinguishes canary from formal calls. It reuses the shared `ModelAdmissionLedger` so every other checkpoint-backed connector can adopt the same contract without re-implementing it.

The Agent loop's system prompt is updated to reflect the contract:

- When both local and remote candidates exist, the Agent must ask the user which execution target to use; silent local/remote choice is forbidden.
- Checkpoint acquisition is a per-operation question, not a discovery-time one. During generic capability discovery the Agent must not ask for, search for, or download checkpoints. Once a specific operation is selected and that operation actually requires a checkpoint, the Agent first asks whether the user already has the file locally; while that question is unanswered it must not search or download. If the user supplies a path, the Agent imports it under `host.stage_model_asset(...)` exact-path approval. Only if the user says they do not have it locally may the Agent request network permission, download from a user-approved source, and stage the resulting bytes the same way.
- Missing source code, an environment, or weights is a bring-up condition, not evidence that the requested model is unavailable.

Supporting plumbing:

- `Store.patch_connector` — an explicit env-patch API so the Web UI can edit connector configuration without round-tripping broker references, keeping secret names brokered and secret values never exposed or erased.
- MCP client argv persistence — introduces the `OPENAI4S_PYTHON` token and `openai4s_python_module()` helper; `argv[0]` is expanded to `sys.executable` only on spawn, so persisted connector configs stay machine-independent.

Closes: n/a (no linked issue).
Related: the bring-up framework is the host-side counterpart to the
`fc299d1 feat(benchmark): add tool bring-up verifier and workflow cases`
benchmark suite — the benchmark covers the same canary/admission contract
end-to-end against a real connector process.

## Area

- [ ] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

(`openai4s/host/` sits between Agent and Kernel — see "Risk and Reviewer Focus" below for the specific files.)

## Change Type

- [x] Feature

## Verification

Ran locally (offline suite under `tests/conftest.py`'s fake provider):

```text
uv run pytest tests/test_host_accelerators.py -v
uv run pytest tests/test_model_admission.py -v
uv run pytest tests/test_model_assets.py -v
uv run pytest tests/test_protein_design_mcp.py -v
uv run pytest tests/test_host_contract.py -v
uv run pytest tests/test_host_mcp_service.py -v
uv run pytest tests/test_mcp_client.py -v
uv run pytest tests/test_mcp_lifecycle.py -v
uv run pytest tests/test_gateway.py -v
```

(All new + affected test modules; each passed individually. The whole offline suite is run as a separate CI job.)

Manual: brought up a local daemon (`./start.sh`), ran `openai4s run "call host.accelerator_status()" -v`, confirmed the response shape and that the new `accelerators` field is present in `host_diagnostics`.

Not run on this machine:

```text
uv run mypy
uv run pytest                          # full offline suite
node tests/browser_smoke.mjs
node tests/browser_matrix.mjs
OPENAI4S_STAGE0_SELF_TEST=1 node tests/browser_stage0_acceptance.mjs
OPENAI4S_KERNEL_SANDBOX=enforce uv run python -m harness.smoke.linux_bwrap_interrupt
python3 scripts/source_secret_scan.py
```

Reason:

```text
uv run mypy and uv run pytest on the full offline suite both exceed the local 5-minute cold-start budget for mypy on this machine (CI Linux cold cache is faster). All seven browser jobs need a Playwright engine and a running daemon; neither is set up here. The container smoke needs Docker. The Stage 0 self-check, response-shape captures, directory-README lint, and secret scan are each run as separate CI jobs. The Linux+bwrap interrupt smoke needs Linux+bwrap, which this environment doesn't expose.
```

## Risk and Reviewer Focus

**Files to inspect first** (each is a hotspot the template calls out):

1. `openai4s/host/accelerators.py` — new. `nvidia-smi` parsing robustness across driver versions; fallback when the binary is missing.
2. `openai4s/host/model_admission.py` — new. Canonical-JSON key composition decides what changes between admissions. If a future field is added without re-running the audit, two actually-different identities could share a key.
3. `openai4s/mcp_servers/protein_design/schemas.py` — new. The `run_mode` enum (`canary` / `formal`) and the canary-only checkpoint-identity fields are intended to be reused by every future checkpoint-backed connector; please flag any deviation.
4. `openai4s/server/gateway.py` (`PUT/PATCH /connectors/{id}` handler) — the new path uses `Store.patch_connector` and disconnects the running connector before applying the change. The DataPro managed-connector bypass (`raise GatewayError(403, ...)`) needs to remain.
5. `openai4s/agent/loop.py` — the new system-prompt paragraphs cover (a) accelerator discovery is independent of model readiness, (b) the Agent must ask the user to pick a route, (c) checkpoint acquisition is per-operation, not discovery-time: generic discovery must not ask/search/download; once an operation is selected and the user has not supplied a path, the Agent first asks whether the file is local, then imports under `host.stage_model_asset(...)`, and only asks for network permission + user-approved download when the user says they don't have it. Review whether all four steps (selected → asked → local-or-network) are reachable from a normal prompt path.

**Backward compatibility:**

- `Store.patch_connector` is additive — existing `upsert_connector` callers are unaffected.
- The MCP client argv persistence is opt-in: only in-tree Python modules (`example_server`, `protein_design`) trigger the substitution. Existing connector configs that hold absolute Python paths are unchanged.
- The two new tools (`StageModelAssetTool`, `AcceleratorStatusTool`) appear under the `remote` tool group; existing tool-catalog consumers see only an additional entry.
- The protein-design MCP server is **not** auto-enabled by default — explicit configuration of a connector against it is required to spawn it.

**Compatibility with origin/main (post-merge `3cc28a8`):**

This branch was rebased on `origin/main` after PR #110/111 landed. The five add/add files (`openai4s/benchmark/bringup.py`, `openai4s/benchmark/steps.py`, `tests/test_benchmark_bringup.py`, `workflows/tool-bringup/{README, README_zh, workflow.json}`) were taken from upstream — those are the strict evolution of the same files my `fc299d1` introduced. No behavior in this PR depends on the older (HEAD-side) variants.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched:
  - `openai4s/server/gateway.py` — only the new `PUT/PATCH /connectors/{id}` handler and the compute-path registrations.
  - `openai4s/host_dispatch.py` — only the two new step-event entries (`accelerator_status`, `stage_model_asset`) and the new `_accelerator_status_payload` method.
  - `openai4s/store.py` — only the new `patch_connector` method.
  - `openai4s/sdk/host.py` — only the new `accelerator_status()` and `stage_model_asset()` host methods. The `_Mcp.tools(server)` signature is unchanged from `main` (positional string list); dual-shape compatibility at the dispatcher side is owned by upstream `46c5142`.
  - `openai4s/agent/loop.py` — only the new system-prompt paragraphs.

### Tests

- [x] Relevant tests added and listed above under **Verification**.
- [x] Host-RPC changes include focused contract coverage:
  - `tests/test_host_accelerators.py` covers missing-binary, parse-success, parse-error, and remote fallback paths.
  - `tests/test_model_admission.py` asserts the live-process admission identity binds namespace / operation / backend revision / SHA-256 / execution target, and that an in-process restart invalidates prior admissions.
  - `tests/test_model_assets.py` covers symlink refusal, secret-path refusal, bounded chunk copy, and SHA-256 verification.
  - `tests/test_protein_design_mcp.py` (743 lines) covers the full service: schema validation, canary-vs-formal gate, scientific-backend isolation, terminal-record parsing, and tool-level canary parse failure paths.

### Core Dependency Policy

- [x] No hard third-party import was added to the core engine, LLM client, or stdlib web server. All new code in `openai4s/host/`, `openai4s/tools/model_assets.py`, `openai4s/host_dispatch.py`, `openai4s/store.py`, `openai4s/sdk/host.py`, and the `openai4s.mcp_servers.protein_design` stdio server is stdlib-only.
- [x] No new optional science imports were added. The protein-design MCP's heavy dependencies (PyTorch, ESM, ProteinMPNN, RFdiffusion, etc.) live in a separately configured scientific worker process that the stdio server defers to over JSON-RPC frames; they are not in the daemon's import graph.
- [x] No new package dependency added to `envs/*.yml`. The scientific worker is operator-provisioned; this PR does not prescribe an environment for it.

- [x] Security-sensitive behavior is documented and unit-tested:
  - `StageModelAssetTool` refuses symlink sources and secret-path sources at the `WorkspaceToolContext` boundary (`tests/test_model_assets.py`).
  - The canary → admitted gate prevents a formal call from running against a backend whose live process has not demonstrated minimal inference with the same adapter, revision, checkpoint digest, and execution target (`tests/test_protein_design_mcp.py`).
  - `Store.patch_connector` keeps secret env names brokered: it cannot expose a secret value back to the caller, and a `remove_env` op retires the broker reference without erasing the secret value (`tests/test_connector_repository.py`).
- [x] Docs do not overstate isolation, sandboxing, privacy, or security guarantees. `docs/model-backend-bringup.md` is explicit that admission is a **process-local** ledger — restarting the backend invalidates admission, and the new process must re-canary.

### Docs

- [x] `docs/model-backend-bringup{,_zh}.md` (new) describes the framework contract: state model, user-visible workflow, framework surfaces (accelerator routing, model admission, `stage_model_asset`).
- [x] `openai4s/host/README{,_zh}.md`, `openai4s/mcp_servers/README{,_zh}.md`, `openai4s/tools/README{,_zh}.md`, `skills/README{,_zh}.md`, `tests/README{,_zh}.md` updated bilingual to reflect the new surface.
- [x] `README.md` and `README_zh.md` updated to reflect the new tool surface (count + new feature line).
- [x] `docs/skills.md`, `docs/webapp.md`, `docs/webapp-api.md`, `docs/plan-crosswalk.json` updated to cross-reference the new bring-up framework.


openai4s/server/webui/app.js
```diff
@@ -408,6 +408,18 @@ Object.assign(I18N.zh, {
   "cust.compute.preinstalledDetail": "已预装 {0} 个科学/联网包：{1}",
   "cust.compute.title": "计算",
   "cust.connectors.cmdPlaceholder": "启动命令，如 npx -y @modelcontextprotocol/server-filesystem .",
+  "cust.connectors.editTitle": "编辑连接器 — {0}",
```
Nobody-Zhang, line 411: **CI is red on all three Python versions because of these 24 i18n keys.**

`tests/test_model_protocol_menu.py` inlines *both* complete I18N dictionaries into a `node -e` script and passes it as a single argv entry. Linux caps one argument at `MAX_ARG_STRLEN` = 128 KiB, and these keys crossed it:

```
base (origin/main)  129585 bytes  ok
head (PR#117)       131349 bytes  OVER -> E2BIG
```

That is the `OSError: [Errno 7] Argument list too long: '/usr/local/bin/node'` failing all 7 tests in that module. It passes on macOS, where the per-argument limit is far higher — a textbook "green locally, red in CI" case.

The defect is really the harness: any translated string can now break an unrelated test module. Fix by loading the script from a temp file instead of argv (keeping `-e` for the `require`, so `process.argv` indices are unchanged), which removes the ceiling permanently.

openai4s/host/mcp.py
```diff
@@ -125,7 +127,23 @@ def _config(self, connector: dict) -> dict:
         # config; only the fixed managed connector receives authenticated HTTP.
         from openai4s.datapro import connector_runtime_config
 
-        return connector_runtime_config(self.store, connector)
+        config = connector_runtime_config(self.store, connector)
+        if (
+            connector.get("connector_id") == "protein-design"
```
Nobody-Zhang: **The admission gate and the workspace path root are wired to one of three spawn paths.**

This branch is the only place that sets `OPENAI4S_PROTEIN_DESIGN_REQUIRE_ADMISSION` and `OPENAI4S_PROTEIN_DESIGN_ROOT`. The two Web connector routes build the launch config directly:

- `gateway.py` `POST /connectors/{id}/probe` → `mcfg = datapro.connector_runtime_config(store, c)`
- `gateway.py` `POST /connectors/{id}/call` → same

So `POST /connectors/protein-design/call` with `run_mode` omitted (which defaults to `"formal"`) starts the server with `require_admission` False — `service.py`'s `if self.require_admission and run_mode == "formal"` never fires — and with `root` falling back to `os.getcwd()`, the daemon's launch directory, as the authority for every user-supplied `checkpoint_path`/`output_dir`. Because `cache_scope` also differs, this ungated child is a *second* cached process running alongside the workspace-scoped one.

The terminal record it writes says `status: "succeeded"` with no `bringup_admission` key, which a reader cannot distinguish from an admitted run.

This is the repo's own documented "mechanism exists, wiring partial" shape. Suggest factoring the confinement into a shared `confine_bundled_connector(config, connector, workspace=...)` that every spawn path calls; the admission requirement is not workspace-dependent, so the Web routes can pass a stable daemon-owned root rather than inheriting `cwd`.

openai4s/tools/model_assets.py
```diff
@@ -0,0 +1,130 @@
...
+    def secret_path(self, arguments: Any) -> str | None:
+        # The caller names a *source*, while the destination is derived. Keep
+        # the hard secret-file refusal without pretending the caller controls
+        # where the copy is written.
+        if not isinstance(arguments, dict):
+            return None
+        source = str(arguments.get("source_path") or "")
+        from openai4s.host.files import is_secret_path
+
+        return source if is_secret_path(source) else None
```
Nobody-Zhang: **The secret denylist and the approval prompt read the caller's raw string; the copy reads the resolved path.**

`secret_path()` and `permission_target()` both see `arguments["source_path"]` verbatim, while `execute()` copies `Path(expanduser(raw)).resolve()`. `is_secret_path` is a *lexical*, directory-aware denylist, so it can only classify the segments it is handed — and the `is_symlink()` check below covers the final component only.

Demonstrated on this checkout:

```
lexical is_secret_path("~/link/config")  -> False      # link -> ~/.ssh
tool.secret_path(...)                    -> None       # gate never consulted
=> ~/.ssh/config copied into <workspace>/model-assets/config, SHA-256 returned
```

Basename `config` is not a secret basename and segments `('~','link','config')` contain no secret directory, so the hard refusal never fires. The approval prompt shows the innocuous `~/link/config` too, so a durable approval recorded against that string keeps authorizing whatever the name points at later.

Classify the resolved destination in both `secret_path()` and `permission_target()`. (Refusing symlinked *ancestors* outright is not the fix — `/tmp` and `/var` are OS-level symlinks on macOS, so that rejects every ordinary download location.)

Separately worth a look: `execute()` hand-rolls staging with `NamedTemporaryFile` + `os.replace` on a raw `Path`, bypassing the `WorkspaceToolContext.secure_parent()` / `create_staged()` / `publish()` pinned-dirfd path that `write_file`, `edit_file` and `web_download` all publish through — so the `O_NOFOLLOW|O_EXCL` staging, hardlink refusal and secret-alias check don't apply to the derived destination either.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+class ProteinDesignService:
+    """Validate calls, launch pinned backends and always persist terminal records."""
+
+    def __init__(self, *, root: str | Path | None = None, timeout: float | None = None):
+        configured_root = (
+            root or os.environ.get("OPENAI4S_PROTEIN_DESIGN_ROOT") or os.getcwd()
+        )
+        self.root = Path(configured_root).resolve()
+        self.timeout = float(
+            timeout or os.environ.get("OPENAI4S_PROTEIN_DESIGN_TIMEOUT_S", "7200")
```
Nobody-Zhang, line 315: **The backend timeout (7200 s) is an order of magnitude past the deadline the transport will actually wait.**

`MCPManager._connect` builds a stdio `MCPConnection` with no `timeout`, so it takes `_deadline_default()`:

```python
DEFAULT_TIMEOUT_S = 60.0
MAX_TIMEOUT_S = 600.0    # hard clamp, even via OPENAI4S_MCP_DEADLINE_S
```

So a real `generate_backbone` or `predict_structure` — minutes to hours of GPU work, which is the connector's headline use case — cannot complete over this transport. At 60 s (600 s at most) the caller raises `MCPTimeout`, `_with_connection` evicts and terminates the server, the GPU child is orphaned, and no `terminal.json` is written. The attempt directory is then left with `stdout.log`/`resolved_config.json` and no terminal record, so the `partial_files` check permanently poisons that `attempt_id` as `InterruptedAttempt` — and the process-scoped admission ledger dies with the killed process, so the next call demands a fresh canary that will time out the same way.

The offline tests don't see this: they construct the service with `timeout=5` and fake fast executables.

This needs a product decision rather than a patch — either long-running tools become submit/poll (returning an attempt handle immediately), or the stdio transport grows a per-tool deadline above `MAX_TIMEOUT_S`. Flagging rather than fixing.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+                        "runtime_seconds": round(time.monotonic() - start, 6),
+                        "error_type": "AttemptManifestUnreadable",
+                        "error": (
+                            "existing terminal manifest is unreadable; manual "
+                            "reconciliation is required and the file was preserved"
+                        ),
+                        "config_digest": requested_digest,
+                        "manifest_path": str(manifest),
+                    }
+                if (
```
Nobody-Zhang, line 701: **The terminal-manifest replay short-circuit bypasses the ledger in both directions.**

This returns before the admission block, so it neither calls `require()` nor `admit()`.

*Re-asserts what a restart is supposed to revoke:* with `REQUIRE_ADMISSION=1`, canary → admit → formal attempt `f1` succeeds and writes `terminal.json`. Restart the child (any connector edit calls `disconnect`; so does a daemon restart), which empties the ledger *by design*. Re-issue the identical `f1` call: this returns the stored record with `"bringup_admission": {"status": "verified"}` and `isError: false` — asserting live-process admission the ledger no longer holds, which `docs/model-backend-bringup.md` says a restart must revoke.

*And refuses what should pass:* replay the *canary* attempt id after the same restart and you get `status: succeeded` with `bringup_admission.status == "verified"`, but `admit()` never ran, so the ledger stays empty and every later formal call fails "not admitted in this live process". An agent following the documented loop (canary → verify terminal output → retry formal) loops forever; only a brand-new `attempt_id` escapes.

Third issue on the same branch: a stored **failed** record is replayed verbatim with no marker. Fix the config, re-issue identical args, and you get the old `error` and old `started_at` back without anything re-running — the agent reads a fixed configuration as still broken.

Suggest re-deriving admission from the live ledger on the replay path and adding a `"replayed": true` field so a caller can tell that nothing executed.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+        self, args: dict[str, Any], output: Path
+    ) -> dict[str, Any]:
+        if args.get("num_designs", 1) != 1:
+            raise DesignToolError(
+                "generate_backbone is one attempt per call; num_designs must equal 1"
+            )
+        target = self._path(args.get("target_pdb"), existing=True)
+        checkpoint, checkpoint_digest = self._verify_checkpoint(args)
+        repository = self._path(
+            os.environ.get("OPENAI4S_RFDIFFUSION_PATH", "vendor/RFdiffusion")
```
Nobody-Zhang: **Operator-configured backend paths go through the agent's path confinement, so every correct install is rejected.**

```python
repository = self._path(
    os.environ.get("OPENAI4S_RFDIFFUSION_PATH", "vendor/RFdiffusion")
)
```

`_path()` enforces `candidate.relative_to(self.root)`, and `host/mcp.py` now pins `self.root` to the per-session workspace. So the documented deployment — `OPENAI4S_RFDIFFUSION_PATH=/opt/RFdiffusion` — raises `path escapes configured protein-design root: /opt/RFdiffusion` before any other validation, and *even when* `OPENAI4S_RFDIFFUSION_COMMAND` is set so the repository is only used for the revision probe. Same at the `OPENAI4S_PROTEINMPNN_PATH` site. The message names neither the variable nor the root.

That confinement exists to keep an *agent-supplied* path inside the session root; an operator's model checkout legitimately lives outside it. This is a refusal aimed at the wrong party.

The same root binding hits user checkpoints: `_verify_checkpoint` calls `self._path(checkpoint_path, existing=True)`, so a user answering the skill's own question with `/data/models/Complex_base_ckpt.pt` fails too — and `SKILL.md` explicitly says not to "make them download a second copy merely because it is outside a conventional directory", while the only workaround (`host.stage_model_asset`) copies the multi-GB file into the workspace once *per session*.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+        if header is not None:
+            records.append((header, "".join(chunks)))
+        if len(records) != expected + 1:
+            raise DesignToolError(
+                "ProteinMPNN FASTA must contain one native record plus exactly the requested designs"
+            )
+        validated: list[dict[str, Any]] = []
+        for index, (header, combined) in enumerate(records[1 : expected + 1]):
+            pieces = combined.split("/")
+            match_designed = re.search(r"designed_chains=(\[[^\]]*\])", header)
```
Nobody-Zhang: **`_validate_mpnn_fasta` requires a chain mapping in every *sampled* record header, but ProteinMPNN writes it only in the first (native) record.**

The loop runs over `records[1 : expected + 1]` — the design records — and raises unless each header matches both `designed_chains=` and `fixed_chains=`.

Upstream `protein_mpnn_run.py` emits:

```
>3HTN, score=..., global_score=..., fixed_chains=['B'], designed_chains=['A'], model_name=v_48_020   <- native, record 0
>T=0.1, sample=1, score=..., global_score=..., seq_recovery=...                                      <- design records
```

The design headers carry no chain mapping, so every genuine ProteinMPNN run would raise `DesignToolError("ProteinMPNN FASTA header lacks designed/fixed chain mapping")` and be recorded as a **failed** attempt.

The offline test passes only because `tests/test_protein_design_mcp.py` fabricates a design header that carries the mapping — freezing a format the real tool does not emit. Worth checking against your actual ProteinMPNN revision; if confirmed, read the mapping from record 0 and apply it to the sampled records.

Not fixed here — I can't run the real backend to confirm the header format, and guessing would change intended behaviour.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+                {
+                    "HF_HUB_OFFLINE": "1",
+                    "TRANSFORMERS_OFFLINE": "1",
+                    "WANDB_MODE": "offline",
+                }
+            )
+        started = time.monotonic()
+        with stdout_path.open("wb") as stdout, stderr_path.open("wb") as stderr:
+            try:
+                completed = subprocess.run(
```
Nobody-Zhang, line 588: **The backend subprocess inherits this process's stdin, which is the MCP protocol pipe.**

`subprocess.run(...)` redirects stdout and stderr to log files but leaves fd 0 inherited, while `server.main()` reads MCP requests from `sys.stdin.buffer`.

A backend that reads stdin — an interactive prompt, a `docker run` invoked without `-i` semantics assumed, a wrapper that drains stdin, a worker that hits `input()` — consumes bytes the server still needed, desynchronising the JSON-RPC stream for the rest of the process lifetime. A backend that *blocks* on stdin hangs until the 7200 s timeout.

```suggestion
                completed = subprocess.run(
                    command,
                    cwd=str(cwd or self.root),
                    env=env,
                    stdin=subprocess.DEVNULL,
                    stdout=stdout,
                    stderr=stderr,
                    timeout=self.timeout,
                    check=False,
                )
```

tests/test_mcp_lifecycle.py
```diff
@@ -605,6 +605,9 @@ def test_abandoned_ids_and_in_flight_requests_are_both_capped(tmp_path, monkeypa
     config = _server(tmp_path, "wedged", body=_HANG)
 
     connection = MCPConnection(config["command"], timeout=30.0)
+    # The cap is exercised by short request deadlines, not by requiring a
+    # freshly spawned Python process to initialize within 200 milliseconds.
+    connection._timeout = 0.2
```
Nobody-Zhang, line 610: **CI is also red on `test_plan_crosswalk.py` because this file changed without re-recording its crosswalk evidence.**

```
AssertionError: 2 closed row(s) rest on evidence that has changed since they were
audited: ['R4/P0-2', 'R5/P0-01'].
```

Both rows list `tests/test_mcp_lifecycle.py,tests/test_kernel_oversized_frame.py` as their evidence, and editing this file changes the digest they were closed against. Reproduces locally on the PR head.

Fix is mechanical — `uv run python scripts/reaudit_crosswalk.py` — but it is worth reading the two rows against their tests first, since that is exactly what the digest is asking you to confirm.

Note this failure also takes the **Frozen response shapes** job down with it: `capture_response_schemas.py` refuses to run on a red suite, so one unrelated red test costs you the contract gate as well.

openai4s/mcp_client.py
```diff
@@ -843,6 +852,11 @@ def _argv(config: dict) -> list[str]:
             argv = cmd.split() + list(args)
         else:
             raise MCPError("connector has no command")
+        in_tree_module = (
+            len(argv) >= 3 and argv[1] == "-m" and argv[2] in _IN_TREE_PYTHON_MODULES
+        )
+        if argv and (argv[0] == OPENAI4S_PYTHON or in_tree_module):
```
Nobody-Zhang: **`argv[0]` is overwritten for *any* command whose argv[1:3] matches an in-tree module, silently discarding an operator's chosen interpreter.**

The `in_tree_module` arm fires regardless of what `argv[0]` was — it is not a token expansion, it is an unconditional override, and it is not scoped to the bundled connector ids that `_migrate_builtin_connector_commands` is careful to restrict itself to.

An operator who configures the connector as `["/opt/conda/envs/protein/bin/python", "-m", "openai4s.mcp_servers.protein_design"]` — the obvious choice, since that env carries pyrosetta/openmm/esm — gets the daemon's zero-dependency `.venv` instead. The new connector editor keeps displaying the conda path in a field labelled "Launch command", so the configuration shown and the configuration executed disagree with nothing surfacing the substitution.

The boot migration makes it durable while its docstring claims the opposite ("arbitrary custom connector commands remain byte-for-byte untouched"): its match is `command[1:] == portable[1:]`, i.e. exactly the custom-interpreter case, so the row is rewritten and the operator's original interpreter is gone from the DB.

Suggest keeping only the explicit `argv[0] == OPENAI4S_PYTHON` check, and narrowing the migration to rows whose interpreter is this daemon's own `sys.executable` or no longer exists. Failing loudly with `ModuleNotFoundError` beats silently running the wrong environment.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+    return value
+
+
+def _network_isolation_prefix() -> list[str]:
+    """Accept only command prefixes that visibly request a separate net namespace."""
+    prefix = _parse_json_array_env("OPENAI4S_PROTEIN_DESIGN_OFFLINE_PREFIX")
+    if not prefix:
+        return []
+    executable = Path(prefix[0]).name
+    flags = set(prefix[1:])
```
Nobody-Zhang, line 256: **`network_isolation_enforced: true` is attested from a token scan that last-flag-wins defeats.**

`flags = set(prefix[1:])` proves a token *appears*, not that it is *in effect*. bwrap applies options in order and Docker takes the last `--network`, so a prefix can name the isolating flag and still run fully networked — while `resolved_config.json` records `"network_policy": "os-isolated"` and the terminal record records `"network_isolation_enforced": true`.

Verified on this checkout, before the fix:

```
["bwrap","--unshare-net","--share-net",...]              ACCEPTED
["docker","run","--net=none","--network=host",...]       ACCEPTED
["docker","run","--network","none","--network","host"]   ACCEPTED
```

All three produce a blind-prediction record asserting the one claim this whole gate exists to make. Per the repo's own rule, docs and records must not overstate isolation.

After rejecting prefixes that also carry a re-enabling flag, honest prefixes still pass and all three above are refused.

Worth noting the altitude too: `openai4s/security/sandbox.py` already *builds* this wrapper (`wrap_bwrap_command` appends `--unshare-net`, `wrap_seatbelt_command` denies network) with `auto|enforce|off` and a real self-test. A syntactic allowlist over an operator-supplied argv is a weaker mechanism than the one the repo already has.

openai4s/mcp_servers/protein_design/server.py
```diff
@@ -0,0 +1,95 @@
...
+        if len(raw) > _MAX_FRAME_BYTES:
+            sys.stderr.write("ignored oversized MCP request frame\n")
+            continue
+        try:
+            message = json.loads(raw)
+            if not isinstance(message, dict):
+                continue
+        except (UnicodeDecodeError, json.JSONDecodeError):
+            continue
+        reply = handle(message, service)
```
Nobody-Zhang: **One failed manifest write kills the connector process mid-session, with no error frame.**

`ProteinDesignService.call`'s `except Exception` ends before the terminal-manifest write, so `_atomic_json(manifest, record)` runs unguarded. `handle()` has no try/except around `service.call`, and this `for raw in sys.stdin.buffer` loop has none either.

ENOSPC during a ColabFold run that writes GBs (or a read-only remount, or a permissions change) therefore raises `OSError` that unwinds all the way out: the interpreter exits with a traceback on stderr, the client is still blocked in `_read_reply` and gets no error frame, and it waits out the full deadline before reporting a timeout rather than the real disk error. The attempt is left with partial files and no `terminal.json`, which permanently poisons that `attempt_id` via the `InterruptedAttempt` check — and the restarted process has an empty admission ledger, so every formal call fails until a fresh canary runs.

The docstring calls terminal records "part of the public contract"; the contract needs the loop to answer the request. Suggest wrapping `handle()` and returning a `-32603` error frame.

openai4s/mcp_servers/protein_design/service.py
```diff
@@ -0,0 +1,1336 @@
...
+            prior_admission = self._admission.get(admission_key)
+            if self.require_admission and run_mode == "formal":
+                try:
+                    prior_admission = self._admission.require(admission_key)
+                except ModelAdmissionError as error:
+                    raise DesignToolError(str(error)) from error
+            handler = getattr(self, f"_tool_{name}")
+            payload = handler(args, output)
+            record = self._base_record(name, args, attempt, seed, started_at, start)
+            record.update({"status": "succeeded", **payload})
```
Nobody-Zhang, line 775: **Spreading `payload` last lets the least-trusted component overwrite the record's own audit identity.**

`_base_record` sets `schema_version`, `tool`, `attempt_id`, `seed`, `backend_revision`, `config_digest`, `runtime_seconds`. Because `**payload` comes after them — and after the literal `"status": "succeeded"` — payload wins for every colliding key. For the five scientific tools `payload` is `{**run, **result, ...}` where `result` is arbitrary JSON read back from `worker_result.json`, screened only by a one-key blocklist (`{"interface_hbonds"}`).

**Live today:** `run` carries `runtime_seconds` = the backend subprocess's wall time, so it overwrites the attempt's own value on every success — the same field name means the backend's span on a succeeded record and the whole attempt's span on a failed one, and checkpoint-hashing time (minutes for a ColabFold data dir) silently vanishes from the successful record.

**Latent:** a worker result key named `config_digest` or `status` rewrites the replay key this same function compares above, turning an idempotent re-call into a permanent `AttemptConflict`.

Suggest re-applying the base identity fields after the payload spread, and renaming `_run`'s field to `backend_runtime_seconds` so the two spans have two names.

openai4s/server/gateway.py
```diff
@@ -1999,20 +2000,31 @@ def _maybe_call(v):
...
-    """Prompt fragment reflecting the current remote-GPU registry.
+    """Prompt fragment reflecting local hardware and the remote-GPU registry.
 
     Sessions can be created before the user adds a GPU in Settings, so this
     context is injected both into the initial system prompt and into later turns.
     """
+    try:
+        from openai4s.host.accelerators import LocalAcceleratorService
+
+        local = LocalAcceleratorService().status()
```
Nobody-Zhang: **An uncached `nvidia-smi` now runs on every user turn, above the early return that discards its result.**

The pre-PR function returned `""` immediately when the compute registry was empty; the new early-out sits ~86 lines *below* this probe. `_remote_gpu_runtime_context` is called per user message and again on session-prompt assembly, and `host_dispatch._m_capabilities` adds the same probe to `host.capabilities()`. There is no memoization and no TTL.

On a GPU host whose driver is wedged — a common state after a GPU fault or with a stale kernel module — every message blocks the request thread for the full 6 s timeout before the LLM call is even assembled. GPU inventory is daemon-lifetime-stable, so re-probing per turn buys nothing.

Two related problems in the same function:

- The `else` branch renders **any** non-available result as `"no usable local GPU observed by the daemon"`, including a probe that timed out or crashed. On a real GPU host with a wedged driver the system prompt tells the model there is no local GPU — a wrong claim rather than an absent one, which is precisely the inversion `accelerators.py`'s own docstring warns against. A failed probe should say the hardware is *unknown*.
- `_m_capabilities` now reads the remote registry twice: once directly, and again inside `_accelerator_status_payload` (whose `AcceleratorRoutingService` was wired with `remote_status=self._remote_gpu_status_payload`).

Suggest probing after the early-return gate and memoizing with a short TTL.

openai4s/host/accelerators.py
```diff
@@ -0,0 +1,189 @@
...
+                [
+                    binary,
+                    "--query-gpu=index,name,memory.total,driver_version",
+                    "--format=csv,noheader",
+                ],
+                capture_output=True,
+                text=True,
+                timeout=6,
+            )
+        except (OSError, subprocess.TimeoutExpired) as error:
```
Nobody-Zhang: **Narrowing the catch-all means `GET /compute/gpu` can now return 500 where it previously could not fail.**

The `_detect_gpu` this replaced wrapped the whole probe in `except Exception: pass` and fell through to the CPU-only dict. This catches only `(OSError, subprocess.TimeoutExpired)`.

`subprocess.run(..., text=True)` decodes with the locale codec, so a non-ASCII GPU or driver string under `LANG=C` raises `UnicodeDecodeError` — a `ValueError`, not an `OSError`. That escapes into `_detect_gpu`, whose route (`GET /compute/gpu`) does not wrap it, and answers 500 instead of the documented `available: false` body.

The same uncaught path reaches `host.capabilities()` via `AcceleratorRoutingService.status()` → `self._local_status()`, which is also unguarded — so one failed local probe takes down the *entire* capabilities reply, including `llm`, `query`, `skills`, and the SSH `remote_gpu` inventory that has no dependency on the local probe at all.

One more shape note a few lines down: `"index": int(fields[0]) if fields[0].isdigit() else fields[0]` emits an `int` on a healthy host and a `str` on a MIG-partitioned one (`0:1`) or some vGPU drivers (`[N/A]`). The comment deleted from `_detect_gpu` documented this exact class of bug for the response *key set*; this reintroduces it as *type* variance in the same payload. Suggest always a string.

openai4s/store.py
```diff
@@ -5192,6 +5193,80 @@ def upsert_connector(
...
+            not isinstance(item, str) or not item or "=" in item or "\x00" in item
+            for item in removals
+        ):
+            raise ValueError("remove_env must contain valid environment names")
+        if set(updates) & set(removals):
+            raise ValueError("an env name cannot be updated and removed together")
+
+        previous_env = current.get("env")
+        merged_env = dict(previous_env) if isinstance(previous_env, dict) else {}
+        brokered = broker_connector_env(self, connector_id, updates)
```
Nobody-Zhang, line 5249: **The new PUT route can move another connector's secret reference into a connector whose command it also controls.**

`broker_connector_env` short-circuits on values that already look like references:

```python
if not text or is_ref(text):
    out[var] = text
    continue
```

A patch carries *new* values only — absent names keep their existing references — so a `secret://`-shaped value in `env_updates` did not come from us. `PUT /connectors/{id}` passes the browser's `env_updates` straight in, so a body like

```json
{"command": [...], "env_updates": {"X": "secret://v2/<ns>/connector_env/other-connector.API_TOKEN"}}
```

persists another connector's reference here, and `resolve_connector_env` resolves it at spawn into whatever command the same request set. The docstring says this method exists because "Public connector projections contain env names but never values" — yet the route can relocate a value it cannot read.

Suggest rejecting reference-shaped values in `env_updates`.

Two smaller things on this method while you're here: `env_updates` name/value validation (non-empty, no `=`, no NUL) is a second copy of the rules `_connector_environment` enforces at spawn time, so the persisted set and the spawnable set can drift apart; and the retired-reference cleanup open-codes `forget_connector_env` from `storage/connectors.py`. Both would sit better next to `broker_connector_env` in the repository than in the `Store` facade.

PR #118, opened 2026-08-24
Add executable retrosynthesis scenario protocols

wangyu-sd:
## Summary

- define six independent retrosynthesis scientific scenarios with explicit public/private data boundaries and evaluator contracts
- add executable benchmark implementations for single-step prediction, multistep planning, atom mapping, forward prediction, condition recommendation, and yield estimation
- add isolated deployment/worker adapters and reproducible manifests for AiZynthFinder, RXNMapper, ReactionT5v2, and Parrot; preserve explicit blocked or quarantined states when licenses, artifacts, or validation are insufficient
- add a cross-model route-admission gate: stock closure is not treated as reaction feasibility, and unsupported critical steps fail the whole route
- add a deterministic reproducibility ZIP builder that rejects local paths, symlinks, binaries, oversized files, and caller-declared sensitive fragments

## Validation

- rebased onto `origin/main` at `329fb8c2`; branch head `f831506f`
- `uv run pytest tests/test_retrosynthesis_planning.py tests/test_retrosynthesis_scenario_protocols.py -q -ra` (passed; two optional RDKit checks skipped in the zero-dependency core environment)
- `uv run python scripts/check_directory_readmes.py` (passed for 120 directories and 1057 files)
- `uv run python -m harness.cli run --tier pr --offline` (38/38 passed before and after rebase)
- `python scripts/source_secret_scan.py` (passed for 3303 files)
- changed-file pre-commit hooks passed, including isort, black, ruff, and strict-boundary mypy
- the deterministic ZIP unit test passes; the generated 27-file Warfarin reproducibility bundle passes archive integrity and path/content de-identification scans

The earlier complete GitHub PR CI passed on Python 3.10, 3.12, and 3.13, including pre-commit, packaging, container, browser, response-contract, Harness, and CodeQL jobs. The rebased head will receive a fresh CI run after push.

## External backend checks

- AiZynthFinder 4.4.1, RXNMapper 0.4.3, ReactionT5v2-forward, and the reviewed Parrot snapshot have completed real worker-level canaries in isolated external environments
- an end-to-end Warfarin run generated 232 routes (24 stock-closed), then froze one three-step route for mapping, condition, and forward validation
- mapping completed for all three fixed steps; forward exact-match ranks were 3, no match, and 1
- a bounded no-condition plus five-Parrot-beam diagnostic still failed the unsupported middle step, so the route is correctly reported as a partial failure rather than experimentally ready
- ReactionT5v2-yield remains quarantined because its published canary is not reproduced; no yield score is used to rescue route admission
- RetroChimera remains blocked where official artifact availability or license evidence is insufficient

These are engineering canaries, not frozen scientific benchmark accuracy claims. Frozen public/private benchmark datasets are still required before reporting Top-K, route, mapping, condition, or OOD yield performance.

## Safety and reproducibility

- repository examples and public artifacts no longer contain operator usernames, host mount paths, environment prefixes, or temporary directories
- no model checkpoint, external environment, credential, stock database, or generated validation artifact is committed
- the external reproducibility ZIP contains path-free source, configuration, requirements, provenance manifests, bounded results, and internal SHA256 checksums; it intentionally excludes binary environments and checkpoints
- external model execution remains isolated from the zero-dependency core
- compatibility/composition facades were not rewritten


Nobody-Zhang:
@cursoragent review this PR.

Nobody-Zhang:
@cursoragent  Review this PR. Newer Version.

skills/retrosynthesis_planning/multistep_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+                visiting.remove(identity)
+                return False
+            if not children:
+                solved = smiles in stock
+                if not solved:
+                    unresolved.append(smiles)
+            else:
+                if any(_kind(child) != "reaction" for child in children):
+                    issues.append("molecule_has_nonreaction_child")
+                solved = any(visit(child, depth) for child in children)
```
Nobody-Zhang: **A route containing no reaction node at all verifies as `solved` and `valid`.**

`verify_route_tree` walks a molecule node's children regardless of their kind, and `molecule_has_nonreaction_child` is deliberately excluded from the invalidating set at lines 173-181. Reproduced:

```python
verify_route_tree(
    {"tree": {"type": "mol", "smiles": "TARGET",
              "children": [{"type": "mol", "smiles": "BUY"}]}},
    target_smiles="TARGET", stock=frozenset({"BUY"}), canonicalizer=lambda s: s)
# -> {'verified_solved': True, 'valid': True, 'reaction_count': 0,
#     'unresolved_leaves': [], 'issues': ['molecule_has_nonreaction_child']}
```

`_kind()` here also lacks the `.strip()` that `route_review._node_kind` (line 22) has, so `type="reaction "` is classified `molecule` and the reaction **AND**-branch silently becomes an **OR**-branch — a route whose precursor `B` is not in stock then reports solved with an empty unresolved-leaf list and `reaction_count: 0`.

`evaluate_routes` reads only `verification["verified_solved"]` (line 315) and never `valid` or `issues`, so `solved_target_rate` reaches 1.0 for routes proposing no chemistry. This is exactly the "stock closure is not treated as reaction feasibility" the PR summary claims to enforce, and the module docstring's "a backend `solved=true` flag is never trusted".

Suggested: add both `molecule_has_nonreaction_child` and `reaction_has_nonmolecule_child` to the invalidating set, require at least one reaction node on any solved path, and give `_kind` the same `.strip()` as `_node_kind`.

skills/retrosynthesis_planning/reaction_model_deployment.py
```diff
@@ -0,0 +1,508 @@
...
+        raise ReactionModelDeploymentError("unsupported artifact snapshot schema")
+    unhashed = {"schema_version": 1, "files": manifest["files"]}
+    fingerprint = hashlib.sha256(
+        json.dumps(unhashed, sort_keys=True, separators=(",", ":")).encode("utf-8")
+    ).hexdigest()
+    if fingerprint != manifest["snapshot_sha256"]:
+        raise ReactionModelDeploymentError("artifact snapshot fingerprint mismatch")
+    root = Path(base).resolve()
+    failures: list[str] = []
+    for entry in manifest["files"]:
```
Nobody-Zhang, line 407: **`verify_artifact_snapshot` cannot detect a file *added* to a reviewed artifact tree.**

The loop iterates only the manifest's recorded entries; nothing enumerates the tree under `base` to compare the file *set*. Reproduced:

```
baseline:              {'ok': True, 'failures': [], 'verified_files': 2}
after two ADDED files: {'ok': True, 'failures': [], 'verified_files': 2}
```

This is directly exploitable downstream, because two code paths in this same PR load *whatever is in the directory*:

- `parrot_mar_inference.py:45-49` — `sys.path.insert(0, str(model_dir))` then `from model import ...`, so an added `model.py` executes.
- `reaction_model_worker._predict_products` — `AutoModelForSeq2SeqLM.from_pretrained(location)`, where `transformers` prefers an added `model.safetensors` over the verified `pytorch_model.bin`.

`SKILL.md` makes this the only gate between a reviewed checkpoint tree and inference ("verify that manifest before each benchmark"), so a "verified" snapshot can load weights and code the manifest never covered.

Suggested: walk `base` (or each snapshotted root) and fail on any file present on disk but absent from the manifest, reporting it as an `unexpected_file` failure.

skills/retrosynthesis_planning/reaction_model_deployment.py
```diff
@@ -0,0 +1,508 @@
...
+    for raw in paths:
+        path = Path(raw).resolve()
+        try:
+            path.relative_to(root)
+        except ValueError as exc:
+            raise ReactionModelDeploymentError(
+                f"artifact {path} is outside snapshot base {root}"
+            ) from exc
+        candidates = (
+            sorted(item for item in path.rglob("*") if item.is_file())
```
Nobody-Zhang, line 356: **Symlinks inside the tree bypass the base-containment guard eight lines above.**

Lines 348-354 correctly refuse any top-level argument that resolves outside `base`. That guard never reaches the `rglob`'d candidates: they are not resolved, and both `item.is_file()` and `_sha256_file` follow symlinks. So a symlink *inside* the tree smuggles in precisely the file the guard just rejected. Reproduced with `base/models/weights.bin -> /tmp/outside/weights.bin`:

```
direct outside path REFUSED: artifact /tmp/outside/weights.bin is outside snapshot base /tmp/base
snapshot files: [..., {"path": "models/weights.bin", "bytes": 15, "sha256": "472339b2fe60..."}]
verify: {'ok': True, 'failures': [], 'verified_files': 2}
```

The manifest claims to describe the reviewed base tree, but those bytes live outside it and can be swapped without touching anything under `base`; verification on a second machine answers about entirely different bytes.

Worth noting the sibling added in this same PR takes the opposite, strict stance — `reproducibility_bundle.py:47-48` hard-fails on *any* symlink, and the README advertises that strictness. The permissive one is the supply-chain gate.

Suggested: `candidate.resolve()` and re-apply the same `relative_to(root)` containment check per file, or refuse symlinks outright as the sibling does.

skills/retrosynthesis_planning/condition_benchmark.py
```diff
@@ -0,0 +1,283 @@
...
+    require_exact_fields,
+    require_text,
+    sha256_json,
+)
+from .single_step_benchmark import normalize_precursor_set, rdkit_canonicalize
+
+Canonicalizer = Callable[[str], str]
+SLOTS = ("catalyst1", "solvent1", "solvent2", "reagent1", "reagent2")
+INPUT_FIELDS = frozenset({"reaction_id", "reactants", "product"})
+PAYLOAD_FIELDS = frozenset({"reaction_id", "predictions", "error"})
```
Nobody-Zhang: **Scenario 5 cannot be run against the Parrot backend this same PR adds.**

`reaction_model_worker._recommend_conditions` (line 635-642) emits records with four keys — `reaction_id`, `predictions`, `raw_output`, `error` — but `PAYLOAD_FIELDS` omits `raw_output` and `normalize_condition_outputs` hard-rejects any extra key (`or fields - PAYLOAD_FIELDS`, line 112-116).

Following `skills/reaction-condition-recommendation/SKILL.md` exactly — invoke the snapshot through `ReactionModelBackend("parrot", ...)`, then normalize with `condition_benchmark` — aborts on the first record:

```
BenchmarkProtocolError: condition payload 1 violates schema
```

after the entire GPU inference run has already been paid for. The only way past it is hand-editing model output, which is what the frozen-trajectory hash exists to prevent.

No test crosses this seam: `test_parrot_worker_parses_all_ranked_condition_beams` stops at `response["result"]["records"][0]["predictions"]`, and the condition-protocol tests hand-build payloads without `raw_output`.

Suggested: add `"raw_output"` to `PAYLOAD_FIELDS` (it is genuinely useful provenance) or drop it from the worker record, and add one test that feeds a real worker record straight into the normalizer.

skills/retrosynthesis_planning/atom_mapping_benchmark.py
```diff
@@ -0,0 +1,362 @@
...
+    finite_number,
+    json_copy,
+    require_exact_fields,
+    require_text,
+    sha256_json,
+)
+
+REACTION_FIELDS = frozenset({"reaction_id", "reaction_smiles"})
+PREDICTION_FIELDS = frozenset(
+    {"reaction_id", "mapped_reaction", "confidence", "atom_correspondence"}
```
Nobody-Zhang: **Scenario 3 cannot be run against the RXNMapper backend this same PR adds.**

`reaction_model_worker._map_reactions` (lines 191-200) always emits an `error` key, but `PREDICTION_FIELDS` omits it and `analyze_mapping_prediction` opens with `require_exact_fields(...)` (line 157), which rejects *unsupported* fields, not just missing ones:

```
BenchmarkProtocolError: mapping prediction has unsupported fields ['error'] or missing fields []
```

Note the asymmetry inside this PR: `forward_benchmark.PAYLOAD_FIELDS`, `condition_benchmark.PAYLOAD_FIELDS` and `single_step_benchmark.PREDICTION_PAYLOAD_FIELDS` **all** carry `"error"`. Atom mapping is the sole omission — and the worker's `error` field is the only place a per-reaction mapper failure is recorded ("mapper returned an empty record"), so the only way to satisfy the normalizer is to strip the failure diagnostics the protocol says to preserve.

Related, at line 158: `require_text(prediction["mapped_reaction"], ...)` raises for the exact record the worker produces on failure (`mapped_reaction: ""`), and there is no `status` bucket like `forward_benchmark`'s `"empty"`/`"invalid"`. Dropping the failed reaction instead hits `"mapping output must cover every reaction"` at line 251 — both exits are closed, so one unmappable reaction in a 500-reaction set makes the scenario unscoreable. It also means `valid_mapping_rate` can never drop below 1.0 for a *backend* failure.

Suggested: add `"error"` to `PREDICTION_FIELDS` and give a failed mapping a recorded status the way `forward_benchmark.py:145-164` does.

skills/retrosynthesis_planning/reaction_model_worker.py
```diff
@@ -0,0 +1,754 @@
...
+    if (
+        isinstance(max_routes, bool)
+        or not isinstance(max_routes, int)
+        or not 1 <= max_routes <= 100
+    ):
+        raise RequestError("invalid_request", "max_routes must be between 1 and 100")
+
+    with contextlib.redirect_stdout(sys.stderr):
+        finder = AiZynthFinder(configfile=config_path)
+        finder.stock.select(stocks or finder.stock.items)
```
Nobody-Zhang: **Passing no `stocks` silently enables every stock the config defines.**

```python
finder.stock.select(stocks or finder.stock.items)                    # ALL stocks
finder.expansion_policy.select(policies or finder.expansion_policy.items[0])   # FIRST policy only
```

The two defaults pull in opposite directions on adjacent lines, and `ReactionModelBackend.plan_routes` defaults `stocks: Iterable[str] = ()` (`reaction_model_backends.py:252`) — so the permissive branch is the default path.

A caller invoking `plan_routes(targets, config_path=cfg)` intending the reviewed stock gets every stock in the config, including any large exploratory or vendor-catalogue stock. More leaves resolve, `statistics["is_solved"]` is true more often, and the worker reports `termination_reason: "solved"` (line 295). Stock closure is precisely the quantity `SKILL.md:56` warns must not be mistaken for feasibility, and this default maximizes it. Meanwhile only the first expansion policy is active, so the route set is *narrower* than the operator assumes.

The response also never echoes which stocks and policies were actually selected into `search_stats`, so the choice is unrecoverable after the fact.

Suggested: require `stocks` explicitly (or default to a single named stock), and echo the resolved stock/policy names into `search_stats`.

skills/retrosynthesis_planning/multistep_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            reaction_count += 1
+            if not children:
+                issues.append("reaction_without_precursors")
+                solved = False
+            else:
+                if any(_kind(child) != "molecule" for child in children):
+                    issues.append("reaction_has_nonmolecule_child")
+                solved = all(visit(child, depth + 1) for child in children)
```
Nobody-Zhang: **`any()`/`all()` short-circuit over a `visit()` that has side effects, so verification results depend on child order.**

`visit()` mutates `issues`, `reaction_count`, `max_depth` and `unresolved`, but its return value drives short-circuiting builtins — so once one child settles the boolean, the remaining siblings are never visited and their defects are never recorded.

Same route, same chemistry, only `children` order reversed:

```
A-first : {'verified_solved': True, 'valid': True,  'issues': [],                          'reaction_count': 1}
B-first : {'verified_solved': True, 'valid': False, 'issues': ['invalid_molecule_smiles'], 'reaction_count': 2}
```

A malformed subtree parked behind a solved sibling passes verification outright. Symmetrically on the `all()` line, a reaction whose precursors `X`, `Y`, `Z` are *all* missing from stock reports only the first:

```
{'verified_solved': False, 'unresolved_leaves': ['X'], 'reaction_count': 1, 'max_depth': 1}
```

A chemist reading the frozen artifact to decide what still needs sourcing sees one missing building block where there are three — and `reaction_count`/`max_depth`, which are reported route-length metrics, under-report whatever was short-circuited away.

Suggested: materialize first — `results = [visit(child, depth + 1) for child in children]` then `all(results)` / `any(results)`.

skills/retrosynthesis_planning/benchmark_common.py
```diff
@@ -0,0 +1,120 @@
...
+        raise BenchmarkProtocolError(f"{field} must be JSON serializable") from exc
+
+
+def require_exact_fields(
+    value: Mapping[str, Any],
+    expected: set[str] | frozenset[str],
+    *,
+    field: str,
+) -> None:
+    actual = set(value)
```
Nobody-Zhang, line 60: **`require_exact_fields` accepts a list as if it were a row, so five validators raise `TypeError` instead of `BenchmarkProtocolError`.**

The parameter is annotated `Mapping[str, Any]` but the body only calls `set(value)`, which succeeds for any iterable — a *list* of the expected field names passes the check. The validators then index the row directly:

```
require_exact_fields ACCEPTED a list: ['reaction_id', 'reaction_smiles']

atom_mapping.validate_public_reactions -> TypeError: list indices must be integers or slices, not str
multistep.validate_targets             -> TypeError: list indices must be integers or slices, not str
yield.validate_yield_inputs            -> TypeError: list indices must be integers or slices, not str
single_step.validate_public_targets    -> SingleStepProtocolError: target row 1 must be an object   <- correct
```

`single_step_benchmark.validate_public_targets:147-148` — the function these were factored out of — kept the guard (`if not isinstance(raw_row, Mapping): raise ...`); the extraction into `benchmark_common` dropped it. A JSON file whose rows are arrays rather than objects therefore crashes with a raw traceback out of a module whose entire contract is that strict admission raises `BenchmarkProtocolError`, so callers catching the protocol error see a crash instead of a rejection.

Suggested:
```suggestion
    if not isinstance(value, Mapping):
        raise BenchmarkProtocolError(f"{field} must be an object")
    actual = set(value)
```

skills/retrosynthesis_planning/multistep_benchmark.py
```diff
@@ -0,0 +1,370 @@
...
+            if route["duplicate_of_rank"] is not None:
+                duplicate_count += 1
+            if route["verification"]["verified_solved"]:
+                solved_routes += 1
+            if route["route_signature"] in ref_signatures and first_exact is None:
+                first_exact = int(route["rank"])
+            best_similarity = max(
+                best_similarity,
+                *(
+                    route_similarity(route["raw_route"], reference)
```
Nobody-Zhang, line 363: **The `evaluate` half of Scenario 2 silently degrades to raw-string comparison when RDKit is absent; the `normalize` half fails closed.**

`normalize_planner_outputs` reaches `rdkit_canonicalize`, which raises `ChemistryDependencyError` without RDKit. But `evaluate_routes` computes the two headline metrics through `route_signature`/`route_similarity`, which route to `route_review._safe_canonicalize` → `kernel.canonicalize_smiles`, documented as "canonical SMILES when RDKit is installed, **else a stripped string**".

RDKit is not in this repo's default venv, so `scenario_benchmark_cli evaluate multistep` runs this way today. A predicted route rooted at `CCO` and a reference route rooted at the identical molecule spelled `OCC`:

```
reference_route_recovery_rate:   0.0    (should be 1.0)
mean_best_reference_similarity:  0.333  (should be 1.0)
```

No exception, no warning, no marker in the result — which is then sealed with `result_sha256` and published as the private benchmark score. A scoring path that quietly answers an easier question is worse than one that refuses.

Suggested: have `evaluate_routes` fail closed the same way `normalize` does (or take an explicit `canonicalizer` and record which one was used in the result envelope).

skills/retrosynthesis_planning/scenario_benchmark_cli.py
```diff
@@ -0,0 +1,180 @@
...
+    parser.add_argument("action", choices=("normalize", "evaluate"))
+    parser.add_argument("scenario", choices=tuple(SCENARIO_IDS))
+    parser.add_argument("--predictions", required=True)
+    parser.add_argument("--output", required=True)
+    parser.add_argument("--inputs")
+    parser.add_argument("--references")
+    parser.add_argument("--stock")
+    parser.add_argument("--config")
+    parser.add_argument("--vocabulary")
+    parser.add_argument("--top-k", type=int, default=10)
```
Nobody-Zhang: **`evaluate` ignores the Top-K budget frozen into the artifact and re-reads `--top-k`, defaulting to 10.**

`_normalize` writes the budget into the artifact (`metadata = {"top_k": args.top_k}`, lines 100/115), and `_evaluate` goes to real trouble to prove the artifact is the frozen one — it pops `trajectory_sha256`, recomputes it, and matches `scenario_id` (lines 131-136). It then passes `args.top_k` straight through and never reads `artifact["metadata"]["top_k"]`.

Normalize a conditions run at `--top-k 1`, then evaluate without `--top-k`:

```
frozen artifact metadata: {'top_k': 1}
cutoffs reported        : ['1', '3', '5', '10']     <- beams that were never allowed to exist
unused_budget_slot_rate : 0.9                       <- 100% of the budget was actually used
```

The hash check still passes, so the reader has false confidence that the numbers are tied to the frozen trajectory. `metadata` is written by `build_intermediate_artifact` and read back nowhere in the repo.

Suggested: read `top_k` from `artifact["metadata"]` in `_evaluate` and, if `--top-k` is also given, refuse on mismatch rather than silently preferring the flag.

skills/retrosynthesis_planning/reaction_model_worker.py
```diff
@@ -0,0 +1,754 @@
...
+                "message": f"{type(exc).__name__}: {_public_error(exc)}",
+                "retryable": False,
+            },
+        )
+    encoded = (json.dumps(response, sort_keys=True) + "\n").encode("utf-8")
+    if reserved is None:
+        sys.stdout.buffer.write(encoded)
+        sys.stdout.buffer.flush()
+    else:
+        os.write(reserved, encoded)
```
Nobody-Zhang: **The response write discards `os.write`'s short count, so a large response can be silently truncated.**

`os.write` is a single `write(2)`; its return value is the number of bytes actually transferred, and it is dropped here. Measured against a realistic `plan_routes` payload:

```
payload bytes: 2080013  ->  os.write returned: 65536   (short write: True)
```

The host allows responses up to `MAX_RESPONSE_BYTES = 32 * 1024 * 1024` and the pipe buffer is 64 KiB, so this is well inside the supported envelope. A `write(2)` interrupted after transferring some bytes returns the short count rather than `-1/EINTR`, so PEP 475's automatic retry does not cover it: the worker exits 0 having emitted truncated JSON, and the host raises `BackendExecutionError("invalid_json", "reaction backend did not return one JSON object")` — attributing a transport bug to the model.

The `reserved is None` fallback two lines above (`sys.stdout.buffer.write` + `flush`) loops correctly, so the *primary* path is weaker than its own fallback, and weaker than the sibling it was derived from:

```python
# syntheseus_worker.py:504-506
with os.fdopen(reserved_fd, "wb", closefd=True) as handle:
    handle.write(payload)
    handle.flush()
```

`os.close(reserved)` is also outside any `try`, so a `BrokenPipeError` skips it. Suggested: adopt the sibling's `os.fdopen(..., "wb", closefd=True)` form.

skills/retrosynthesis_planning/reaction_model_backends.py
```diff
@@ -0,0 +1,312 @@
...
+    ModelManifest,
+    load_model_manifest,
+)
+
+WIRE_SCHEMA_VERSION = 1
+SUPPORTED_MODELS = {
+    "aizynthfinder": "AiZynthFinder",
+    "rxnmapper": "RXNMapper",
+    "reactiont5_forward": "ReactionT5v2",
+    "reactiont5_yield": "ReactionT5v2",
```
Nobody-Zhang, line 28: **The quarantine and block decisions exist only in prose — no code path enforces them — and the two ReactionT5v2 checkpoints share one identity.**

`SUPPORTED_MODELS` maps `reactiont5_forward` and `reactiont5_yield` to the **same** string `"ReactionT5v2"`, and the only identity gate in `__init__` is `loaded.model != SUPPORTED_MODELS[model]` (line 156). So a manifest built for the quarantined `sagawa/ReactionT5v2-yield` checkpoint is accepted by *both*:

```
ReactionModelBackend('reactiont5_yield',   manifest=<QUARANTINED yield manifest>) -> accepted
ReactionModelBackend('reactiont5_forward', manifest=<QUARANTINED yield manifest>) -> accepted
```

`predict_yields` then runs the quarantined regression head and returns `warnings: []` — indistinguishable downstream from an admitted model, and the numbers flow into `yield_benchmark.evaluate_yield_predictions` and out as MAE/RMSE/R²/top-decile-enrichment sealed with `result_sha256`.

`grep -rni "quarantin|blocked" skills/**/*.py` returns **zero** hits, while the state is asserted in seven prose locations (`MODEL_BACKENDS.md:41,53`, `MODEL_BACKENDS_zh.md:28,40`, `MODEL_TASKS.md:34`, `SKILL.md:72`, `reaction-yield-estimation/SKILL.md:27`). Two sources of truth, only one of which says no.

The machine-readable fields already exist and are dead: `EnvironmentSpec.requires_terms_review` and `ModelManifest.provenance_status` are only ever *serialized*, never read by a gate; and the pinned `UPSTREAM_DISTRIBUTIONS` / `PARROT_HF_ARTIFACTS` byte sizes and SHA256s are never compared against a downloaded file. `external_backends.SyntheseusBackend` shows the intended shape — it carries `allow_model_download: bool = False` and refuses.

Suggested: give the registry an explicit `state: admitted | quarantined | blocked`, refuse at construction unless the caller passes an explicit override, and distinguish the two ReactionT5v2 checkpoints by `checkpoint_id`, not just `model`.

skills/retrosynthesis_planning/condition_benchmark.py
```diff
@@ -0,0 +1,283 @@
...
+        ranks: list[int] = []
+        for candidate in prediction["predictions"]:
+            submitted += 1
+            oov += bool(candidate["oov_slots"])
+            duplicates += candidate["duplicate_of_rank"] is not None
+            values = tuple(candidate["condition_signature"])
+            if candidate["valid"] and values in refs:
+                ranks.append(candidate["rank"])
+        first = min(ranks) if ranks else None
+        top_one = prediction["predictions"][0] if prediction["predictions"] else None
```
Nobody-Zhang: **`top1_slot_recall` scores whatever beam is first in the list, not the beam whose `rank == 1`.**

`normalize_condition_outputs` validates ranks only as unique and `1 <= rank <= budget` — never as 1-based or contiguous. So a backend that omits its rank-1 beam and submits `[{"rank":2,...},{"rank":3,...}]` is accepted, and its rank-2 candidate is then credited into a metric published as *Top-1* slot recall, with no warning row.

This is reachable through the worker path in this PR: `_parse_parrot_rank` (`reaction_model_worker.py:98`) derives each rank from Parrot's `top-k` CSV column and falls back to the row ordinal, so a missing or unparseable first row shifts the whole set.

The adjacent exact-tuple path at lines 229-231 gets it right — it keys off `candidate["rank"]`.

Separately, the slot comparison at 234-237 counts `None == None` as a hit, so a model that predicts nothing but `catalyst1` posts a high `top1_slot_recall` against references whose `solvent2`/`reagent2` are mostly empty.

Suggested: select `next((c for c in prediction["predictions"] if c["rank"] == 1), None)`, and require a non-`None` reference slot for a slot hit.

skills/retrosynthesis_planning/yield_benchmark.py
```diff
@@ -0,0 +1,306 @@
...
+        * sum((b - right_mean) ** 2 for b in right)
+    )
+    return numerator / denominator if denominator else None
+
+
+def _group_metrics(rows: Sequence[Mapping[str, Any]]) -> dict[str, Any]:
+    errors = [float(row["predicted"]) - float(row["observed"]) for row in rows]
+    observed = [float(row["observed"]) for row in rows]
+    predicted = [float(row["predicted"]) for row in rows]
+    mean_observed = sum(observed) / len(observed)
```
Nobody-Zhang, line 190: **Every `validate_*_inputs` rejects an empty set cleanly; not one `evaluate_*` does, and all six then divide by zero.**

The non-empty invariant is clearly intended — it is enforced on the `normalize` side in all six modules:

```
yield.validate_yield_inputs([])            -> BenchmarkProtocolError: yield inputs must not be empty
condition.validate_condition_inputs([])    -> BenchmarkProtocolError: condition inputs must not be empty
forward.validate_forward_inputs([])        -> BenchmarkProtocolError: forward inputs must not be empty
atom.validate_public_reactions([])         -> BenchmarkProtocolError: reaction set must not be empty
multistep.validate_targets([])             -> BenchmarkProtocolError: planning targets must not be empty
```

and missing on the `evaluate` side in all six. `scenario_benchmark_cli._evaluate` admits any artifact whose `records` is a `list` (line 125-127) and any references array, and an *empty* artifact is self-consistent, so its `trajectory_sha256` validates. Reproduced through the real CLI:

```
$ python -m retrosynthesis_planning.scenario_benchmark_cli evaluate yield \
    --predictions empty_art.json --references empty_refs.json --output out.json
  File ".../yield_benchmark.py", line 190, in _group_metrics
    mean_observed = sum(observed) / len(observed)
ZeroDivisionError: division by zero
```

Same shape at `multistep_benchmark.py:337-353`, `condition_benchmark.py:242-268`, `forward_benchmark.py:250-289`, `atom_mapping_benchmark.py:347-348`, `single_step_benchmark.py:508`. `yield_benchmark.py:286` additionally calls `max()` over a possibly-empty sequence (`ValueError`). The tell is that `duplicate_route_rate` and `invalid_prediction_rate` *are* guarded (`if submitted else 0.0`) while the by-`count` ones are not.

Suggested: reject empty `records`/`references` at the top of each `evaluate_*` with the same `BenchmarkProtocolError` the validators raise.

skills/retrosynthesis_planning/route_review.py
```diff
@@ -177,13 +177,25 @@ def _route_feature_set(route: Mapping[str, Any]) -> set[str]:
...
     if not left and not right:
         return 1.0
     union = left | right
     return len(left & right) / len(union) if union else 0.0
 
 
+def route_similarity(left: Mapping[str, Any], right: Mapping[str, Any]) -> float:
+    """Compare routes by reaction, product, precursor, and terminal features."""
+
+    return _jaccard(set(route_feature_set(left)), set(route_feature_set(right)))
```
Nobody-Zhang: **A route with no usable `tree` collapses to one constant signature and scores a perfect reference match.**

(Anchored on the new `route_similarity`; the mechanism starts at `_route_tree`, line 110.)

`_route_tree` returns `{}` for anything that is not a mapping, `_jaccard` returns `1.0` for two empty sets (line 187-188), and `route_signature` is then a fixed constant:

```python
route_similarity({"rank": 1}, {"tree": {}})          # -> 1.0
route_signature({"rank": 1}) == route_signature({"tree": {}})
# -> True, both "route:b09ffc9a7ce247cda121"
```

`multistep_benchmark.evaluate_routes` validates reference routes only as `isinstance(route, Mapping)` (lines 291-297), so references curated in the natural bare-tree PaRoutes shape — no `{"tree": ...}` wrapper — all collapse to that one digest. A submission that `normalize_planner_outputs` already flagged `issues: ["missing_route_tree"]` then reports:

```
reference_recovered: True
best_reference_similarity: 1.0
reference_route_recovery_rate: 1.0
```

sealed with `result_sha256`. The constant signature also makes every such route a `duplicate_of_rank` of the first.

`1.0` for two empty sets is the right convention for de-duplication (`select_diverse_routes`) and an inverted one for scoring. Suggested: have `route_similarity` return `0.0` when either side has no recoverable feature set, and have `evaluate_routes` reject a reference route that carries no tree.

PR #121, opened 2026-08-25
Feat/single cell spatial omics skills

Devin-jun:
## Summary

Add an OpenAI4S-maintained `single-cell-rna-analysis` Skill for reproducible,
CPU-first analysis of human or mouse, cell-called 10x GEX scRNA-seq and
snRNA-seq count matrices.

The Skill provides:

- A versioned input and configuration contract with explicit `preflight(config)`
  validation.
- Two analysis modes:
  - `descriptive` for a single H5AD without a valid donor/condition contrast.
  - `comparative` for donor-aware, tested-versus-reference analyses.
- Sample-aware QC, optional Scrublet doublet detection, count preservation,
  normalization, PCA, UMAP, Leiden resolution sweeps, descriptive cluster
  markers, and evidence-assisted annotation.
- Explicitly requested Harmony integration with batch/confounding validation.
- Donor-aware pseudobulk differential expression through PyDESeq2 and
  differential abundance through Milo/Pertpy when replication requirements are
  satisfied.
- Conservative scientific gates that prevent normalized-only input, invented
  metadata, cell-level pseudoreplication, unsupported contrasts, and
  unconfirmed annotations from being treated as ground truth.
- Restartable checkpoints through `resume(run_dir)`, input/configuration
  fingerprinting, SHA-256 manifests, structured status reporting, and
  Artifact-ready output bundles.
- Separate reference contracts for input validation, scientific processing,
  annotation, statistics, and output delivery.
- English and Chinese Skill documentation and index updates.

The change also adds a Python 3.11 `singlecell` optional dependency group with
pinned Scanpy, Pertpy, PyDESeq2, Harmony, and related CPU dependencies; updates
the portable science environment and lockfile; and adds a dedicated offline CI
job and focused contract tests.

Spatial transcriptomics and other spatial-omics workflows are intentionally out
of scope for this PR and should be added separately.

## Area

- [ ] Web UI / frontend
- [ ] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

Ran:

```text
uv sync --locked --extra singlecell

uv run python -c "import harmonypy, pertpy, pydeseq2, scanpy, skmisc"

uv run python -m pytest \
  tests/test_single_cell_rna_analysis_skill.py \
  -m "not network"

uv run python -m pytest \
  tests/test_setup_environments.py \
  tests/test_kernel_recovery.py::test_bundled_sidecar_recovery_compatibility_is_explicit

Manual PBMC3K descriptive-mode smoke test:
- Loaded a public PBMC3K H5AD with raw integer counts in X.
- Used descriptive mode without donor, condition, or contrast metadata.
- preflight status: valid
- run status: completed
- QC, embedding, clustering, annotation, report, manifest, tables, figures,
  and final analysis H5AD were generated.
- DE status: not_applicable_descriptive
- DA status: not_applicable_descriptive
- No donor/condition metadata or inferential result was synthesized.
```

Not run:

```text
uv run python -m pytest \
  tests/test_single_cell_rna_analysis_skill.py \
  -m "network"

Full repository test suite.
GPU, SSH, Docker, remote-compute, and live-LLM tests.
Spatial transcriptomics workflows.
```

Reason:

```text
The optional network test downloads an external single-cell dataset and is
excluded from the deterministic offline CI job. The remaining repository-wide
and infrastructure-dependent suites are outside this Skill-focused change.
This workflow is CPU-first and does not require GPU, SSH, Docker, secrets, or a
live LLM. Spatial analysis is outside the scope of this PR.
```

## Risk and Reviewer Focus

The primary review target is
`skills/single-cell-rna-analysis/kernel.py`, especially:

- `preflight(config)` validation and the separation between descriptive and
  comparative modes.
- Raw-count validation and preservation in `layers["counts"]`.
- The rule that descriptive mode must not create donor, condition, batch,
  treatment, replicate, DE, or DA outputs.
- Donor replication, pairing, contrast direction, and design-rank checks for
  comparative analyses.
- Harmony opt-in behavior and rejection of missing or condition-confounded batch
  keys.
- Annotation evidence handling, `Unknown` fallback behavior, and the boundary
  between candidate and user-confirmed labels.
- Checkpoint invalidation, resume behavior, manifest checksums, and cleanup of
  stale comparative outputs when an output directory is reused for a
  descriptive run.

Dependency review should focus on `pyproject.toml`, `envs/python.yml`,
`scripts/bundled_packages.txt`, and `uv.lock`. The single-cell stack requires
Python 3.11 and is kept in a separate optional dependency group so that the
OpenAI4S core does not gain hard scientific dependencies. Scientific imports
inside the Skill remain lazy and produce actionable failures when the optional
stack is unavailable.

The new CI job installs the locked CPU stack and runs deterministic offline
contracts. Reviewers should confirm that the dependency pins remain mutually
compatible and that the additional CI runtime is acceptable.

No core gateway, host-RPC, store, Web UI vendor, or fixture behavior is changed.

### Diff Readiness

      `host_dispatch.py`, `store.py`, `app.js`, `worker.py`, `manager.py`.
      reformatted.

### Tests

      a documented smoke test.

### Core Dependency Policy

      stdlib web server.
      in-tree use site.

      verification notes.
      guarantees.

### Docs

      appropriate.
      affected.


Nobody-Zhang:
@cursoragent Review this PR.

envs/python.yml
```diff
@@ -50,5 +49,12 @@ dependencies:
   - pip
   - pip:
       - fair-esm==2.0.0
+      # Workflow-pinned CPU single-cell analysis stack. Scientific imports are
+      # lazy in the Skill. Pertpy 1.0.3 is the latest release compatible with
+      # this Python 3.11 environment and the pinned Scanpy 1.11.5.
+      - scanpy[harmony,skmisc]==1.11.5
```
Nobody-Zhang: **[correctness / CONFIRMED — blocks merge]** These pip entries with extras (`scanpy[harmony,skmisc]==1.11.5`) are unparseable by `openai4s/kernel/readiness.py`'s `_DEPENDENCY_NAME` regex, so `load_standard_profile_requirements()` raises `_ManifestError`. Reproduced at this head: `tests/test_environment_readiness.py` / `tests/test_env_generations.py` fail, and at runtime env-generation verification and standard-profile readiness fail closed (`manifest_unavailable`; with trusted delivery the gateway 503s every science turn). Fix: teach the parser optional `[extras]` after the name and update the pinned authoritative-list/digest tests.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+    """Validate provenance and return or rebuild an interrupted/stale run."""
+    output = Path(run_dir).expanduser().resolve()
+    config_path = output / "config.resolved.json"
+    if not config_path.exists():
+        raise ValueError(f"missing resolved configuration: {config_path}")
+    config = _read_json(config_path)
+    check = preflight(config)
+    manifest_path = output / "run_manifest.json"
+    prior_manifest: dict[str, Any] | None = None
+    if manifest_path.exists() and check["status"] == "valid":
```
Nobody-Zhang: **[correctness / CONFIRMED — data loss]** `prior_manifest` is only loaded when the fresh preflight is valid. A *transient* preflight failure (input on an unmounted share, file renamed) leaves `earliest="preflight"`, and the invalidation block then moves every deliverable of a completed run — `analysis.h5ad`, tables, figures, `report.md`, `run_manifest.json` — into `.invalidated/` and overwrites the manifest with a failed one. "Cannot validate right now" should fail without touching the run, not be treated as "everything is stale".

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+        manifest["status"] = "failed"
+        manifest["stages"]["preflight"] = "failed"
+        manifest["errors"] = check["errors"]
+        _write_report(output, manifest)
+        _finalize_manifest(output, manifest)
+        return _result(output, manifest)
+    manifest["stages"]["preflight"] = "completed"
+
+    try:
+        if resolved["analysis_mode"] == "descriptive":
```
Nobody-Zhang: **[correctness / CONFIRMED]** Stale-output cleanup only happens in descriptive mode. A comparative rerun into a reused dir whose DE/DA is skipped (`skipped_insufficient_replicates`) leaves the previous run's `tables/differential_expression.csv` on disk, `_plot_outputs` re-plots it into a fresh volcano PDF, and `_finalize_manifest` checksums both into a manifest whose `statistics_status.de` says skipped. Stale `annotation_evidence.csv` leaks the same way (it is only written when non-empty). Suggest removing `INFERENTIAL_OUTPUTS` (and the evidence table) unconditionally before the statistics/annotation stages regenerate them.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+        {
+            "upstream": hashes["embedding"],
+            "resolutions": (config.get("clustering") or {}).get("resolutions"),
+            "selected_resolution": (config.get("clustering") or {}).get(
+                "selected_resolution"
+            ),
+            "seed": config.get("seed"),
+        }
+    )
+    hashes["annotation"] = digest(
```
Nobody-Zhang, line 778: **[correctness / CONFIRMED]** The annotation stage hash digests only the annotation config (paths + margin), and `_input_fingerprint` covers only `input.path` + matrices — the *contents* of `marker_panel`, `reference_h5ad`, `confirmed_mapping` are never fingerprinted, and the resume fast path compares only `config_sha256`/`input_fingerprint`. Editing a mapping CSV in place and calling `resume()` silently returns the stale prior result with the old labels. references/annotation-contract.md promises a recorded mapping hash that is implemented nowhere. Suggest hashing these files into the annotation stage hash and making the fast path require matching stage hashes.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+
+def _read_json(path: Path) -> dict[str, Any]:
+    with path.open("r", encoding="utf-8") as handle:
+        value = json.load(handle)
+    if not isinstance(value, dict):
+        raise ValueError(f"JSON object required: {path}")
+    return value
+
+
+def _write_json(path: Path, value: Any) -> None:
```
Nobody-Zhang, line 79: **[correctness / CONFIRMED]** `run_manifest.json` is written once at the very end of `_run`, non-atomically (`_write_json` truncates in place), and `resume()` reads it with no error handling. Consequences: (a) a run interrupted before completion has no manifest, so `resume()` invalidates *all* completed checkpoints and recomputes from scratch — the advertised resume-after-interruption case; (b) a SIGKILL during `_finalize_manifest` leaves a truncated JSON that makes every later `resume()` raise `JSONDecodeError`. Suggest temp-file + `os.replace` (the repo's established pattern, e.g. `env_generations._write_json`), persisting the manifest after each stage, and tolerating an unreadable prior manifest.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+        {
+            "input_fingerprint": input_fingerprint,
+            "analysis_mode": config.get("analysis_mode"),
+            "input": config.get("input"),
+            "organism": config.get("organism"),
+            "modality": config.get("modality"),
+            "reference": config.get("reference"),
+        }
+    )
+    hashes["qc"] = digest(
```
Nobody-Zhang, line 741: **[correctness / CONFIRMED]** The qc/embedding stage hashes omit the design/integration metadata keys `_load_data` bakes into the checkpoint's `obs`. Adding `design.covariates=["sex"]` (or a Harmony batch key) changes only a downstream hash, so resume reuses `04_annotation.h5ad` whose `obs` never received the new column → `KeyError` in `_pseudobulk` — and every subsequent `resume()` fails identically, a permanent loop only a manual full rerun escapes. The obs-affecting keys need to be part of the qc-stage hash.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+    covariates = [str(value) for value in design.get("covariates", [])]
+    formula_terms = [donor] if design.get("paired") else covariates
+    formula = "~ " + " + ".join([*formula_terms, condition])
+    result_frames = []
+    gene_columns = [
+        column
+        for column in pseudobulk.columns
+        if column not in {"sample_id", "analysis_group"}
+    ]
+    for group in sorted(metadata["analysis_group"].unique()):
```
Nobody-Zhang, line 1581: **[science-validity / CONFIRMED]** The per-group DESeq2 loop has (a) no per-group eligibility gate — the ≥3-donors-per-level rule from the statistics contract is enforced only globally, so a cell type present in 1 control + 1 stim sample still gets published padj values; and (b) no per-group error isolation — a condition-exclusive group makes `DeseqDataSet` raise on a single-level factor and the exception aborts DE for *every* group (results are only flushed after the loop). Suggest a per-group both-levels + donor-count check (skip + record) and a per-group try/except.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+            adata.obs[sample_key] = [
+                str(value).strip() for value in adata.obs[sample_key]
+            ]
+        ok, reason = _is_raw_counts(adata.X, np)
+        if not ok:
+            raise ValueError(reason)
+        if not adata.obs_names.is_unique:
+            raise ValueError("h5ad cell IDs are not unique")
+        if not descriptive:
+            records = adata.obs[list(metadata_keys)].astype(str).to_dict("records")
```
Nobody-Zhang: **[correctness / CONFIRMED]** `astype(str)` turns NaN obs metadata into the literal level `"nan"`, which passes every nonempty gate (the `isna` check exists only in descriptive mode): missing donors collapse into one phantom donor that can unlock the ≥3-donor replication gate and enters the DESeq2 design as a real level. The CSV path has the same class of hole: `DictReader` fills missing trailing fields with `None`, and `str(None)` → `"None"` passes validation, so ragged sheet rows silently create a `"None"` condition/covariate level.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+    evidence = pd.DataFrame(evidence_rows)
+    return adata, status, warnings, evidence
+
+
+def _pseudobulk(
+    adata: Any, config: Mapping[str, Any], pd: Any, np: Any
+) -> tuple[Any, Any, str]:
+    design = config["design"]
+    sample_key = design.get("sample_key", "sample_id")
+    group_key = (
```
Nobody-Zhang, line 1482: **[correctness / CONFIRMED]** Pipeline-reserved obs columns already present in the *input* h5ad are adopted as this run's outputs: a pre-existing `confirmed_cell_type` column survives untouched and becomes the pseudobulk DE grouping here — without any confirmed mapping, while `annotation_status` says otherwise — and foreign `leiden_*` columns are exported in `cluster_assignments.csv` and plotted in the resolution sweep as if this run produced them. Suggest dropping/renaming reserved columns (`leiden_*`, `cluster`, `candidate_cell_type`, `confirmed_cell_type`, qc flags) at load with a warning.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+    if counts_layer == "X":
+        return adata.X
+    if counts_layer not in adata.layers:
+        raise ValueError(f"declared counts layer {counts_layer!r} is missing")
+    return adata.layers[counts_layer]
+
+
+def _read_one(path: str, matrix_format: str, counts_layer: str, sc: Any) -> Any:
+    if matrix_format == "10x_mtx":
+        adata = sc.read_10x_mtx(path, var_names="gene_symbols", make_unique=False)
```
Nobody-Zhang: **[correctness / CONFIRMED]** `sc.read_10x_mtx(var_names="gene_symbols", make_unique=False)` + the hard duplicate-gene rejection below means every standard CellRanger matrix fails to load: stock human/mouse 10x references contain duplicated gene symbols (e.g. TBCE in GRCh38-2020-A), so any real `filtered_feature_bc_matrix/` raises `ValueError("sample ... has duplicate gene IDs")`. The `10x_mtx` format is effectively unusable on genuine data — reading `gene_ids` (with a symbol column for QC prefixes) or `make_unique=True` is required. Left unfixed here since the resolution is a design decision.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+        errors.append("integration.method must be none or harmony")
+    batch_keys = integration.get("batch_keys", [])
+    if not isinstance(batch_keys, list):
+        errors.append("integration.batch_keys must be a list")
+    elif integration.get("method") == "harmony" and not batch_keys:
+        errors.append("Harmony requires explicit integration.batch_keys")
+    if analysis_mode == "descriptive" and integration.get("method") != "none":
+        errors.append("descriptive analysis does not support integration")
+
+    statistics = config.get("statistics") or {}
```
Nobody-Zhang, line 304: **[correctness / CONFIRMED]** `_config_errors` never type-checks boolean fields and consumers use bare truthiness, so JSON-string values silently invert behavior: `"paired": "false"` runs the *paired* model, `"statistics": {"de": "false"}` runs DESeq2 anyway in comparative mode — while descriptive mode's `any()`-based validator coincidentally rejects the same string, so the two modes disagree about the same value. Suggest validating `design.paired`, `statistics.de/da`, and `qc.doublet_detection` as real booleans.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+        else np.asarray(matrix)
+    )
+    values = np.asarray(values)
+    if values.size == 0:
+        return False, "count matrix is empty"
+    if not np.all(np.isfinite(values)):
+        return False, "counts contain non-finite values"
+    if np.any(values < 0):
+        return False, "counts contain negative values"
+    if not np.allclose(values, np.rint(values), atol=1e-8):
```
Nobody-Zhang: **[correctness / CONFIRMED]** `np.allclose(values, np.rint(values), atol=1e-8)` keeps numpy's default `rtol=1e-5`, making the integer-count gate magnitude-dependent: once a value ≥ ~5e4 the tolerance `1e-5·|v|` exceeds the max rounding distance, so e.g. 50000.4 validates as a raw count and normalized/scaled data can flow into `layers["counts"]` and PyDESeq2. The explicit `atol=1e-8` shows the tolerance was meant to be tiny — add `rtol=0`.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+    try:
+        from pydeseq2.dds import DeseqDataSet
+        from pydeseq2.ds import DeseqStats
+    except ImportError:
+        return "failed_missing_dependency"
+    design = config["design"]
+    condition = str(design["condition_key"])
+    donor = str(design["donor_key"])
+    covariates = [str(value) for value in design.get("covariates", [])]
+    formula_terms = [donor] if design.get("paired") else covariates
```
Nobody-Zhang, line 1571: **[science-validity / CONFIRMED]** `formula_terms = [donor] if design.get("paired") else covariates` silently drops all user-declared covariates from the DESeq2 and Milo formulas (the preflight rank check mirrors this), with no error or warning. A paired design with `covariates=["processing_batch"]` fits `~ donor + condition`; the declared covariate's effect loads onto the condition coefficient while `config.resolved.json` records it as part of the design. Given the skill's reject-don't-silently-ignore philosophy, preflight should reject (or at least warn on) `paired=True` with nonempty covariates.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+                    manifest["warnings"].append(
+                        "Milo DA failed; see the adjacent error file."
+                    )
+            else:
+                de_status = "skipped_insufficient_replicates"
+                da_status = "skipped_insufficient_replicates"
+                manifest["warnings"].append(
+                    "DE/DA skipped: each contrast level requires at least three independent donors."
+                )
+        manifest["statistics_status"] = {"de": de_status, "da": da_status}
```
Nobody-Zhang, line 2197: **[correctness / CONFIRMED]** DE/DA failures are downgraded: `stages.statistics` is marked completed and the overall run reports `completed`/`completed_with_warnings` even when a status is `failed`, and `failed_missing_dependency` appends no warning at all — a run without pydeseq2 reports plain `completed`. The resume fast path keys off status alone, so it also vouches for such runs. At minimum, any `failed*` status should append a warning so the run reports `completed_with_warnings`.

skills/single-cell-rna-analysis/kernel.py
```diff
@@ -0,0 +1,2012 @@
...
+                raise ValueError("sample identifiers must be nonempty")
+            if len(sample_values - {""}) != 1:
+                raise ValueError("descriptive analysis requires exactly one sample")
+            adata.obs[sample_key] = [
+                str(value).strip() for value in adata.obs[sample_key]
+            ]
+        ok, reason = _is_raw_counts(adata.X, np)
+        if not ok:
+            raise ValueError(reason)
+        if not adata.obs_names.is_unique:
```
Nobody-Zhang, line 1012: **[correctness / CONFIRMED]** The single-h5ad path checks `obs_names.is_unique` but never `var_names.is_unique` (the sample-sheet path does, per sample). Downstream, `_pseudobulk`'s per-gene dict and `_annotate`'s name→index lookup silently keep only the *last* duplicate gene's values — wrong pseudobulk totals in the table DESeq2 consumes, and marker scores against the wrong column, with no error. Add the same duplicate-gene rejection here.

PR #128, opened 2026-08-26
feat: add Volcengine SSO onboarding

difficulttopickaname:
## Summary

Add Volcengine Ark SSO onboarding for OpenAI4S. Users can authorize through the official Ark CLI and continue with their available Ark plans and credentials without manually copying API keys.

## Area

- [x] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Feature
- [x] Test / harness
- [x] Documentation

## Flow

1. User selects `Continue with Volcengine`.
2. OpenAI4S starts the official Ark CLI authorization flow.
3. After authorization, it reads the selected Project and available Ark resources.
4. One usable plan/key is selected automatically; multiple choices require user selection.
5. Missing plans, keys, seats, or quota remain visible with an explicit next action.
6. OpenAI4S rechecks the account and stores the selected credential through the local SecretBroker.

## Verification

Ran:

- `git diff --check main...feat/volc-sso`
- Manual Volcengine authorization, Project setup, plan/key discovery, and recheck flow

Not run:

- `uv run pytest`
- `uv run pre-commit run --all-files`
- `python scripts/check_directory_readmes.py`

Reason:

Python and `uv` were unavailable in the authoring environment. CI should run the full offline gates.

## Notes

- Requires the official `@volcengine/ark-cli`.
- The selected Project is reused from Ark CLI; no `default` Project is assumed.
- Raw Ark API keys and cloud Key IDs are never returned to the browser.
- Live Volcengine access is not required by the offline tests.

## Risk and Reviewer Focus

- Review the Ark CLI bridge and Volcengine resource projection.
- Review the Gateway and Web UI onboarding states.
- Confirm credential handling and incomplete-resource states.

### Diff Readiness

- [x] This PR is single-purpose.
- [x] No unrelated formatting or broad rewrite is included.
- [x] Hotspot files were edited surgically.

### Tests

- [x] Manual checks are listed above.
- [x] No new test requires live LLM, network, GPU, SSH, Docker, lab hardware, or secrets.
- [x] No tests were deleted without replacements.
- [x] Gateway changes include focused connector tests.

### Core Dependency Policy

- [x] No hard third-party import was added to the core engine or stdlib web server.
- [x] The official Ark CLI requirement is documented.

- [x] No secrets, API keys, tokens, real credentials, or local paths are included.
- [x] Raw Ark API keys are not returned to the browser.
- [x] Security-sensitive behavior has focused regression tests.

### Docs

- [x] User-facing behavior is documented.
- [x] `README.md` and `README_zh.md` stay in sync.


openai4s/server/gateway.py
```diff
@@ -11896,6 +11898,7 @@ def make_handler(cfg: Config, hub: WSHub, runner: SessionRunner):
         cfg,
         providers=provider_specs,
     )
+    volcengine_connector = VolcengineConnectorService()
```
Nobody-Zhang, line 11950: **Offline suite spawns real arkcli (including `auth login`)**
`CONFIRMED · conventions` — /code-review xhigh

`make_handler` wires a real `VolcengineConnectorService` (whose `ArkCliBridge` falls through to `shutil.which("arkcli")`), and the route sweep in `tests/test_response_contract_coverage.py` (`EXEMPT = {}`) drives all `/volcengine/*` routes against it with no stub or env pin.

Empirically reproduced: with a fake `arkcli` on `PATH`, `uv run pytest tests/test_response_contract_coverage.py -q` invoked `arkcli --version`, `auth whoami --format json` (twice), and `arkcli auth login --no-browser --format json`. On a machine that followed this PR's own startup guide (`npm i -g @volcengine/ark-cli`), plain `uv run pytest` reaches the Volcengine control plane and starts a real SSO authorization; a schema-capture run would freeze that machine's live account state into `docs/response-schemas.json`.

Fix applied: `tests/conftest.py` pins `OPENAI4S_ARKCLI_PATH` to a nonexistent sentinel (module-level + per-test), so `executable()` resolves to "" everywhere and the sweep records the deterministic `not_installed` projection; a regression test asserts the pin wins even with an arkcli on PATH.

openai4s/server/volcengine_arkcli.py
```diff
@@ -0,0 +1,663 @@
...
+        if result.cancelled:
+            raise ArkCliError("login_cancelled", "Volcengine login was cancelled")
+        if result.timed_out:
+            raise ArkCliError("arkcli_timeout", "Ark CLI did not finish in time")
+        if result.returncode != 0:
+            detail = (result.stderr + "\n" + result.stdout).lower()
+            if "project" in detail and (
+                "non-interactive" in detail
+                or "noninteractive" in detail
+                or "éžäº¤äº’å¼ç»ˆç«¯" in detail
```
Nobody-Zhang: **Chinese literals are mojibake — both locale branches are dead**
`CONFIRMED · correctness` — /code-review xhigh

This literal (and the `授权` in the redaction regex at line 244) is UTF-8 double-encoded mojibake (`éžäº¤äº’å¼ç»ˆç«¯`), so it can never match real Chinese-locale Ark CLI output. A Chinese-locale CLI that still needs Project setup raises generic `arkcli_failed` (503) instead of `project_selection_required` (409), and the UI shows "连接失败，请重试" instead of the dedicated `arkcli auth login volc-sso` recovery guidance — for the product's primary audience. The 授权-code redaction rule is equally dead.

Fix applied: both literals repaired to proper UTF-8 (`非交互式终端`, `授权码`), the redaction rule extended to match `授权码` without an ASCII `code` label and with full-width colons; regression test drives a Chinese failure through the bridge and asserts `project_selection_required`.

openai4s/server/volcengine_connector.py
```diff
@@ -0,0 +1,849 @@
...
+            "plan_key": "platform",
+            "has_api_key": None,
+            "error_code": "ark_profile_ambiguous",
+        }
+    key_choice_profiles = [
+        profile
+        for profile in platform_profiles
+        if _profile_key_state(profile) == "key_choice_required"
+    ]
+    if len(key_choice_profiles) == 1:
```
Nobody-Zhang, line 291: **Multi-key + multi-endpoint platform accounts loop forever**
`CONFIRMED · correctness` — /code-review xhigh

`_access_projection` surfaces `key_choice_required` (without `endpoint_choices`) before any endpoint state is consulted; the panel's key-choice branch submits only `api_key_choice`; `provisioning_material` then raises `ark_endpoint_choice_required` (409); the UI's recovery refresh recomputes `key_choice_required` again. The endpoint chooser is unreachable while a key choice is pending and no call site ever sends both choices, so an account with multiple usable keys and multiple invocable endpoints (no defaults) can never finish setup.

Fix applied: the `key_choice_required` projection now carries `endpoint_choices` when the profile also needs one, and the panel renders both selects and submits both choices in one configure; regression test covers the combined resolution.

openai4s/server/webui/app.js
```diff
@@ -11732,6 +11876,394 @@ function modelProtocolOptions(served) {
...
+}
+
+async function refreshVolcengine(root, { autoConfigure = true, announce = false } = {}) {
+  const next = await api("/volcengine/refresh", { method: "POST" });
+  const plans = (next.plans || []).filter(plan => plan && plan.available !== false);
+  const accessState = (next.access || {}).state;
+  const checkFailed = ["check_failed", "key_check_failed", "endpoint_check_failed"].includes(accessState);
+  root._volcRefreshMessage = announce && !checkFailed ? t("cust.volc.rechecked") : "";
+  renderVolcenginePanel(root, next);
+  if (autoConfigure && !next.configured && plans.length === 1 && accessState === "ready") {
```
Nobody-Zhang: **Recheck auto-configure silently overrides the user's active profile**
`CONFIRMED · correctness` — /code-review xhigh

`configured` means "the Volcengine profile is the currently active profile", so for a user who linked Volcengine and then deliberately activated a different model profile, `!next.configured && plans.length === 1 && accessState === "ready"` all hold — clicking the read-labeled 重新检查/Recheck button (or a background key-poll tick) re-activates the Ark profile and rewrites the instance-wide LLM settings without confirmation.

Fix applied: auto-configure now also requires `!next.linked` — it runs only during first-time onboarding (and the key-poll continue), never after the user has switched away.

openai4s/server/gateway.py
```diff
@@ -12014,6 +12017,135 @@ def _doubao_search_config_payload() -> dict[str, Any]:
...
+            None,
+        )
+        profile_body = {
+            "name": name,
+            "provider": "ark",
+            "base_url": base_url,
+            "model": model,
+            "api_key": material.api_key,
+        }
+        if existing is None:
```
Nobody-Zhang, line 12191: **Concurrent configure duplicates the profile and orphans a live credential**
`CONFIRMED · correctness` — /code-review xhigh

The read (`volcengine_model_profile_id`) → check (`existing is None`) → `model_profiles.create(...)` → `set_setting` sequence is not serialized in a `ThreadingHTTPServer`. The UI can genuinely produce concurrent configures (the 5s key-poll and the Recheck button both auto-configure with no in-flight guard): both threads create a profile holding a live Ark key, the setting keeps one id, and `/volcengine/disconnect` deletes only the profile the setting names — the orphan with a working credential survives disconnect indefinitely.

Fix applied: `_configure_volcengine` now runs under a dedicated `threading.Lock`, and `configureVolcengine` in the panel has an in-flight guard.

openai4s/server/volcengine_arkcli.py
```diff
@@ -0,0 +1,663 @@
...
+    if not detail:
+        return ""
+    detail = re.sub(r"https?://[^\s<>\"']+", "[redacted URL]", detail)
+    detail = re.sub(
+        r"(?i)((?:authorization|æŽˆæƒ)?\s*code\s*[:=]\s*)[A-Za-z0-9+/=_-]{8,}",
+        r"\1[redacted]",
+        detail,
+    )
+
+    def redact_token(match: re.Match[str]) -> str:
```
Nobody-Zhang, line 331: **Purely alphanumeric secrets in CLI errors reach the browser**
`CONFIRMED · correctness` — /code-review xhigh

`redact_token` only masks 8+-char tokens containing one of `=+/_-`, and the targeted authorization-code rule requires the ASCII word `code` (its Chinese branch is mojibake). An unpadded alphanumeric authorization code echoed in stderr (e.g. `授权码 YWJjZGVmZ2hp 已过期`) flows verbatim into `ArkCliError.message` → the gateway error body → the stored login `error_detail` → the browser, violating this module's own "non-secret reason" contract. The shared `openai4s.host.bash.redact_shell_text` (Bearer / sk- / ark- shapes, NAME=value assignments) is not composed here, so credential shapes added there never protect this surface.

Fix applied: `redact_shell_text` is now composed as the baseline, the code rule matches `授权码`/full-width colons, and the fallback also masks digit+letter runs (≥10) and long mixed-case runs (≥12); regression test asserts a letters-only token is masked.

openai4s/server/volcengine_arkcli.py
```diff
@@ -0,0 +1,663 @@
...
+    argv: Sequence[str],
+    timeout_s: float,
+    cancel_event: threading.Event | None = None,
+) -> CommandResult:
+    """Run one fixed argv with bounded stdout/stderr and cancellable waiting."""
+
+    creationflags = 0
+    if os.name == "nt":
+        creationflags = int(getattr(subprocess, "CREATE_NO_WINDOW", 0))
+    process = subprocess.Popen(
```
Nobody-Zhang, line 136: **Popen OSError escapes the ArkCliError envelope into a 500**
`CONFIRMED · correctness` — /code-review xhigh

`executable()` validates configured/packaged paths with `is_file()` only (no `X_OK`), and nothing on the path catches `OSError` from `subprocess.Popen`: `_run` wraps nothing, `version()`/`connection()`/the routes catch only `ArkCliError`. `OPENAI4S_ARKCLI_PATH` pointing at a non-executable file, or an npm shim with a dead interpreter, turns every `GET /volcengine/connection` (each Models-pane open) into a generic 500 instead of the projected `not_installed`/`error` state.

Fix applied: `_run` wraps the runner call and converts `OSError` to `ArkCliError("arkcli_failed", ...)`; regression test asserts a `PermissionError`-raising runner yields the projected `error` state.

openai4s/server/volcengine_connector.py
```diff
@@ -0,0 +1,849 @@
...
+    def refresh(self) -> dict[str, Any]:
+        """Recheck live Ark resources once and replace the cached projection."""
+
+        # api_key_inventory and endpoint_inventory both query the control plane
+        # directly.  A second profile refresh plus a second full connection scan
+        # only doubled the browser wait without making this projection fresher.
+        self.invalidate()
+        return self.connection(force=True)
+
+    def start_device_login(self) -> dict[str, Any]:
```
Nobody-Zhang, line 688: **A stale pending login is reused forever; start also races**
`CONFIRMED · correctness` — /code-review xhigh

`start_device_login` returns the cached `awaiting_code` login without checking `expires_at`, so after the ~10-minute TTL every "Continue with Volcengine" click returns the same dead URL — and the awaiting-code panel renders no cancel control, so the only in-UI escape is deliberately submitting a bogus code. Separately, the idle check and the state write are split across lock scopes with a 60s subprocess between them: a double-click on Connect (never disabled) spawns two concurrent `arkcli auth login` runs whose pending OAuth state overwrite each other.

Fix applied: an expired pending login now starts fresh, the flow is claimed under one lock scope (placeholder state) so concurrent starts collapse into one, and the awaiting-code panel gained a cancel button; regression test covers the expiry re-mint.

openai4s/server/volcengine_connector.py
```diff
@@ -0,0 +1,849 @@
...
+                self._login = {
+                    "state": "failed",
+                    "login_id": login_id,
+                    "error_code": error.code,
+                    "error_detail": _text(error.message, 240),
+                    "method": "browser_oauth",
+                }
+            raise
+        with self._lock:
+            self._login = {
```
Nobody-Zhang: **Cancel doesn't cancel — a cancelled login can flip to succeeded**
`CONFIRMED · correctness` — /code-review xhigh

`POST /volcengine/login/cancel` only flips the in-memory dict; no caller ever passes the `cancel_event` that `_run_bounded` supports, so the 90s `arkcli auth login --code` subprocess always runs to completion — and `complete_device_login`'s final write sets `succeeded`/`failed` without rechecking state, so an acknowledged cancellation is silently overwritten when the CLI finishes anyway.

Fix applied: a per-login `threading.Event` is passed through `login_device_complete` into the runner (killing the subprocess on cancel), and both final writes are compare-and-set on `login_id` + `state == "connecting"`; regression test asserts a cancel during token exchange stays cancelled.

openai4s/server/volcengine_connector.py
```diff
@@ -0,0 +1,849 @@
...
+
+    def complete_device_login(self, code: Any) -> dict[str, Any]:
+        with self._lock:
+            if self._login.get("state") != "awaiting_code":
+                raise ArkCliError(
+                    "device_login_not_pending", "No device login is waiting for a code"
+                )
+            login_id = self._login.get("login_id")
+            authorize_url = str(self._login.get("authorize_url") or "")
+            self._login = {
```
Nobody-Zhang, line 744: **A local paste-format error destroys the still-valid pending authorization**
`CONFIRMED · correctness` — /code-review xhigh

The awaiting_code record (authorize_url, bound OAuth state) is replaced with the `connecting` record *before* `_normalize_device_code` runs. A pasted code with an embedded newline from a line-wrapped copy (outer `trim()` doesn't remove it) fails the regex, the except arm marks the login `failed`, and the user must redo the whole Volcengine authorization even though both the pending session and their code were valid.

Fix applied: normalization/validation now happens before the state transition, so a local rejection leaves the pending authorization intact; regression test asserts the same login_id survives an invalid paste.

openai4s/server/gateway.py
```diff
@@ -13888,6 +14020,82 @@ def _api(self, method: str, sub: str) -> None:
...
+                            self._body().get("code")
+                        )
+                    )
+                except ArkCliError as error:
+                    _raise_volcengine_error(error)
+                return
+            if sub == "/volcengine/login/cancel" and method == "POST":
+                self._json(volcengine_connector.cancel_login())
+                return
+            if sub == "/volcengine/configure" and method == "POST":
```
Nobody-Zhang, line 14136: **ModelProfileError in configure surfaces as an opaque 500**
`CONFIRMED · correctness` — /code-review xhigh

This route catches only `ArkCliError`, unlike every sibling model-profile route which converts `ModelProfileError` to its `status_code`. `_store_key`'s "could not store the API key securely; it was not saved" (500 with a real message) or `edit()`'s 404 (profile tombstoned between the lookup and the edit) bypass `_raise_volcengine_error` and reach the generic handler — the browser shows "Automatic setup failed: HTTP 500" with the actionable message discarded.

Fix applied: the route now also catches `ModelProfileError` and returns `{error}` with `exc.status_code`, matching the sibling routes.

openai4s/server/gateway.py
```diff
@@ -13888,6 +14020,82 @@ def _api(self, method: str, sub: str) -> None:
                 self._json({"service": "openai4s", "ok": True})
                 return
 
+            # ---- Volcengine account connection through the official Ark CLI ----
+            if sub == "/volcengine/connection" and method == "GET":
```
Nobody-Zhang, line 14086: **Open connection GET leaks the pending login envelope to non-admin team members**
`PLAUSIBLE · security` — /code-review xhigh

`/volcengine/connection` is deliberately left readable in team mode, but its payload embeds `_public_login()` — including a pending login's `authorize_url` (with its OAuth `state` parameter), `expires_at`, and `error_detail`. While an operator's sign-in is in flight, any team member can read the authorize URL and complete authorization with their own Volcengine account, burning or racing the state-bound pending authorization; failed attempts' `error_detail` is also visible to every member. The team_policy comment justifying the open read considered identity/plans/usage, not the live login envelope.

Fix applied: for non-admin team members the route now projects `login` down to `{state}` only.

openai4s/server/webui/app.js
```diff
@@ -11732,6 +11876,394 @@ function modelProtocolOptions(served) {
...
+    }
+    renderVolcenginePanel(root, { ...state, _error: t("cust.volc.configureFailed", apiErrorText(error)) });
+  }
+}
+
+async function startVolcengineLogin(root) {
+  // Reserve a tab during the user gesture so popup blockers do not swallow the
+  // later navigation while the local API prepares the authorization URL.
+  let authWindow = null;
+  try { authWindow = window.open("about:blank", "_blank"); } catch (_) { /* Use the fallback button. */ }
```
Nobody-Zhang, line 12558: **SSO popup keeps window.opener — reverse tabnabbing**
`CONFIRMED · security` — /code-review xhigh

The tab reserved during the user gesture is opened without `noopener` and then navigated via `target.location.href`, so the authorization tab (and anything its redirect chain lands on) holds a live `opener` reference onto the workbench; `opener.location = ...` is not blocked cross-origin. Only the fallback branch uses `noopener,noreferrer`.

Fix applied: `authWindow.opener = null` immediately after the reservation — the handle stays usable for the later `location.href` assignment, and the popup-blocker rationale is preserved.

openai4s/server/volcengine_arkcli.py
```diff
@@ -0,0 +1,663 @@
...
+        "USER",
+        "USERPROFILE",
+        "WAYLAND_DISPLAY",
+        "WINDIR",
+        "XDG_CONFIG_HOME",
+        "XDG_DATA_HOME",
+        "XDG_RUNTIME_DIR",
+    }
+)
+_DEVICE_CODE = re.compile(r"^[A-Za-z0-9+/=_-]{8,8192}$")
```
Nobody-Zhang: **`_DEVICE_CODE` rejects the plaintext `code=...&state=...` envelope**
`CONFIRMED · correctness` — /code-review xhigh

The charset excludes `&`, so the plaintext envelope that `_device_payload`'s first candidate branch was explicitly written to parse is rejected as `invalid_authorization_code` at the regex gate — that documented support path is unreachable from both entry points. A user pasting the decoded querystring exactly as shown is told the code is invalid with no hint that only the Base64 form is accepted.

Fix applied: `&` added to the charset, and `_normalize_device_code` re-wraps a pasted plaintext envelope into the Base64 form the CLI expects; regression test covers the round trip.

openai4s/server/volcengine_connector.py
```diff
@@ -0,0 +1,849 @@
...
+                    "authorize_url",
+                    "expires_at",
+                    "error_code",
+                    "error_detail",
+                    "method",
+                    "phase",
+                }
+            }
+
+    def connection(self, *, force: bool = False) -> dict[str, Any]:
```
Nobody-Zhang, line 545: **No single-flight: concurrent scans double the N+1 CLI spawns**
`CONFIRMED · efficiency` — /code-review xhigh

`connection()` releases the lock for the entire bridge scan, so concurrent cache-miss or forced callers (the 5s key-poll plus a Recheck click, two tabs, or a configure's forced scan) each spawn the whole arkcli chain — itself an N+1 of subprocesses (2 CLI calls per profile plus endpoint inventory) — and compute identical projections: double the Node spawns and control-plane calls, two gateway threads pinned for tens of seconds, for one 60s-TTL cache entry. (The deeper N+1 — probing every profile, re-fetching inventories inside `provisioning_material`, and the poll using the heaviest endpoint — is left as a follow-up.)

Fix applied: a single-flight slot under the existing lock — later callers wait for the running scan and share its payload; regression test asserts two concurrent forced callers produce one scan.

openai4s/server/gateway.py
Nobody-Zhang: Fixed in cb53c03 — `tests/conftest.py` now pins `OPENAI4S_ARKCLI_PATH` to a nonexistent sentinel (module-level + per-test), so the sweep and every capture record the deterministic `not_installed` projection on any machine. Regression test: `test_the_offline_suite_never_resolves_a_real_arkcli` (fails if the pin is removed while an arkcli is on PATH).

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_arkcli.py
Nobody-Zhang: Fixed in cb53c03 — both literals repaired to proper UTF-8 (`非交互式终端`, `授权码`), and the redaction rule now matches `授权码` without an ASCII `code` label and with full-width colons. Regression test: `test_chinese_project_selection_failure_maps_to_the_recovery_state`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_connector.py
Nobody-Zhang: Fixed in cb53c03 — the `key_choice_required` projection now carries `endpoint_choices` when the platform profile also needs one, and the panel renders both selects and submits both choices in a single configure. Regression test: `test_key_choice_projection_surfaces_a_pending_endpoint_choice` (also exercises the combined provisioning resolution).

CI on the updated head is green (all 31 checks passed).

openai4s/server/webui/app.js
Nobody-Zhang: Fixed in cb53c03 — auto-configure now additionally requires `!next.linked`, so it only runs during first-time onboarding (and the key-poll continue); once the Volcengine profile exists, Recheck/refresh never re-activates it over a deliberately selected different profile.

CI on the updated head is green (all 31 checks passed).

openai4s/server/gateway.py
Nobody-Zhang: Fixed in cb53c03 — `_configure_volcengine` is now serialized under a dedicated `threading.Lock` on the gateway side, and `configureVolcengine` in the panel has an in-flight guard, so concurrent configures can no longer create a duplicate profile holding an orphaned live credential.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_arkcli.py
Nobody-Zhang: Fixed in cb53c03 — `redact_shell_text` (the shared approval/audit redactor) is now composed as the baseline, the code rule matches `授权码`/full-width colons, and the fallback also masks digit+letter runs (≥10 chars) and long mixed-case runs (≥12 chars). Regression test: `test_purely_alphanumeric_secrets_in_cli_errors_are_redacted`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_arkcli.py
Nobody-Zhang: Fixed in cb53c03 — `_run` now wraps the runner call and converts `OSError` (PermissionError, FileNotFoundError, ENOEXEC) into `ArkCliError("arkcli_failed", …)`, so the routes project the designed error state instead of a generic 500. Regression test: `test_an_unrunnable_executable_is_a_controlled_failure_not_a_crash`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_connector.py
Nobody-Zhang: Fixed in cb53c03 — an expired pending login now starts fresh (expires_at is checked), the flow is claimed inside one lock scope via a placeholder state so concurrent starts collapse into a single CLI run, and the awaiting-code panel gained a cancel button. Regression test: `test_an_expired_pending_login_mints_a_fresh_authorization`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_connector.py
Nobody-Zhang: Fixed in cb53c03 — a per-login `threading.Event` is now threaded through `login_device_complete` into the runner (the in-flight `arkcli auth login --code` subprocess is killed on cancel), and both final state writes are compare-and-set on `login_id` + `state == "connecting"`, so an acknowledged cancellation can never flip to succeeded/failed. Regression test: `test_cancel_during_token_exchange_stays_cancelled`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_connector.py
Nobody-Zhang: Fixed in cb53c03 — normalization/validation now runs before the state transition, so a purely local paste-format rejection leaves the pending authorization (same `login_id`, same authorize_url) intact. Regression test: `test_a_local_paste_format_error_keeps_the_pending_authorization`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/gateway.py
Nobody-Zhang: Fixed in cb53c03 — the `/volcengine/configure` route now also catches `ModelProfileError` and returns `{error}` with `exc.status_code`, matching the sibling model-profile routes.

CI on the updated head is green (all 31 checks passed).

openai4s/server/gateway.py
Nobody-Zhang: Fixed in cb53c03 — for non-admin team members the connection GET now projects `login` down to `{state}` only, so a pending sign-in's `authorize_url` (with its OAuth state parameter), `expires_at`, and `error_detail` never leave the operator's view.

CI on the updated head is green (all 31 checks passed).

openai4s/server/webui/app.js
Nobody-Zhang: Fixed in cb53c03 — `authWindow.opener = null` immediately after the tab reservation; the handle stays usable for the later `location.href` navigation, so the popup-blocker workaround is preserved while the reverse handle is severed.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_arkcli.py
Nobody-Zhang: Fixed in cb53c03 — `&` added to the `_DEVICE_CODE` charset, and `_normalize_device_code` now re-wraps a pasted plaintext `code=…&state=…` envelope into the Base64 form the CLI expects. Regression test: `test_plaintext_envelope_paste_is_accepted_and_rewrapped`.

CI on the updated head is green (all 31 checks passed).

openai4s/server/volcengine_connector.py
Nobody-Zhang: Fixed in cb53c03 — `connection()` now has a single-flight slot under the existing lock: later callers (including forced ones) wait for the running scan and share its payload instead of spawning a second identical CLI chain. Regression test: `test_connection_scan_is_single_flight_across_threads`. The deeper per-profile N+1 and the poll's heavy probe remain a follow-up, as noted.

CI on the updated head is green (all 31 checks passed).

Nobody-Zhang: Follow-up landed in 5a0de57 — the deeper N+1 is now addressed too: the scan's three independent legs (plans/usage/profiles) run concurrently, `provisioning_material` reuses the forced scan's probed key/endpoint inventories instead of re-spawning the CLI, and an explicit key choice skips the inventory preflight entirely (`get_raw` is the authority, with its failure mapped to `ark_key_choice_invalid`). Regression tests: `test_provisioning_reuses_the_scan_probes_without_re_spawning_the_cli`, `test_an_explicit_key_choice_skips_the_inventory_preflight`. CI is green on the new head (all 31 checks).

Nobody-Zhang:
**Review follow-up (5a0de57):** the secondary observations from the xhigh review — below its 15-finding reporting cap — are now fixed in this PR as well:

- `endpoint_inventory` accepts a top-level-list payload and string/numeric `invocable` flags; `whoami` fields tolerate snake/camel/PascalCase like every other projection
- seat errors are recognized in Chinese (席位); an untagged aggregate usage row no longer exhausts or seat-blocks every plan (tagged rows win, untagged is the fallback)
- `version()` no longer caches a transient probe failure forever; `ProvisioningMaterial` hides `api_key` from `repr`
- proxy env vars (HTTP(S)_PROXY/NO_PROXY/ALL_PROXY) reach the arkcli child; Windows npm `.cmd` shims are bypassed via the resolved Node entry script (`_resolve_batch_shim`), avoiding cmd.exe argv corruption
- efficiency: the scan's three legs run concurrently, provisioning reuses the scan's probed inventories, and an explicit key choice skips the inventory preflight — one "Use this plan" click drops from ~8–12 CLI spawns to ~2–3
- route-status tests pin the 400/409/503 error mapping; dead i18n keys and the orphan `.volc-wait` CSS rule are removed

19 new regression tests; all 31 CI checks green on 5a0de57.

PR #130, opened 2026-08-27
fix(windows): harden WSL launcher and runtime

Nobody-Zhang:
## Summary

- Harden the supported Windows launcher path around WSL2 installation, distro selection, bind/client host selection, mirror disable semantics, and daemon lifecycle state.
- Preserve Linux shell LF endings in Windows checkouts and improve WSL/NAT failure guidance.
- Fix bubblewrap process-session ownership, artifact change detection, and partial-clone error handling found during Windows acceptance.
- Add narrowly scoped Clash-style Fake-IP DNS compatibility for approved public domains without weakening literal-IP, private-address, or cloud-metadata SSRF protection.
- Add regression coverage and update Windows/security documentation.

## Area

- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Test / harness
- [x] Documentation
- [x] Security-related change
- [x] Release / packaging

## Verification

Ran:

```text
# 7335 passed, 103 skipped

uv run python3 -m harness.cli run --tier pr --offline
# 38/38 scenarios passed

# 191/191 routes passed

# 1054 response shapes checked; no breaking changes

# 3284 files passed

uv build
uv run python scripts/verify_release_artifacts.py dist
git diff --check

# Additional Windows checks
PowerShell 5.1 Parser.ParseFile(scripts/windows/openai4s.ps1)
wsl.exe -d Ubuntu-24.04 --exec sh -n scripts/windows/bootstrap.sh
```

Manual checks:

- Started the Windows package through WSL2 and completed a real browser task with an Ark-backed model.
- Verified an actual Python Cell execution returned stdout `899`, with bubblewrap enforcement and sandbox self-test active.
- Verified trusted Fake-IP DNS can reach OpenAlex and web search while literal benchmark-range IPs, private addresses, and cloud metadata remain blocked.
- Verified daemon status/url use only a live, generation-matched sidecar and that temporary test services and credentials were cleaned up.

Not run:

```text
uv run python -m harness.smoke.linux_bwrap_interrupt
```

Reason:

The change targets the Windows/WSL2 package and was exercised there with a real Chromium UI and enforced bubblewrap. Container, cross-engine, and hosted-Linux-only jobs remain for CI.

## Risk and Reviewer Focus

- Review `scripts/windows/openai4s.ps1` and `scripts/windows/bootstrap.sh` first: bind-vs-client addressing, WSL distro names containing spaces, explicit mirror disable behavior, and Fake-IP auto-detection.
- Review `openai4s/webtools.py` with `openai4s/egress.py`: synthetic `198.18.0.0/15` answers require launcher opt-in and a catalogued or explicitly approved hostname; IP literals and other private ranges remain blocked.
- Review `openai4s/cli/main.py` for stale/corrupt daemon-sidecar rejection and exact PID-start matching.
- Review kernel process-session cleanup and artifact digesting for restored-mtime/same-size writes.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched.

### Tests

- [x] No new default-suite test requires a live LLM, network, GPU, SSH, Docker, lab hardware, or secrets.
- [x] Kernel/runtime changes include focused contract coverage and a documented smoke test.

### Core Dependency Policy

- [x] Optional science imports remain guarded.
- [x] No new dependency was added.

- [x] No unpublished research roadmap, internal assignment, benchmark detail, or unreleased result is disclosed.
- [x] Security-sensitive changes include synthetic tests and manual verification.

### Docs

- [x] User-facing behavior changes are documented.
- [x] English and Chinese Windows documentation remains synchronized.


openai4s/server/artifacts.py
```diff
@@ -3047,18 +3055,40 @@ def _live_fingerprint(path: Path) -> WorkspaceFileState | None:
         """Identity of the exact regular live file a child already captured."""
 
         try:
-            status = path.stat(follow_symlinks=False)
+            descriptor = os.open(
```
Nobody-Zhang, line 3142: **correctness · CONFIRMED — the Cell boundary can now hang forever**

`lstat` never blocked; `os.open(..., O_RDONLY)` on a writer-less FIFO blocks until a writer appears, and the `S_ISREG` rejection that used to make every non-regular entry free now runs *after* the open, so it cannot save it.

`snapshot()` walks `workspace.rglob("*")` over a tree the agent controls, and `_ignored_file` only filters dot-paths/`__pycache__`/`.pyc`. One `os.mkfifo("stream.fifo")` — an ordinary samtools/bcftools streaming pattern — parks `snapshot()` with no timeout on the synchronous Cell-completion path. The FIFO persists, so every later Cell in that session re-wedges.

Measured, same inputs:
```
OLD stat(fifo) -> None, instantly
NEW os.open(fifo, O_RDONLY|O_NOFOLLOW|O_CLOEXEC) -> still blocked after 8s
BASE snapshot(dir) -> returned;  PR snapshot(dir) -> never returns
```

`O_NONBLOCK` is the missing flag: it returns a valid fd immediately for a FIFO (then `S_ISREG` rejects it) and is a no-op for regular files.

openai4s/security/sandbox.py
```diff
@@ -1017,6 +1019,14 @@ def wrap_command(self, command: Sequence[str]) -> list[str]:
...
                 read_isolation=self._read_isolation,
                 info_fd=info_fd,
+                # PipeTransport starts the bwrap launcher in a new session.
+                # Asking bwrap to create another session for its command splits
+                # the wrapper and Cell subprocesses into different process
+                # groups, so a watchdog kill reaches only the wrapper and
+                # leaves the actual work alive. Inheriting the Popen-created
+                # session preserves terminal isolation and gives the transport
+                # one group that contains the complete worker tree.
+                new_session=False,
```
Nobody-Zhang, line 1039: **security · CONFIRMED — this is set on a shared wrapper, but only one of its three callers creates a session**

The justification is written entirely about `PipeTransport`, and it is true there (`kernel/transport.py` sets `start_new_session=True`). It is false for the other two callers of `KernelSandbox.wrap_command`:

- `openai4s/tools/dynamic.py:338` — `subprocess.run(command, input=…, capture_output=True, …)`, no `start_new_session`
- `openai4s/kernel/preinstall.py:507` — `subprocess.run(argv, capture_output=True, …)`, no `start_new_session` **and** no stdin redirect

`wrap_bwrap_command` mounts `--dev /dev`, and bubblewrap's `--dev` provides `/dev/tty`. With the daemon launched in the foreground (`start.sh` ends in `exec … serve`), model-authored dynamic-tool code — the module whose own docstring calls it "the most direct way to get new code onto the machine" — now holds the operator's controlling terminal and can `ioctl(TIOCSTI)` into it. That is exactly the escape `--new-session` exists to close.

The benefit is real (`--new-session` would `setsid()` the worker out of `self._pgid`, so `_stop_group_or_leader`'s SIGKILL would reach only the wrapper) — so the fix is to make the session a precondition at *every* spawn site, not to keep the flag.

openai4s/webtools.py
```diff
@@ -113,7 +146,8 @@ def _guard_url(url: str) -> None:
     if _host_is_private(host):
         raise SSRFBlocked(
             f"refusing to fetch a private/loopback/metadata address: {host!r} "
-            "(set OPENAI4S_ALLOW_PRIVATE_FETCH=1 to allow)"
+            "(the Windows launcher auto-detects trusted Fake-IP DNS; set "
```
Nobody-Zhang, line 161: **security · CONFIRMED — the guard inspects a host the client never dials**

Pre-existing, but this is the guard the PR is widening and the docs it adds claim "loopback, link-local, metadata … still fail closed". Percent-encoding the dots walks straight past it:

```
blocked : http://169.254.169.254/latest/meta-data/
ALLOWED : http://169%2e254%2e169%2e254/latest/meta-data/
ALLOWED : http://127%2e0%2e0%2e1:8760/

guard hostname  : '169%2e254%2e169%2e254'   -> getaddrinfo raises -> fail-open
urllib req.host : '169.254.169.254'         -> connects
```

`urllib.request.Request._parse` does `self.host = unquote(self.host)` and `requests.PreparedRequest.prepare_url` normalizes the same way, so both clients decode before connecting. `tests/test_host_network_capability.py` asserts the *decoded* spelling is blocked; the encoded one is not. Default posture is `OPENAI4S_EGRESS=off`, so `check_url` is a no-op and this is the only control.

Fix: `urllib.parse.unquote(...)` the hostname before `_host_is_private`.

openai4s/server/artifacts.py
```diff
@@ -3047,18 +3055,40 @@ def _live_fingerprint(path: Path) -> WorkspaceFileState | None:
...
+                path,
+                os.O_RDONLY
+                | getattr(os, "O_CLOEXEC", 0)
+                | getattr(os, "O_NOFOLLOW", 0),
+            )
+            status = os.fstat(descriptor)
```
Nobody-Zhang: **correctness · CONFIRMED — an unreadable regular file now vanishes from provenance**

`lstat` needs only search permission on the parent; `open` needs read permission on the file. The bare `except OSError: return None` turns `EACCES` into "not a live file", so the path is absent from **both** the before and after snapshots and can never register as changed.

```
mode-000 regular file:  OLD stat() -> (dev, ino, 12, …)   NEW os.open() -> EACCES -> None
```

Consequences: a deliverable written into a `0o300` directory, or by a root/container step, is silently never captured; and a Cell can hide a file from Artifact capture with one `chmod 000`. Second blast radius — `_matches_delegated_claim` computes `current = self._live_fingerprint(path)`, so for such a file `current != claim.fingerprint` always holds and the delegated-claim guard degrades to a no-op.

Falling back to a metadata-only fingerprint (`digest=None`) restores exactly the pre-PR identity for these files.

openai4s/egress.py
```diff
@@ -234,18 +258,7 @@ def domain_allowed(host_or_url: str) -> bool:
...
-    exact = exact_builtin_domains()
-    if host in exact:
-        return True
-    # Re-approving an exact managed origin at runtime must not silently turn
-    # that same string back into a suffix grant for attacker-controlled child
-    # labels. Other user grants retain the documented base-domain semantics.
-    runtime = granted_domains()
-    allowed = (builtin_domains() - exact) | (runtime - exact)
-    return any(_host_matches(host, a) for a in allowed)
+    return domain_in_allowlist(host_or_url)
```
Nobody-Zhang, line 271: **correctness · CONFIRMED — a documented fail-open was inverted, and the docstring three lines above still promises it**

The deleted body was `host = domain_of(...)` / `if not host: return True` — *"let the request fail on its own rather than mis-block"*. `domain_in_allowlist` returns `False` for an empty host **by design** (its own new test says it "must not inherit `domain_allowed`'s intentional fail-open answer"), and `domain_allowed` now inherits that instead.

Verified base vs head with `OPENAI4S_EGRESS=allowlist`:
```
'file:///home/me/notes.txt'  BASE True  -> HEAD False
'https://'                   BASE True  -> HEAD False
'http://[::1'                BASE True  -> HEAD False
```

`openai4s/permissions.py:91` gates only on `"://" in target`, so a `file:///abs/path` target now makes `_guardian_hard_deny` return `True`: the action is refused as "denied by an existing hard policy" **and writes a durable audit row naming a policy that never issued** — verbatim the regression the comment at `permissions.py:82-87` was added to fix. `tests/test_permissions.py` only covers scheme-less paths, so it stays green.

Keep the `if not domain_of(...): return True` guard in `domain_allowed` before delegating.

scripts/windows/openai4s.ps1
```diff
@@ -589,29 +641,56 @@ function Test-SandboxIndependentCli([string[]] $CliArgs) {
...
-    $AppHost = $fallbackHost
-    $Url = "http://${AppHost}:${AppPort}/"
-    Write-Host "  localhostForwarding=false detected; using WSL NAT address $AppHost." -ForegroundColor Yellow
+    # A wildcard is a valid bind address but not a destination Windows can
+    # dial. Keep its listen-on-all-interfaces semantics while probing and
+    # opening the UI through WSL's routable NAT address.
+    if ($BindHost -ne '0.0.0.0') {
+        $BindHost = $fallbackHost
+    }
+    $ClientHost = $fallbackHost
```
Nobody-Zhang, line 731: **correctness · CONFIRMED — the new wildcard branch opens the one address that cannot work**

Journey: `$env:OPENAI4S_HOST='0.0.0.0'` + `[wsl2] localhostForwarding=false`.

`$BindHost` stays `0.0.0.0` and `Get-WslBootstrapArgs` exports `OPENAI4S_HOST=0.0.0.0` into the `cli … url` call too. `openai4s url` renders that through `_reachable_host`, which maps `0.0.0.0` -> `localhost`:

```
_url(cfg, endpoint=('0.0.0.0', 8760)) -> 'http://localhost:8760/'
```

So the launcher prints `ready: http://172.20.x.x:8760/` (correct, **tokenless**) and then `Open-AppUrl $appUrl` opens `http://localhost:8760/?token=…` -> ERR_CONNECTION_REFUSED, because localhost forwarding is off. Recovery is a dead end: the printed NAT URL 401s and `OpenAI4S.cmd url` returns the same `localhost` string.

This contradicts the PR's own doc line — *"the launcher probes **and opens** the routable WSL IPv4 address"*. Only the probe does. The token is in the query, so re-authority the returned URI with `$ClientHost` rather than discarding it.

scripts/windows/openai4s.ps1
```diff
@@ -213,9 +255,11 @@ function Get-WslDistros {
         # whose last field is not a number.
         $wslVersion = 0
         if (-not [int]::TryParse($fields[-1], [ref] $wslVersion)) { continue }
+        $state = $fields[-2]
+        $name = @($fields[0..($fields.Count - 3)]) -join ' '
```
Nobody-Zhang: **correctness · CONFIRMED — regression on any localized Windows**

`$fields[0..($fields.Count - 3)]` assumes STATE is exactly one whitespace-delimited field, in a function whose own next comment acknowledges the listing is localized. WSL's state string is translated, and several translations contain a space:

```
'* Ubuntu-24.04    Wird ausgefuehrt   2'   BEFORE 'Ubuntu-24.04'  AFTER 'Ubuntu-24.04 Wird'
"* Ubuntu-24.04    En cours d'execution 2" AFTER 'Ubuntu-24.04 En cours'
'  My Custom Distro   Stopped   2'         AFTER 'My Custom Distro'   <- the case this fixes
```

With a wrong name every later `wsl -d "$Distro"` fails: `Test-DistroHasInstall` false, `OPENAI4S_WSL_DISTRO` never matches, and `Invoke-Bootstrap` dies with the generic "the daemon did not start". The pre-PR `Name = $fields[0]` was immune.

`-split '\s+'` cannot tell the two cases apart, but the columns are *padded*: `-split '\s{2,}'` keeps `My Distro` whole **and** keeps `Wird ausgefuehrt` out of the name.

(The `$a[0..-1]` hazard does not occur — the `Count -lt 3` guard above makes the slice at minimum `0..0`.)

scripts/windows/bootstrap.sh
```diff
@@ -88,43 +88,66 @@ MANAGED_MARK="managed-by-openai4s-windows-launcher"
...
-    PIP_CONF="$APP/runtime/pip.conf"
+    # On a fresh install the unmarked file is the pristine bundle baseline and
+    # may be claimed. On later launches, removing the marker transfers
+    # ownership to the user, and the launcher preserves the whole file rather
+    # than guessing which individual setting was intentional.
     if [ -f "$PIP_CONF" ] && ! grep -q "$MANAGED_MARK" "$PIP_CONF" 2>/dev/null \
-        && grep -q '^index-url' "$PIP_CONF" 2>/dev/null; then
-      echo "note: $PIP_CONF has a user-managed index-url; leaving it unchanged" >&2
-      echo "      (set OPENAI4S_WSL_PYPI_INDEX to manage mirrors from the launcher)" >&2
+        && [ "$FRESH_INSTALL" != "1" ]; then
```
Nobody-Zhang, line 133: **correctness · CONFIRMED — the pristine bundle `pip.conf` becomes permanently unclaimable**

Dropping `grep -q '^index-url'` in favour of `FRESH_INSTALL` is strictly narrower. The bundle ships an *unmarked* `pip.conf`, and if the first install ran with no mirror selected (`PYPI_MODE=unchanged`) the whole block is skipped, so that file stays unmarked. Every later launch then reads "no marker, not fresh" and declares it user-managed — so setting `OPENAI4S_WSL_PYPI_INDEX` afterwards silently never takes effect, forever.

Reproduced end to end against both scripts:
```
BASE  install(no mirror) -> install(mirror)  MIRROR APPLIED? True
HEAD  install(no mirror) -> install(mirror)  MIRROR APPLIED? False
      note: …/pip.conf is user-managed; leaving it unchanged
```

The release gate that pinned the old heuristic was deleted with no replacement covering this path, and the new `note:` names no remedy (the removed second line was the only text naming the variable).

Restoring the `grep` would undo the PR's real improvement (a user file with no `index-url` is now preserved). Claiming the file on the fresh install instead — `[ "$PYPI_MODE" != "unchanged" ] || [ "$FRESH_INSTALL" = "1" ]` — keeps both.

Also: the comment above lost its second half in the edit and now dangles mid-sentence ("…is this launcher's job. A").

openai4s/server/artifacts.py
```diff
@@ -3047,18 +3055,40 @@ def _live_fingerprint(path: Path) -> WorkspaceFileState | None:
...
         except OSError:
             return None
-        if not stat.S_ISREG(status.st_mode):
+        try:
+            if not stat.S_ISREG(status.st_mode):
+                return None
+            digest = None
+            if int(status.st_size) <= _MAX_WORKSPACE_FINGERPRINT_BYTES:
```
Nobody-Zhang: **efficiency · CONFIRMED — the 8 MiB cap is per file, not per workspace**

The removed docstring guarantee was *"keeping the Cell boundary proportional to directory entries"*. There is now no aggregate bound, and `snapshot()` runs at least twice per Cell (before + after), again per writing native tool call, and again per delegated child action.

Measured on a warm page cache — the best case, not the target one:
```
200 x 4 MiB (839 MB):  BASE 0.002s   PR 0.388s     (~190x)
200 x 1 MiB:           BASE 0.3 ms   PR 98.8 ms    (~330x)
5000 x 4 KiB:          BASE 12.3 ms  PR 108.7 ms   (~9x, pure syscall overhead)
```

A 20,000 x 5 MiB genomics workspace hashes 100 GB per snapshot, 200 GB per Cell. And the comment names the cause as WSL-only ("WSL's ext4/VHD metadata can report the same ctime tick") while the code has no platform gate — macOS/Linux pay the full cost for a defect they cannot hit, and WSL DrvFs, the target, is the slowest filesystem of the three.

Not changed here: an aggregate byte budget, or gating on a `ctime == mtime` tie, is a product call. Flagging the cost.

scripts/windows/openai4s.ps1
```diff
@@ -91,9 +118,24 @@ function Write-Section([string] $Text) {
 function Open-AppUrl([string] $appUrl) {
     if ($env:OPENAI4S_NO_OPEN) {
         Write-Host '  browser open suppressed by OPENAI4S_NO_OPEN.' -ForegroundColor DarkGray
+        Write-Host '  run OpenAI4S.cmd url to print a secure browser URL.' -ForegroundColor DarkGray
```
Nobody-Zhang: **correctness · CONFIRMED — the launcher discards the tokenized URL it is already holding**

By the time `Open-AppUrl` runs, the caller has paid a `wsl.exe` round trip for `$appUrl` = `http://…/?token=…`. With `OPENAI4S_NO_OPEN` set (headless/CI/scripted), this prints "suppressed" plus the new advice line and returns without printing it. `$appUrl` is never written to the console anywhere in the file — lines 736 and 766 both interpolate `$Url`, which carries **no token** and answers 401.

The header comment advertises `OPENAI4S_NO_OPEN` as "print readiness without opening a browser"; the readiness it prints is unusable. The same applies to the new `catch` branch: the browser failed, and the answer is right there.

Print `$appUrl` instead of pointing at `OpenAI4S.cmd url`.

scripts/windows/bootstrap.sh
```diff
@@ -272,6 +343,7 @@ cli)
     echo "not installed: $APP" >&2
     exit 1
   fi
+  configure_fake_ip_dns
```
Nobody-Zhang: **efficiency · CONFIRMED — a live third-party DNS lookup in front of every CLI command**

`configure_fake_ip_dns` runs on every `cli` action, and `auto` is the default, so `status`, `url`, `stop`, `doctor` and `--help` each block on `getent ahostsv4 api.openalex.org` before doing anything.

Measured with a stub `getent` that sleeps 5s:
```
cli status (auto mode) -> status-ran FAKEIP=0 ; elapsed=5s
```

On the machines this feature targets — a Clash/TUN proxy with hijacked or stalled WSL DNS — glibc's default `timeout:5 attempts:2` is 10s per invocation. The launcher's readiness loop calls `Test-OpenAI4SServing` (`cli … status`) every 500 ms against a 60s budget, so the wait collapses to a handful of polls; and `openai4s stop`, the command whose job is to work when things are wedged, now needs DNS first. It is also an unconsented outbound signal on every launcher invocation.

The local resolver check already gates the verdict — running it *first* and only probing when it matches makes the ordinary machine read one local file and stop.

openai4s/cli/main.py
```diff
@@ -790,7 +894,13 @@ def cmd_stop(args) -> int:
...
 
 def cmd_url(args) -> int:
-    print(_url(get_config()))
+    cfg = get_config()
+    endpoint = None
+    if getattr(cfg, "pidfile", None) is not None:
+        pid = _read_pid(cfg)
+        if pid and _daemon_alive(cfg, pid):
+            endpoint = _recorded_endpoint(cfg, pid)
+    print(_url(cfg, endpoint=endpoint))
```
Nobody-Zhang: **correctness · CONFIRMED — the new live-endpoint mechanism is wired into two callers out of many**

`cmd_status` and `cmd_url` now resolve the recorded endpoint; `_daemon_request` (`main.py:1607`) still builds `url = _url(cfg, with_token=False)…` from config alone, and it is the generic caller behind ~14 daemon-backed subcommands (`/frames`, `/shares`, `/share/status`, `/orchestration/jobs`, `/orchestration/profiles`).

So in exactly the scenario this PR fixes — WSL `localhostForwarding=false`, daemon bound to the NAT IPv4 — `openai4s url` and `openai4s status` work while `openai4s share create/import/status` and the orchestration subcommands still dial `http://127.0.0.1:8760/` and fail, or reach an unrelated local service. `_require_daemon` passes, because pidfile identity is port-independent.

`_open_daemon`'s own docstring names this case ("Under WSL2 the daemon can be reached through its NAT address rather than loopback"), so the gap is inside the helper written for it. `cmd_serve`'s two "daemon already running at …" banners have the same staleness.

One shared `_live_endpoint(cfg)` resolver keeps the CLI from disagreeing with itself.

openai4s/security/sandbox.py
```diff
@@ -1017,6 +1019,14 @@ def wrap_command(self, command: Sequence[str]) -> list[str]:
                 deny_read=self._deny_read,
                 read_isolation=self._read_isolation,
                 info_fd=info_fd,
+                # PipeTransport starts the bwrap launcher in a new session.
```
Nobody-Zhang: **test-coverage · CONFIRMED — the self-test now attests to an argv nothing runs**

`_run_self_test` (`sandbox.py:1663`) still calls `wrap_bwrap_command(probe, …)` with no `new_session=`, taking the `True` default, while every real kernel / dynamic-tool / preinstall launch goes through `KernelSandbox.wrap_command` with `new_session=False`. The self-test is the gate that decides whether `auto` degrades visibly and whether `enforce` fails closed, and its whole claim is that it "proves it by establishing a boundary and probing it" — it now proves a different one.

The Windows preflight has the same split and its comment explicitly claims the parity it no longer has (`bootstrap.sh`: "Exercise the same lifecycle and namespace flags emitted by `wrap_bwrap_command()`"), and `scripts/verify_windows_zip.py:_REQUIRED_BOOTSTRAP_RUNTIME` pins `"--new-session"`, so the release gate now *requires* the divergence. The new `tests/test_kernel_sandbox.py` case pins it as intentional rather than closing it.

Not changed here: the probe validates a strict superset, so the risk is theoretical, and which side should move is the author's call.

scripts/windows/bootstrap.sh
```diff
@@ -232,7 +302,7 @@ install)
     exit 1
   fi
   printf '%s\n' "$EXPECTED" > "$MARKER"
-  configure_network "$APP"
+  configure_network "$APP" 1
```
Nobody-Zhang, line 322: **correctness · PLAUSIBLE — the `.installed` marker is written before the install finishes**

The marker is what makes the next launch take the `already-installed` fast path, and it is written *before* `configure_network`. This PR turns the `pip.conf` write from a simple redirect into a compound `{ … } > "$PIP_CONF"`, whose `set -e` behaviour on a redirection failure now differs by shell:

```
dash    -c 'set -eu; { printf x; } > /nonexistent/f; echo CONTINUED'  -> exit 2, no CONTINUED
bash-as-sh                     (same command)                        -> CONTINUED, exit 0
bash-as-sh   'set -eu; printf x > /nonexistent/f; echo CONTINUED'    -> exit 1, no CONTINUED
```

WSL Ubuntu's `/bin/sh` is dash. So on an unwritable tree (missing `runtime/`, ENOSPC, read-only mount) the install aborts *after* the marker exists: every later launch skips the extraction, re-enters `configure_network` with `FRESH_INSTALL=0`, fails identically, and there is no route back to a working install. macOS `sh` is bash, so `tests/test_release_gates.py` cannot exercise the dash path locally.

Writing the marker last makes a failed install retryable.

scripts/windows/openai4s.ps1
```diff
@@ -91,9 +118,24 @@ function Write-Section([string] $Text) {
...
+        Start-Process $appUrl
+    } catch {
+        Write-Host '  the default browser could not be opened.' -ForegroundColor Yellow
+        Write-Host '  run OpenAI4S.cmd url to print a secure browser URL.' -ForegroundColor Yellow
+    }
+}
+
+function Get-WslLogCommand([string] $Distro) {
+    $linuxLog = if ($WslDataDir) {
+        $WslLogPath
```
Nobody-Zhang: **correctness · CONFIRMED — the guidance command's two log paths already disagree**

`Get-WslLogCommand` re-derives the log path instead of using `$WslLogPath` (line 79-83), and the two fallbacks differ: `$WslLogPath` uses `'~/.openai4s/logs/app.out'`, the function uses `'$HOME/.openai4s/logs/app.out'`. Inside the `sh -lc 'tail -40 "…"'` string the function emits, `$HOME` expands and `~` inside double quotes stays literal — so the `$WslLogPath` spelling produces `tail: cannot open '~/.openai4s/logs/app.out'`. After this refactor `$WslLogPath` is read from exactly one place: the branch of `Get-WslLogCommand` that duplicates its own condition.

Two further hazards in the same emitted string, not addressed here:
- `Assert-WslDataDir` refuses `"`, `'`, `` ` ``, `$`, `\` *because* the value is interpolated into sh command strings, but permits **spaces** — and `OPENAI4S_WSL_DATA_DIR='/home/me/My Data'` breaks this new interpolation site.
- the outer `'…'` is PowerShell quoting; `OpenAI4S.cmd` runs from `cmd.exe`, which does not process single quotes, so a user pasting it there gets `Unterminated quoted string`.

This lands in all four fatal paths — every place the user is already stuck.

Nobody-Zhang:
@cursoragent Review this PR.

scripts/windows/openai4s.ps1
```diff
@@ -91,9 +120,38 @@ function Write-Section([string] $Text) {
...
+    # Windows cannot reach with localhostForwarding=false, which is exactly the
+    # case the NAT fallback exists for. The sign-in token lives in the query, so
+    # re-authority the URL rather than discarding it.
+    try {
+        $builder = New-Object System.UriBuilder $Url
+        if ($builder.Host -ne $NewHost) { $builder.Host = $NewHost }
+        return $builder.Uri.AbsoluteUri
+    } catch {
+        return $Url
+    }
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 4b4987f. The CLI passthrough now re-authorities what `url`/`status` print: when `$BindHost -ne $ClientHost` it captures the output and rewrites `http://(localhost|127.0.0.1):$AppPort` to `$ClientHost`, the same host the auto-open path already used. `_reachable_host` is untouched.

openai4s/cli/main.py
```diff
@@ -234,7 +345,12 @@ def _reachable_host(host: str) -> str:
     return "localhost" if host in ("0.0.0.0", "::", "") else host
```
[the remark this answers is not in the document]
Nobody-Zhang: Agreed on not inverting `_reachable_host` — left as is, and the gap is closed on the PowerShell side instead (see the `Set-UrlHost` thread). The two `cmd_serve` banners now pass `endpoint=_live_endpoint(cfg)` in 4b4987f.

scripts/windows/openai4s.ps1
```diff
@@ -91,9 +120,38 @@ function Write-Section([string] $Text) {
...
+    try {
+        Start-Process $appUrl
+    } catch {
+        Write-Host '  the default browser could not be opened. Open this URL:' -ForegroundColor Yellow
+        Write-Host "  $appUrl" -ForegroundColor Yellow
+    }
+}
+
+function Get-WslLogCommand([string] $Distro) {
+    return "wsl -d `"$Distro`" --exec sh -lc 'tail -40 `"$WslLogPath`"'"
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 4b4987f. `Get-WslLogCommand` drops `--exec sh -lc` for `wsl -d "X" -- tail -40 ~/...`: without `--exec` the line runs through the default shell, so `~` still expands and no single quotes are needed — cmd-pasteable. `Assert-WslDataDir` now also refuses whitespace, for the reason it already refuses quotes and `$`.

docs/windows-wsl.md
```diff
@@ -265,6 +287,10 @@ openai4s url
 openai4s stop
 ```
 
+`status` and `url` read the live endpoint recorded by the daemon, so the WSL
+IPv4 fallback for an explicit `localhostForwarding=false` needs no repeated
+`OPENAI4S_HOST` setting.
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 4b4987f, both ways: the launcher now rewrites what `url`/`status` print for the wildcard bind, and the paragraph (plus the Chinese counterpart and the troubleshooting row) now says the sidecar records the *bind* address and which half rewrites it, instead of promising there is no exception.

scripts/windows/bootstrap.sh
```diff
@@ -272,6 +367,7 @@ cli)
     echo "not installed: $APP" >&2
     exit 1
   fi
+  configure_fake_ip_dns
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 4b4987f. `bootstrap.sh cli` skips `configure_fake_ip_dns` for `status|url|stop|--help|-h|help`; anything not named still gets the probe, so a new subcommand fails safe.

openai4s/server/artifacts.py
Nobody-Zhang: Fixed in `32e32b7`. `_live_fingerprint` now opens with `O_NONBLOCK`, so a writer-less FIFO returns a valid fd immediately and the `S_ISREG` check rejects it as before. Regression test: `test_snapshot_survives_the_file_types_a_cell_can_leave_behind`, asserted with a join timeout because the failure mode is a hang.

openai4s/security/sandbox.py
Nobody-Zhang: Fixed in `32e32b7`, at the spawn sites rather than by restoring the flag: `tools/dynamic.py` and `kernel/preinstall.py` now pass `start_new_session=True` (the preinstall probe also gets `stdin=DEVNULL`), so the session boundary the argv assumes exists at every caller of `wrap_command`.

openai4s/webtools.py
Nobody-Zhang: Fixed in `32e32b7`. `_guard_url` percent-decodes the hostname before `_host_is_private`, so it inspects the host `Request._parse` will actually dial. Regression test: `test_the_guard_reads_the_host_the_client_will_actually_dial`.

openai4s/server/artifacts.py
Nobody-Zhang: Fixed in `32e32b7`. An open that fails now falls back to `_metadata_fingerprint` — the pre-PR `lstat` identity with `digest=None` — so an unreadable regular file keeps an identity in both snapshots instead of disappearing from the delta.

openai4s/egress.py
Nobody-Zhang: Fixed in `32e32b7`. `domain_allowed` keeps its own `if not domain_of(...): return True` before delegating, so the fail-open its docstring promises survives while `domain_in_allowlist` stays fail-closed. Regression test: `test_an_unparseable_target_stays_fail_open_for_domain_allowed`.

scripts/windows/openai4s.ps1
Nobody-Zhang: Fixed in `32e32b7` for the auto-open path (`Set-UrlHost` re-authorities the tokenized URL with `$ClientHost`), and completed in `4b4987f` for the `OpenAI4S.cmd url` / `status` passthrough, which was still printing the loopback rendering.

scripts/windows/openai4s.ps1
Nobody-Zhang: Fixed in `32e32b7`. `Get-WslDistros` splits on `\s{2,}` — the padded-column separator — so a name containing a space and a localized multi-word STATE are no longer conflated; single-space separation falls back to the first field only.

scripts/windows/bootstrap.sh
Nobody-Zhang: Fixed in `32e32b7`. A fresh install always claims the file (`|| [ "$FRESH_INSTALL" = "1" ]`), so the pristine bundle baseline carries the marker and a mirror set later can still apply. Regression test: `test_windows_launcher_claims_the_pristine_bundle_pip_conf_on_install`.

openai4s/server/artifacts.py
Nobody-Zhang: Fixed in `32e32b7`. The digest is now gated on `_metadata_ctime_can_lag()` (a cached WSL probe), with `OPENAI4S_ARTIFACT_CONTENT_FINGERPRINT` as an override in both directions. Regression test: `test_the_content_digest_is_paid_only_where_the_ctime_can_lag`.

scripts/windows/openai4s.ps1
Nobody-Zhang: Fixed in `32e32b7`. `Open-AppUrl` prints `$appUrl` under `OPENAI4S_NO_OPEN` and on the browser-failure path, instead of pointing the user back at a command whose answer it is already holding.

scripts/windows/bootstrap.sh
Nobody-Zhang: Fixed across `32e32b7` and `4b4987f`. The local resolver check now gates the `getent` probe, so an ordinary machine reads one file and stops; and `cli` skips `configure_fake_ip_dns` entirely for `status|url|stop|--help|-h|help`, which cannot egress.

openai4s/cli/main.py
Nobody-Zhang: Fixed in `32e32b7` via a shared `_live_endpoint(cfg)` used by `_daemon_request`, `cmd_url` and `cmd_status`; `4b4987f` adds the two `cmd_serve` "daemon already running at" banners, which were the last callers still rendering from config alone.

openai4s/security/sandbox.py
Nobody-Zhang: Fixed in `32e32b7`. `_run_self_test` passes `new_session=False`, so the probe argv matches the runtime's flag for flag. Regression test: `test_the_self_test_probes_the_argv_the_runtime_actually_launches`, which compares the probed flags against `sandbox.wrap_command`'s.

scripts/windows/bootstrap.sh
Nobody-Zhang: Fixed in `32e32b7`. The `.installed` marker is written last, after `configure_network` and `install_cli_link`, so an install that fails part-way stays retryable instead of being marked complete.

scripts/windows/openai4s.ps1
Nobody-Zhang: Fixed across `32e32b7` (one spelling, `$WslLogPath`) and `4b4987f`, which also addressed the two hazards this comment listed as unaddressed: the command drops `--exec sh -lc` for the bare `wsl -d "X" -- tail -40 ~/...` form (cmd.exe-pasteable), and `Assert-WslDataDir` now refuses whitespace.

PR #131, opened 2026-08-27
chore(deps): consolidate Ruff, mypy, and CodeQL patch updates

Nobody-Zhang:
## Summary

- carry #112: bump `ruff-pre-commit` from v0.16.3 to v0.16.4
- carry #113: bump `mypy` from 2.3.0 to 2.3.1
- carry #114: bump `github/codeql-action/upload-sarif` from v4.37.7 to v4.37.8
- preserve the three Dependabot changes as separate commits with their original authorship
- carry #134: bump the container base image from `python:3.12-slim-bookworm` to `3.14-slim-bookworm`

Supersedes #112, #113, #114, #134 and #135. #135 proposed the same CodeQL SHA this branch already carried, so it needed nothing beyond closing.

## Scope, after review

The bumps themselves are three lines. Everything else here is what a max-effort
review of them turned up, applied in place rather than deferred — the lock was
wrong in a way no gate could catch, and the reason it went unnoticed was a set
of gaps in the gates themselves.

**The lock (`c0bfeed`).** The branch was cut before the `singlecell` extra
landed, so Dependabot re-resolved a 54-package `uv.lock` against a `main` that
now carries 165. Two consequences, both invisible to CI:

- ~40 of the changed lock lines are marker normalization in packages the mypy
  bump never touches. Every one of them already exists byte-identically on
  `main`; they were stale re-derivation, and they were what made the file
  conflict.
- Two lines were **not** equivalent, and they hid inside that churn. Under
  `pandas 2.3.3`, `main` carries two numpy edges (`2.2.6 ; != '3.11.*'` and
  `2.4.6 ; == '3.11.*'`); this branch had one unmarked `2.2.6`. Taking the
  shorter side double-pins numpy on Python 3.11 — `uv export --extra
  singlecell --python 3.11` emits both versions — while `uv lock --check` and
  every required check stay green. The mirror case under `pandas 3.0.3` wrote
  a `< '3.12'` edge into a fork that is `>= '3.12'`-only.

Resolved by regenerating rather than hand-merging: take `main`'s lock at the
conflict, then `uv lock --upgrade-package mypy`. The lock diff against `main`
is now 101 lines, all inside the mypy `[[package]]` block, and `uv export
--extra singlecell` matches `main` except `mypy==2.3.0 → 2.3.1`.

**The gates that let it through (`e449d75`).**

- `tests/test_governance.py` matched only bare `uses:`, so `ci.yml`'s 42 pins
  in `- uses:` list style were invisible to it, and the `# vX.Y.Z` comment —
  the only readable part of a 40-hex pin — was optional. Both fixed, plus a
  glob-driven sweep over every workflow file; `ci.yml` and
  `publish-image.yml` were previously named by no test at all.
- `.pre-commit-config.yaml`: every `rev:` frozen to its commit SHA with the tag
  in a `# frozen:` comment. No version changed. These tools are not in
  `uv.lock`, `ruff-check` runs with `--fix`, and `setup.sh` installs the hooks
  for every contributor, so a re-pointed tag rewrites source on developer
  machines rather than failing a check.
- `.github/dependabot.yml`: `hook-versions` filtered on `update-types`, which
  the pre-commit ecosystem never emits — the group had batched nothing since it
  was added, which is why the hook PRs it was meant to stop kept arriving and
  kept being consolidated by hand. Switched to `exclude-patterns` for black.
  Gave `workflow-actions` the minor/patch filter its neighbours have and it
  lacked; a `setup-uv` major already went through as one row of a version table.
- `scorecard.yml`: the code-scanning upload sat behind a debugging artifact
  upload with no `if:`, so an artifact-storage hiccup skipped SARIF ingestion
  entirely. Reordered, with `always()` + `continue-on-error` on the artifact
  and `retention-days` 5 → 14 so two weekly SARIFs can coexist. Added the
  `timeout-minutes` this was the only job in the repo to lack, and the
  `concurrency` block the other three workflows have.

**The container base image (`25f5927`).** #134 could not have merged as opened:
`tests/test_container_deployment.py` spells the minor version out
(`python:3\.12-slim-bookworm@sha256:…`) and the Dependabot PR does not touch it.
The literal is kept rather than globbed — the `base-image` group is filtered to
minor/patch, and a Docker tag's "minor" is 3.12 → 3.14, a different CPython, so
the literal is what turns that into a red test somebody reads instead of a
digest swap that looks like a routine security rebuild.

Checked before taking it, because the runtime stage installs with
`--only-binary=:all: --require-hashes` and so has no sdist fallback: evaluating
the markers in `deploy/container-requirements-science.txt` for cp314-linux
selects 17 of its 27 lines and every one has a cp314 wheel — numpy 2.5.0,
scipy 1.18.0 and contourpy 1.3.3 get picked there, not the
`python_full_version < '3.11'` fork's 2.2.6 / 1.15.3 / 1.3.2, which have none
and would have broken the build. The digest matches what Docker Hub currently
resolves `3.14-slim-bookworm` to.

## Notes

- mypy 2.3.1 drops **every macOS x86_64 wheel** (six tags; thirteen cp315/cp315t
  added, nothing else regressed — `win_arm64` and `musllinux_1_2_x86_64` are
  intact). Faithful to upstream. Consequence: Intel Macs fall back to
  `mypy-2.3.1-py3-none-any.whl`, so the `always_run` `mypy-core` hook goes from
  ~0.9 s to ~3.2 s cold there. No CI job can see it — every macOS runner is
  `macos-14` (arm64) — while `docs/platforms.md` lists macOS (Intel) as
  `stable`.
- Both action SHAs were dereferenced against the upstream tags by hand:
  `ff2f1c62…` = v4.37.7, `db488dde…` = v4.37.8.
- ruff 0.16.3 and 0.16.4 produce byte-identical diagnostics on this tree with
  the hook's exact args, so the `[tool.ruff.lint] select` guard held.
- Two follow-ups this does **not** close are recorded in `TODO.md` /
  `TODO_zh.md` rather than here: a PR-time check that an action SHA matches its
  version comment (needs the network; this suite is offline), and
  cross-ecosystem Dependabot batching via `multi-ecosystem-groups`. A third is
  not a file edit at all: `Single-cell workflow (Python 3.11)` is a release gate
  but not a required check, so the interpreter both lock conflicts lived on is
  still only advisory at merge time.

## Validation

- `uv lock --check`
- `uv sync --locked --extra science --python 3.10`
- `uv run --locked mypy`
- `uv run --locked pre-commit run --all-files` — clean on a second pass, with
  the frozen SHAs cloning and installing normally
- `uv run --locked python scripts/check_directory_readmes.py`
- `uv run --locked pytest -n auto --maxprocesses=4 --dist loadfile` —
  **7805 passed, 42 skipped**
- the new pin sweep was checked for vacuity by stripping one version comment
  and confirming it fails — including against `fuzz.yml`, the workflow `main`
  added after the sweep was written, which the glob picked up with no edit
- the suite was also run on **CPython 3.14**, the interpreter the container now
  ships: **6 failed, 7855 passed**. Not a blocker for the image and not caused
  by this bump — all six are one CPython change (through a bare symlink, 3.13
  reports the symlink path in `sys.executable`, 3.14 reports the resolved
  binary) meeting one test fixture that builds `prefix/bin/python` as exactly
  such a symlink. A control run of the same tests on 3.13 passes. Recorded in
  `TODO.md` with the mechanism and what "done" looks like, because `ci.yml`'s
  matrix stops at 3.13 and `Container image builds and serves` boots the daemon
  without running tests.


Nobody-Zhang (commented):
Review completed: no blocking findings.

- `ruff-pre-commit` v0.16.4 exists upstream and the repository keeps the explicit E/F Ruff rule-set boundary.
- `github/codeql-action/upload-sarif` remains pinned to the full commit SHA for v4.37.8.
- `mypy` is the only package version changed in `uv.lock`; the additional marker rewrites are equivalent normalization and introduce no dependency or platform-resolution change.
- Local validation passed: lock check, sync, mypy, full pre-commit, and 7,382 offline tests (26 skipped).

Non-blocking residual risk: `scorecard.yml` runs on `main` push/schedule rather than pull requests, so the exact updated SARIF upload step gets its true workflow execution after merge.

Nobody-Zhang:
@cursoragent Review this PR using Grok 4.6 xhigh.

uv.lock
```diff
@@ -1103,10 +1110,10 @@ resolution-markers = [
     "python_full_version < '3.11'",
 ]
 dependencies = [
-    { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.11'" },
-    { name = "python-dateutil", marker = "python_full_version < '3.11'" },
-    { name = "pytz", marker = "python_full_version < '3.11'" },
-    { name = "tzdata", marker = "python_full_version < '3.11'" },
+    { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" } },
```
Nobody-Zhang: **Merge-conflict hunk 1 — this resolution double-pins numpy on Python 3.11.**

The PR is `mergeable=CONFLICTING`, and git's real merge yields exactly two conflicts, both in pandas blocks. This is the first.

On `main` this fork also spans 3.11 and carries **two** numpy edges:

```toml
{ name = "numpy", version = "2.2.6", marker = "python_full_version != '3.11.*'" },
{ name = "numpy", version = "2.4.6", marker = "python_full_version == '3.11.*'" },
```

Taking this branch's shorter, unmarked line deletes the `2.4.6 ; == '3.11.*'` edge while anndata/scanpy still require it in the same fork. Reproduced: `uv export --frozen --extra singlecell --python 3.11` then emits **both** `numpy==2.2.6 ; python_full_version == '3.11.*'` and `numpy==2.4.6 ; python_full_version == '3.11.*'`, and `uv sync --locked --extra singlecell` plans 133 installs instead of 132 — two numpy versions into one venv.

Nothing catches it: `uv lock --check` passes, and `Single-cell workflow (Python 3.11)` is not one of the four required contexts on `protect-main` (Lint, Offline tests py3.10, py3.12, Branch naming).

**Do not hand-resolve this file.** Merge `main`, take main's lock at the conflict, then regenerate:

```bash
uv lock --upgrade-package mypy
```

uv.lock
```diff
@@ -1178,10 +1185,10 @@ resolution-markers = [
     "python_full_version == '3.11.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "numpy", version = "2.4.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version == '3.11.*'" },
+    { name = "numpy", version = "2.4.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
```
Nobody-Zhang: **Merge-conflict hunk 2 — this writes a dependency edge that can never be selected.**

On `main`, pandas 3.0.3's `resolution-markers` all start at `python_full_version >= '3.12'` (scanpy pins `pandas<3` below that), so its numpy edge there is a single unmarked `numpy 2.5.0`. This branch re-adds `numpy 2.4.6 ; python_full_version < '3.12'` into that block.

It is inert at install time — the py3.11 export is unchanged — and `uv lock --check` and `uv sync --locked` both accept it. That is the problem: the lock records a state **no `uv lock` run would ever produce**, so the next relock (any Dependabot lock PR, `openai4s setup`, or a maintainer running `uv lock`) silently deletes these two lines and they resurface as unexplained churn in an unrelated PR's diff. It also destroys the only signal that the file was hand-resolved rather than regenerated.

uv.lock
```diff
@@ -743,7 +743,7 @@ wheels = [
 
 [[package]]
 name = "mypy"
-version = "2.3.0"
+version = "2.3.1"
```
Nobody-Zhang: **The safest-looking resolution of the conflict silently drops this bump.**

Resolving the `uv.lock` conflict by keeping main's file wholesale (`git checkout --ours uv.lock`) discards mypy 2.3.0 → 2.3.1 entirely, and no gate notices:

- `main` still pins mypy 2.3.0.
- `pyproject.toml` declares `mypy>=2.3,<3`, which 2.3.0 satisfies, so `uv sync --locked` stays valid.
- `uv run mypy` (ci.yml:113 and the always-run `mypy-core` hook) stays green.
- `grep -rn mypy .github scripts` finds only invocation sites — nothing anywhere names 2.3.1.

The PR would then merge under a title claiming a mypy update it did not apply, and Dependabot would close #113 as superseded. After regenerating, please confirm the merged lock actually says `version = "2.3.1"` in this block.

uv.lock
```diff
@@ -84,7 +84,7 @@ resolution-markers = [
     "python_full_version < '3.11'",
 ]
 dependencies = [
-    { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.11'" },
+    { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" } },
```
Nobody-Zhang: **These ~40 marker rewrites are already on `main`, verbatim.**

45% of the `chore(deps-dev): bump mypy` commit is marker normalization in packages the bump never touches — contourpy, matplotlib, pandas, scikit-learn, scipy, numpy consumers. Measured: `git show 335b69d -- uv.lock | grep -c '^[+-]'` = 181; the mypy block accounts for 99, the other 80 are these markers.

`main` landed every one of them in the single-cell relock, so relative to the merge target this half of the diff is a no-op — and it is exactly what makes the file conflict, and what the two genuinely-wrong pandas lines hide inside.

Regenerating on top of `main` collapses the whole lock diff to the mypy block:

```bash
git merge origin/main          # take main's uv.lock at the conflict
uv lock --upgrade-package mypy
```

Verified end to end: the result's diff against `main` is 101 lines, **all inside the mypy `[[package]]` block**, and `uv export --extra singlecell` is identical to main's except `mypy==2.3.0 → 2.3.1`.

uv.lock
```diff
@@ -753,52 +753,59 @@ dependencies = [
     { name = "tomli", marker = "python_full_version < '3.11'" },
     { name = "typing-extensions" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/12/af/4e516a05d3ca2eb9283e9ec45b2c02225c1514dd6da49fd3c9eaa6639370/mypy-2.3.0.tar.gz", hash = "sha256:465965d41cd9a2726694e983e8ce7113259327bec798115d1e1dfa2a52fb666e", size = 3988104, upload-time = "2026-07-13T11:34:53.387Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/82/6a/878cc1097d4035f82bd516658d0c528d2a9955bc7b363afcbd0b07fea11b/mypy-2.3.1.tar.gz", hash = "sha256:47c1b1207258513a9d93495f69c8be9de73916186f0e52703e8c461b7a623419", size = 3992554, upload-time = "2026-08-15T03:03:38.549Z" }
 wheels = [
```
Nobody-Zhang, line 1908: **Heads-up worth a line in the PR body: mypy 2.3.1 dropped every macOS x86_64 wheel.**

Diffed tag by tag against 2.3.0 — exactly six tags disappeared and all six are macOS x86_64 (cp310/cp311 `macosx_10_9`, cp312/cp313 `macosx_10_13`, cp314/cp314t `macosx_10_15`); thirteen were added, all cp315/cp315t. Nothing else regressed: `win_arm64` and `musllinux_1_2_x86_64` remain for every version, so the container image and the Windows zip are unaffected.

This is faithful to upstream (PyPI: 2.3.1 ships 8 macOS wheels, arm64 only, vs 2.3.0's 12) — no code change needed. But `docs/platforms.md:10` lists **macOS (Intel) as a `stable` tier**, and every macOS runner in ci.yml and release.yml is `macos-14` (arm64), so no CI job can surface it. Intel-Mac contributors now fall back to `mypy-2.3.1-py3-none-any.whl` on the `always_run: true` `mypy-core` hook: benchmarked on this repo's 8-file strict boundary, 0.91 s → 3.22 s cold (warm is unchanged).

.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
   # arrived and a bare `ruff` became ambiguous. The rule set is pinned in
   # pyproject.toml's [tool.ruff.lint]; see the comment there for why.
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.16.3'
+    rev: 'v0.16.4'
```
Nobody-Zhang: **`rev:` is a mutable tag, while every GitHub Action in this repo is SHA-pinned.**

ruff, black and isort appear nowhere in `uv.lock` (only `pre-commit` and `mypy` do), so pre-commit clones them from these tags at job time — and the `Lint (pre-commit)` job has no `~/.cache/pre-commit` restore step anywhere in `.github/`. If `v0.16.4` is ever re-pointed (a botched retag, a maintainer-account compromise), the next CI lint run and every contributor's next `git commit` execute the new code, and this hook runs with `--fix`, i.e. write access to the whole working tree — so the first symptom is rewritten source, not a red check. `setup.sh` runs `pre-commit install`, so the blast radius is not CI-only.

`tests/test_governance.py` would reject this trust model for a workflow action, and `.github/workflows/README.md` makes the same argument in the repo's own words ("a tag is mutable").

Not for this PR to fix — `pre-commit autoupdate --freeze` rewrites all four `rev:` lines and changes what Dependabot writes from then on, so it is a pinning-policy call. Flagging it because this PR moves one of those tags. It would render as `rev: <40-char sha>  # frozen: v0.16.4`, and Dependabot keeps maintaining it.

.github/workflows/scorecard.yml
```diff
@@ -33,6 +33,6 @@ jobs:
           path: results.sarif
           retention-days: 5
       - name: Upload to code scanning
-        uses: github/codeql-action/upload-sarif@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
+        uses: github/codeql-action/upload-sarif@db488ddef3bf6cb639b32c2e9a7c0a7ea8271d28 # v4.37.8
```
Nobody-Zhang: **Nothing validates this line before merge — not the trigger, not the test.**

`scorecard.yml` fires only on `push: branches: [main]` and the Saturday cron, so the step never executes for a PR. The one PR-side consumer is `tests/test_governance.py`, whose whole assertion is:

```python
PINNED_ACTION = re.compile(r"^\s*uses:\s*[^@\s]+@[0-9a-f]{40}(?:\s+#.*)?$")
```

Any 40 hex characters pass, and the `# v4.37.8` comment is optional and never cross-checked against the SHA. So a transposed digit, a SHA from a fork, or a stale comment merges fully green and first surfaces on a post-merge run nobody watches — with the visible symptom being SARIF that stops reaching code scanning, not a blocked merge. `grep -rn 'zizmor|actionlint|pinact'` over `.github/`, `.pre-commit-config.yaml` and `scripts/` returns nothing.

For the record, I dereferenced both pins by hand and they are correct: `ff2f1c62…` = v4.37.7, `db488dde…` = v4.37.8.

The PR body notes this as a "residual". Per `CONTRIBUTING.md:28` — *"A deferred follow-up goes in `TODO.md`, not in a comment… A loose end recorded only in a code comment or a merged PR description is a loose end nobody will find again."* — that belongs in `TODO.md` with a *Done when:* line.

Nobody-Zhang:
## Max-effort review — remaining findings (outside the diff)

Seven findings are posted inline. These eight cannot be anchored to a diff line, so they are collected here. Ranked most severe first.

**Verified clean first, so it is on the record:** both action SHAs dereference to the real upstream tags (`ff2f1c62…` = v4.37.7, `db488dde…` = v4.37.8); ruff 0.16.3 and 0.16.4 produce **byte-identical** diagnostics on this tree with the hook's exact args; mypy 2.3.1 passes the strict boundary unchanged; and the ~40 marker rewrites are genuinely equivalent — proven with a PEP 508 evaluator over 18–24 environments and with `uv export`, not by reading. The PR body's equivalence claim is correct.

---

### 1. `uv.lock` — a file-level conflict shortcut drops the `singlecell` extra (integration)

This branch's lock is a whole-file relock of a `pyproject.toml` that predates the `singlecell` extra: **54 packages vs main's 165**, `provides-extras = ["science", "chemistry"]`. During `git rebase main` the ours/theirs labels invert relative to a merge — `--theirs` is the replayed PR commit — so the reflexive `git checkout --theirs uv.lock` installs the 1970-line lock over main's 4650-line one.

Verified: `uv lock --check` against main's `pyproject.toml` then fails hard —

```
No solution found ... Because there is no version of pertpy[de]==1.0.3 and
openai4s[singlecell] depends on pertpy[de]==1.0.3
```

— killing all 15 `uv sync --locked` steps in `ci.yml` plus `release.yml:177/252`. Worth stating precisely because this failure is *loud*, which invites the assumption that the quiet hunk-1 case is equally protected. It is not.

### 2. `.github/workflows/ci.yml:247` — no Python 3.11 anywhere in this PR's signal (test-coverage)

`python-version: ["3.10", "3.12", "3.13"]`, and `grep -c singlecell` on this branch's `ci.yml` is 0 — yet **both** conflict hunks live exclusively in the 3.11 fork of the resolution graph. Run 33033022823 finished 28 SUCCESS / 3 SKIPPED four hours before #121 merged the `singlecell-skill` job to main; `Single-cell workflow (Python 3.11)` appears in none of the 31 status-check entries here.

Compounding it: the `protect-main` ruleset requires only **Lint (pre-commit), Offline tests (py3.10), Offline tests (py3.12), Branch naming policy** — none of which is 3.11 or pulls scanpy — and `grep -rn 'uv.lock' tests/ scripts/` returns nothing, so there is no repo-side lock-consistency assertion either. Rebasing gives the singlecell job a chance to run, but it stays advisory.

### 3. `.github/workflows/scorecard.yml:29` — a debug artifact upload gates the code-scanning upload (ci-correctness)

`Upload to code scanning` (the step this PR edits) sits *behind* `Upload SARIF artifact`, with no `if:` on either, so GitHub's implicit `if: success()` makes SARIF ingestion conditional on artifact storage. A transient `upload-artifact` 5xx or an artifact-quota hit skips the codeql upload: the job goes red for a storage reason, no SARIF reaches code scanning that week, and the Security tab keeps stale Scorecard alerts with no indication. Symmetrically, when `Run Scorecard` fails, *both* uploads are skipped — so the artifact whose whole purpose is post-mortem debugging is never produced on exactly the runs that need it.

The repo has a dedicated test for this anti-pattern (`tests/test_ci_gate_independence.py`), hard-scoped to `ci.yml`. Fix: `continue-on-error: true` + `if: always()` on the artifact step, or swap the two steps.

### 4. `.github/workflows/scorecard.yml:12` — the only job in the repo with no `timeout-minutes` (ci-correctness)

Counted mechanically: `ci.yml` declares it 22 times, `release.yml` 12, `publish-image.yml` 1, **`scorecard.yml` 0** — so `analysis` falls back to GitHub's 360-minute default. `ossf/scorecard-action` paginates commits for the Maintained/Contributors checks and `publish_results: true` posts to a third-party host; a hang burns six hours instead of the ~15 every peer job is capped at, and since this workflow also runs on `push: branches: [main]`, the next PR's merge commit carries a six-hour pending check.

### 5. `.github/workflows/scorecard.yml:3` — the only workflow with no `concurrency:` block (ci-correctness)

`ci.yml`, `release.yml` and `publish-image.yml` all have one. Because scorecard triggers on every push to `main` as well as weekly, two `analysis` jobs can run at once against the same code-scanning analysis key — and code scanning treats each `upload-sarif` for a key as the *complete* result set, so whichever finishes second wins, possibly the one that started against the older commit. Alerts close and reopen nondeterministically and `publish_results: true` double-posts to the OpenSSF API. Suggested: `concurrency: { group: scorecard-${{ github.ref }}, cancel-in-progress: true }`, matching `ci.yml`.

### 6. `tests/test_governance.py:12` — the pin gate validates shape, not identity (gate-coverage)

```python
PINNED_ACTION = re.compile(r"^\s*uses:\s*[^@\s]+@[0-9a-f]{40}(?:\s+#.*)?$")
```

The `# vX.Y.Z` comment is optional and never cross-checked against the SHA, and `@pytest.mark.parametrize("name", ["scorecard.yml"])` plus the release-workflow test cover two of four workflow files — **`ci.yml` (40 `uses:` lines) and `publish-image.yml` are pinned by convention only.** Dependabot's `workflow-actions` group rewrites `uses:` in all four. A grouped bump landing a mutable tag in `ci.yml` — the file every contributor's code passes through, including fork PRs — passes every gate and reads exactly like this PR.

### 7. `.github/dependabot.yml:28` — the `hook-versions` group is inert (altitude)

`update-types: ["minor", "patch"]` can never match: Dependabot resolves pre-commit hooks as git refs and does not semver-classify them. Proof is in this PR's own trailers — ruff commit `ac5b216` carries `dependency-type: direct:production` with **no** `update-type:` and **no** `dependency-group:`, while the same-week `335b69d` (uv) and `1cd9209` (github-actions) carry both. The group has batched zero updates since it was added; the problem its comment claims to have solved is the one this PR is hand-fixing, for at least the fourth time (#75, #97, now #131).

Right depth: drop the `update-types` filter and keep black out with `exclude-patterns`, a filter this ecosystem *can* evaluate. And the cross-ecosystem batching this PR is actually doing by hand needs the top-level **`multi-ecosystem-groups`** key plus `multi-ecosystem-group:` on each `updates` entry — `groups:` is per-ecosystem by construction and could never have batched #112/#113/#114.

### 8. `.github/dependabot.yml:38` — `workflow-actions` batches majors, unlike its neighbours (altitude)

`patterns: ["*"]` with no `update-types` filter, while the pre-commit and uv groups both restrict to minor/patch — the pre-commit one with a written rationale that a major *"has to be read on its own, not merged as one row of a version table"*. It already happened: `69468da`, titled *"bump the workflow-actions group with 2 updates"*, batched `astral-sh/setup-uv` **9.0.0 → 10.0.1** (`update-type: version-update:semver-major`) with a codeql patch. Majors land in `release.yml` (`id-token: write`, OIDC PyPI publish) and `scorecard.yml` (`security-events: write`) — the two workflows with the largest blast radius, and the two this policy protects least. Unlike pre-commit, github-actions *is* semver-classified, so the same filter would work here.

---

### Suggested remedy for the lock (verified end to end)

```bash
git merge origin/main
git checkout --theirs -- uv.lock && git add uv.lock   # take main's lock, do not hand-merge
uv lock --upgrade-package mypy
```

Result, measured: the `uv.lock` diff against `main` collapses to **101 lines, all inside the mypy `[[package]]` block**; `uv export --extra singlecell` is identical to main's except `mypy==2.3.0 → 2.3.1`; `uv lock --check`, `uv sync --locked --extra science --python 3.10` and `uv run --locked mypy` all pass.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #132, opened 2026-08-27
fix(security): address Security & quality alerts

Nobody-Zhang:
## Summary

Addresses the 23 open GitHub Security & quality findings at the code and workflow layer.

- Fixes the 17 code-actionable CodeQL/Scorecard findings expected to clear after merge and rescanning.
- Adds an exact-path CodeQL fixture exclusion for alert 23; default setup must select `.github/codeql-config.yml` after merge.
- Documents the five findings that require repository governance or external state rather than a code change: CodeQL 133 and Scorecard 67, 89, 91, and 92.
- Adds defense-in-depth for an additional stored-XSS path found during review: untrusted Artifact previews now use an inactive, sandboxed CSP and an empty iframe sandbox.

Key changes:

- Freeze release jobs to the triggering commit, validate annotated tags repeatedly, disable unsafe caches, and pin the PyPI publisher action.
- Pin Docker base images by digest and install build/science requirements from hash-locked exports.
- Lock Playwright with npm and use `npm ci`; add Dependabot coverage.
- Add a bounded Atheris protocol fuzzer and CI workflow.
- Remove inline UI bootstrap JavaScript and harden response CSP.
- Canonicalize and bound retrosynthesis SVG/HTML embedding.
- Avoid secret-shaped URL handling and clear-text diagnostic output.

## Area

- [x] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [x] Security-related change
- [x] Release / packaging

## Verification

Ran:

```text
UV_CACHE_DIR=/tmp/openai4s-uv-cache uv run pytest -n auto --maxprocesses=4 --dist loadfile
# 7714 passed, 32 skipped before the clean, conflict-free rebase onto current main

UV_CACHE_DIR=/tmp/openai4s-uv-cache uv run pytest \
  tests/test_security_headers.py tests/test_gateway.py \
  tests/test_response_contract_downloads.py tests/test_response_contract_coverage.py \
  tests/test_volcengine_connector.py tests/test_team_isolation_hardening.py -q
# passed after rebase onto current main

UV_CACHE_DIR=/tmp/openai4s-uv-cache uv run pre-commit run --all-files
# all hooks passed after rebase

UV_CACHE_DIR=/tmp/openai4s-uv-cache uv run python scripts/capture_response_schemas.py --check
# 193/193 routes; no breaking response-shape change

UV_CACHE_DIR=/tmp/openai4s-uv-cache uv run python scripts/capture_response_contract.py --check
# contract up to date: 193/193 routes

npm ci
npm audit
# 0 vulnerabilities

node --test tests/node_dependency_contract.test.mjs
node tests/browser_stage0_acceptance.mjs
node tools/skills-installer/check_package.mjs

uv run python3 -m harness.cli run --tier pr --offline
# 38/38 passed

uv build
uv run python scripts/verify_release_artifacts.py dist
```

Not run:

```text
full live-browser smoke
```

Reason:

```text
Docker and a runnable Chromium installation were unavailable on this macOS host.
Static UI contracts, focused gateway/CSP tests, and the complete offline suite passed.
```

## Risk and Reviewer Focus

Please inspect:

1. `.github/workflows/release.yml`: immutable commit/tag binding and the remaining external requirement to protect `v*` tags.
2. `openai4s/server/security_headers.py` and Artifact delivery in `gateway.py`: preview CSP must stay inactive while remaining same-origin embeddable.
3. `Dockerfile` plus generated hash-locked requirement files.
4. `skills/retrosynthesis_planning/kernel.py`: canonical SVG data URI and HTML/JavaScript boundary encoding.
5. `.github/codeql-config.yml`: the ignore is intentionally one exact, non-executable fixture path.

The dashboard will not reach zero from this PR alone. CodeQL 133 should be dismissed as a synthetic-data false positive; Scorecard 67/89 require review governance; Scorecard 91 is age/time based; Scorecard 92 requires external CII registration.

### Diff Readiness

- [x] This PR is single-purpose.
- [x] Hotspot files were edited surgically if touched: `gateway.py`, `host_dispatch.py`, `store.py`, `app.js`, `worker.py`, `manager.py`.

### Tests

- [x] Kernel / host-RPC / gateway changes include focused contract coverage or a documented smoke test.

### Core Dependency Policy

- [x] Optional science imports are guarded by `try/except ImportError` at every in-tree use site.

- [x] Security-sensitive changes include appropriate synthetic tests or manual verification notes.

### Docs

- [x] `README.md` and `README_zh.md` stay in sync where translated sections are affected.


Nobody-Zhang:
@cursoragent Review this PR.

openai4s/server/security_headers.py
```diff
@@ -6,66 +6,58 @@
...
+            "style-src 'unsafe-inline'",
+            "img-src data: blob:",
+            "font-src data:",
+            "media-src data: blob:",
+            "connect-src 'none'",
+            "object-src 'none'",
+            "base-uri 'none'",
+            "form-action 'none'",
+            "frame-ancestors 'self'",
+            "sandbox",
```
Nobody-Zhang: **`sandbox` puts the artifact on an opaque origin, so its own figures stop loading.**

Verified in Chromium against a probe server, three variants of this exact policy:

| policy | `GET /pic.png`, `GET /s.css` |
| --- | --- |
| this PR verbatim | blocked — *"violates … `img-src data: blob:`"* |
| `img-src 'self' data: blob:` + `sandbox` | still blocked (`ERR_BLOCKED_BY_CLIENT`) |
| `img-src 'self' data: blob:` + `sandbox allow-same-origin` | **200** |

So dropping `'self'` from `img-src`/`style-src` is not the cause and restoring it is not the fix — `sandbox` is, because `'self'` cannot match an opaque origin.

The regression is on the top-level **View** path, which is not sandboxed today: an agent writes `report.html` + `figure.png` (the standard Code-as-Action pair), the user clicks View (`app.js:9586`, `target="_blank"`), and before this PR `_serve_artifact` carried `img-src 'self' data: blob:` with no sandbox, so the figure loaded. `store.artifact_by_unique_filename` exists precisely so `` resolves, so this is a first-party pattern.

`sandbox allow-same-origin` keeps the document script-free (`script-src 'none'` and no `allow-scripts`) while restoring sibling sub-resources. Left unapplied: the docstring states the opaque origin as the intent, so the trade is yours.

openai4s/server/security_headers.py
```diff
@@ -6,66 +6,58 @@
...
-        return []
-    out = []
-    for body in _INLINE_SCRIPT_RE.findall(raw):
-        digest = base64.b64encode(hashlib.sha256(body).digest()).decode("ascii")
-        out.append(f"'sha256-{digest}'")
-    return out
+    return "; ".join(
+        [
+            "default-src 'none'",
+            "script-src 'none'",
```
Nobody-Zhang, line 50: **The A/OR graph script this PR hardens cannot run in either in-product view.**

`script-src 'none'` + bare `sandbox` here, plus `frame.setAttribute("sandbox", "")` in `app.js:8662`. Verified in Chromium on a **top-level navigation** (not just the iframe):

> Blocked script execution in '…' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.

So `/preview/<id>` and `/api/v1/artifacts/<id>` both render the panel chrome — empty `<svg id="andor-svg">`, the permanent *"Select a molecule or reaction node."* — and never draw the graph. Same for `skills/admet_genetic/examples/optimization_dashboard.html`.

To be fair to the change: this is **not** a regression. Inline script was already unauthorized under the pre-PR hash policy (only `index.html`'s bootstrap was hashed). But the PR spends 124 lines hardening `_ANDOR_TREE_SCRIPT`/`safeStructureSource` — including a JS allowlist that duplicates the Python one — for a script that is dead code wherever the product displays it, while `SKILL.md` still advertises an "interactive retrosynthesis knowledge graph". No browser test covers the `html-preview` renderer, so nothing in the tree states the trade either way.

Worth deciding explicitly: either the artifact viewer gets `sandbox allow-scripts` on a separate origin, or the skill stops shipping an interactive panel and the JS half goes.

fuzz/protocol_fuzzer.py
```diff
@@ -0,0 +1,47 @@
...
+            max_len=1 << 20,
+        )
+        return
+
+    try:
+        if selector == 3:
+            decode_control(payload)
+        else:
+            decode_data(payload)
+    except ProtocolError:
```
Nobody-Zhang, line 44: **`decode_control` leaks `RecursionError`, so this new PR-blocking gate will report a crash.**

```
>>> decode_control(b"[" * 20000)
RecursionError: maximum recursion depth exceeded while decoding a JSON array
```

~1 KB is already enough at CPython's default limit — far under `MAX_CONTROL_JSON` (65536) and the job's `-max_len=1048576` — and libFuzzer mutates its way to a run of `[` almost immediately. `RecursionError` subclasses `RuntimeError`, so `protocol.py`'s `except (UnicodeDecodeError, ValueError)` does not convert it and this `except ProtocolError` does not catch it.

It is a production bug too, not just a fuzzer artifact: `share/tunnel.py:213` and `share/relay.py:399` both catch only `p.ProtocolError`, so an untrusted relay peer kills the reader thread with a ~1 KB frame instead of the documented drop-the-frame path.

Fixed at the source rather than by widening the target's `except`, with a regression test in `tests/test_share_protocol.py`:

```python
    except RecursionError as error:
        raise ProtocolError("control frame nests too deeply") from error
```

openai4s/server/gateway.py
```diff
@@ -12425,7 +12433,8 @@ def _send(
                 self.send_header("X-Request-Id", _sanitize_header_value(request_id))
             # Applied here rather than at the HTML route so no response can be
             # added later that quietly opts out.
-            for k, v in security_headers(WEBUI_DIR / "index.html").items():
+            hardened = security or security_headers(WEBUI_DIR / "index.html")
```
Nobody-Zhang: **Truthiness, not `is None` — an empty profile silently becomes the permissive one.**

```python
hardened = security or security_headers(WEBUI_DIR / "index.html")
```

A caller that builds the profile conditionally (`security=artifact_headers if artifact else {}`) passes `{}`. One layer up, `response_capture.observing_send` tests `if security is None`, so it classifies `{}` as a real profile and forwards it; `_send` classifies the same value as "no profile" and emits `script-src 'self' 'wasm-unsafe-eval'`, `connect-src 'self'` and `X-Frame-Options: DENY` for bytes the caller intended to serve inactive. The two layers disagree about the same value, and the failure direction here is toward the *less* restrictive policy, with no error anywhere.

Also: `security`'s values are the only ones in `_send` that bypass `_sanitize_header_value`, which the `extra` loop two lines down applies to a parameter with identical trust properties.

Fixed both, plus a regression test that fails on the `or` form.

openai4s/server/response_capture.py
```diff
@@ -488,7 +488,13 @@ def observing_send(code, body, ctype, extra=None):
                     )
                 except Exception:  # noqa: BLE001 - never break a response
                     pass
-                return raw(code, body, ctype, extra)
+                if security is None:
```
Nobody-Zhang, line 491: **The compatibility branch guards the case that did not need guarding — and the third writer was missed.**

`_serve_file` now *always* calls `self._send(..., extra=extra, security=security)`, so a four-parameter stub breaks even on the `security=None` path, with or without the capture installed. Reproduced by driving `GET /artifacts/<id>` through `tests/test_artifact_workbench.py`'s handler helper:

```
gateway.py:13530: TypeError: <lambda>() got an unexpected keyword argument 'security'
```

The `if security is None` branch cannot rescue the artifact route either, because `_serve_artifact` always passes a non-`None` profile — it only keeps the other seven `_send` call sites inside `_api` working. Twenty-plus stubs across `tests/` still had the four-parameter shape; this PR updated two.

The same gap in `_probe_handler`'s **`_stub_stream`** is not latent at all once the bundle route selects a profile: the `TypeError` is swallowed by `drive_all_routes`' bare `except Exception`, and both `artifacts.zip` routes regress to `kinds: ["json"], statuses: [404]` in the captured contract — verbatim the regression `_stub_stream`'s own comment says it exists to prevent.

Fixed: `_stub_stream` grew the parameter, and every four-argument `_send`/`_stream_file` stub in `tests/` now accepts `security=None`.

skills/retrosynthesis_planning/kernel.py
```diff
@@ -2228,14 +2233,11 @@ def _render_molecule_briefs_panel(
...
+            build_molecule_structure_src(str(brief["smiles"]))
+        )
         alt = f'alt="Structure of {html.escape(str(brief["smiles"]))}"'
-        img = f'<img class="mol-structure" src="{html.escape(structure_sources["primary"])}" '
-        if structure_sources["fallback"]:
-            img += (
-                f'data-fallback-src="{html.escape(structure_sources["fallback"])}" '
-                "onerror=\"this.onerror=null;this.src=this.dataset.fallbackSrc;this.classList.add('structure-fallback');\" "
-            )
+        img = f'<img class="mol-structure" src="{html.escape(structure_src)}" '
```
Nobody-Zhang: **The 1 MiB cap is production-reachable and fails closed into `src=""`.**

With RDKit 2026.03.5 installed, a 2,162-character peptide SMILES (`"N" + "C(=O)C(N)Cc1ccccc1"*120 + "O"`) renders to a **1,104,898-character** data URI — I ran it. The RDKit SVG grows about 660 bytes per heavy atom, so roughly 1,570 atoms crosses the cap; that is an ordinary polymer/peptide target, not an adversarial input.

`_canonical_structure_src` then returns `""`, and this line emits `Structure of …`. An empty `src` selects no source, so the card shows the browser's broken-image glyph with nothing to explain it. Both sibling consumers guard the same value — `_svg_node` via `has_structure = bool(...)`, the graph script via `if (node.structureSrc)` — and drop the element instead.

`build_molecule_fallback_structure_src` only runs when RDKit cannot *parse* the molecule, never when the depiction is merely oversized, so the labelled placeholder the removed `onerror` path used to provide is now unreachable on this path.

Fixed with one helper used by all three call sites — canonical primary, else the labelled placeholder — plus a test that asserts the cap is still reachable and that the placeholder comes back.

openai4s/server/response_capture.py
```diff
@@ -657,8 +663,10 @@ def _probe_handler(
         public_failure(value, code, _CAPTURE_REQUEST_ID),
         route=route,
     )
-    handler._send = lambda code, body, ctype, extra=None: recorder.observe_raw(
-        method, path, code, ctype, len(body or b""), route=route
+    handler._send = (
+        lambda code, body, ctype, extra=None, security=None: recorder.observe_raw(
```
Nobody-Zhang, line 667: **The frozen-response gates cannot see any of this hardening.**

Both capture writers accept `security` only to drop it, and `Recorder.observe_raw(method, path, code, content_type, length, route)` has no header parameter at all. `docs/response-contract.json` and `docs/response-schemas.json` contain zero occurrences of `content-security-policy`, `x-frame-options` or `nosniff`.

Tested it: removing `security=artifact_headers` from **both** `_serve_artifact` call sites and re-running `scripts/capture_response_contract.py --check` produces byte-identical output. The mutation is invisible to the gate, so a later edit to the artifact routes can silently revert the whole change with both contract gates green.

Only the pytest suite gates it. I added `tests/test_gateway.py` coverage for the writer, the empty-profile case and the bundle route (each verified to fail when its fix is reverted), which is the honest short-term answer — but if the header profile is now part of what a route *answers*, `observe_raw` recording it is the deeper one.

.github/codeql-config.yml
```diff
@@ -0,0 +1,9 @@
+name: OpenAI4S CodeQL configuration
+
+# This is one exact, non-executable upstream capture rather than a blanket tree
+# exclusion. It is a byte-exact arXiv page read only by parser tests. Executable
+# bioSkills examples remain scanned even when an individual alert is a false
+# positive; those alerts must be dismissed individually so a future re-pin does
+# not inherit an invisible path.
+paths-ignore:
+  - tests/fixtures/arxiv_abs_2503.06687.html
```
Nobody-Zhang: **This exclusion is inert today, and the test that "gates" it measures only the file's own bytes.**

```
$ gh api repos/PKU-YuanGroup/OpenAI4S/code-scanning/default-setup
{"state":"configured", …}
$ gh api repos/PKU-YuanGroup/OpenAI4S/properties/values
[]
```

There is no CodeQL workflow in `.github/workflows/`, so the repo is on **default setup**, where a repo config file is honored only when the repository property `github-codeql-config-file` points at it — and the repo has no custom properties set. `.github/CONTENTS.md` states the requirement in prose, but nothing enforces it, so the alert this file is meant to close stays open while the suite reports success.

`test_codeql_ignores_only_the_two_exact_upstream_captures` reads the YAML and compares `paths-ignore` to a one-element list; it is green either way. (Its name also says "two … captures" while both the list and the file's own comment say one — renamed.)

The two honest depths are the one this file's own comment prescribes for the sibling case (dismiss the individual alert in the UI) or an advanced `.github/workflows/codeql.yml` with `config-file:`, which is in-repo, reviewable and self-enforcing. Not applied — setting the repository property is outside the tree.

openai4s/cli/main.py
```diff
@@ -918,6 +918,9 @@ def cmd_init(args) -> int:
 
     payload = result.as_dict()
     if args.json:
+        # OnboardingResult.as_dict() is an explicit secret-free projection:
+        # has_api_key is a boolean and the credential value never reaches it.
+        # codeql[py/clear-text-logging-sensitive-data]
```
Nobody-Zhang: **A CodeQL suppression scope is the comment's own line, so this suppresses nothing.**

`py/clear-text-logging-sensitive-data` is reported at the `print(json.dumps(payload, …))` statement on the next line. The standard `AlertSuppression.ql` `SuppressionScope` spans column 1 to the end of the **comment's** line, so a comment placed above the call does not cover it — the alert stays open while the hunk reads as though it closes it.

The claim in the comment is correct and is now backed by `test_cmd_init_json_stdout_is_a_secret_free_projection`; only the placement is wrong. Fixed by giving the sink its own short line so the marker rides it without black reflowing it away:

```python
        rendered = json.dumps(payload, ensure_ascii=False, indent=2)
        print(rendered)  # codeql[py/clear-text-logging-sensitive-data]
```

fuzz/protocol_fuzzer.py
```diff
@@ -0,0 +1,47 @@
+"""Coverage-guided fuzz target for the two untrusted wire decoders."""
+
+from __future__ import annotations
+
+import io
+import sys
+
+import atheris
+
+with atheris.instrument_imports():
```
Nobody-Zhang: **This instruments the whole daemon to fuzz two decoders.**

`openai4s/server/__init__.py` eagerly imports `gateway`, so `import openai4s.server.ws_frames` inside this block loads **246 `openai4s` modules** (441 total) — the store, the kernel manager, the LLM client, the tool registry, the skills loader — and atheris instruments all of them.

That inflates start-up and RSS inside the job's 15-minute budget, and dilutes libFuzzer's coverage map with counters no input can ever reach, which degrades the coverage guidance the gate exists for. Importing the two modules outside the instrumentation block, or naming them in `instrument_imports(include=[...])`, is what the docstring ("the two untrusted wire decoders") already claims.

Not applied: atheris is Linux/x86_64-only, so I cannot validate the `include=` behaviour from here, and an unvalidated edit to a CI gate is worse than the waste.

openai4s/server/security_headers.py
```diff
@@ -6,66 +6,58 @@
...
 
 def content_security_policy(index_html: Path) -> str:
-    key = _cache_key(index_html)
-    cached = _cache.get(key)
-    if cached is not None:
-        return cached
-
-    script_src = ["'self'", *_inline_script_hashes(index_html)]
+    """Return the static policy; ``index_html`` remains for API compatibility."""
+    _ = index_html
```
Nobody-Zhang: **`index_html` is now dead in all three functions, but still built per response.**

`content_security_policy` does `_ = index_html`; `security_headers` passes it only there; `artifact_security_headers` passes it only to `security_headers`. Meanwhile `gateway.py` constructs `WEBUI_DIR / "index.html"` on **every** response (12436) and again per artifact request (13569), to feed a parameter nothing reads.

It also costs the tests their meaning: eight tests in `tests/test_security_headers.py` take the `index_html` fixture only to feed a discarded argument, and `test_policy_never_parses_html_to_authorize_scripts` / `test_missing_index_html_still_yields_a_usable_policy` write adversarial and missing HTML files that nothing reads — they now assert only that a constant-returning function returns a constant.

The retained signature tells every future reader that the policy still follows the file, which the new module docstring says it deliberately no longer does. Not applied, since "remains for API compatibility" reads as a deliberate call — but with three in-tree call sites and no external consumer, dropping it is cheap.

.github/workflows/release.yml
```diff
@@ -616,23 +619,41 @@ jobs:
       - name: Check out the pipeline at the frozen release SHA
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          ref: ${{ needs.freeze.outputs.sha }}
+          ref: ${{ github.sha }}
           fetch-depth: 0
           persist-credentials: false
           path: source
+      - name: Revalidate the remote annotated tag before staging
```
Nobody-Zhang, line 626: **Three byte-identical copies, and the new test pins them rather than reconciling them.**

Lines 626-643, 736-753 and 791-808 are character-for-character the same fetch / `cat-file -t` / `rev-parse` / compare sequence, including the `openai4s-release-candidate` ref name and both `::error::` strings; only the step `name:` differs. The `pypi` job gained a whole `actions/checkout` and `contents: read` purely to host its copy.

`test_every_publication_boundary_revalidates_the_remote_annotated_tag` loops over a hardcoded `("attach", "pypi", "finalize")` asserting four substrings of each copy — so it mirrors the copy-paste instead of removing the need for it, and a fourth publication boundary (the ghcr push, a homebrew/winget job) is invisible to it. Any hardening — a different local ref name, a signature check, a quoting fix — has to land in three places or the boundaries disagree silently.

A `scripts/revalidate_release_tag.sh` (the repo already keeps release tooling in `scripts/`) or a composite action under `.github/actions/` would leave one copy and let the test assert that each boundary *calls* it. Not applied: extracting it means rewriting the gate test in the same change, which is your call rather than a reviewer's.

openai4s/server/gateway.py
```diff
@@ -13506,13 +13515,19 @@ def _serve_static(self, path: str) -> bool:
...
+            extra: dict | None = None,
+            security: dict[str, str] | None = None,
+        ) -> None:
             try:
                 body = path.read_bytes()
             except OSError:
                 self._json({"error": "not found"}, 404)
                 return
-            self._send(200, body, ctype)
+            self._send(200, body, ctype, extra=extra, security=security)
```
Nobody-Zhang, line 13551: **`_stream_file` is the other artifact-bytes writer and did not get `security`.**

`GET /api/v1/frames/{fid}/artifacts.zip` and `GET /api/v1/projects/{pid}/artifacts.zip` reach `_serve_artifact_bundle` → `_stream_file`, which hardcodes the UI-shell profile — `script-src 'self' 'wasm-unsafe-eval'`, `X-Frame-Options: DENY` — while `GET /api/v1/artifacts/{id}` for the same bytes now returns `script-src 'none'; sandbox`. Two writers of one fact, allowed to disagree, in the function whose own comment says it "streams artifact bytes … the one that most needs nosniff and a closed CSP".

`_serve_file` was given both new parameters in this PR; its sibling was skipped. Fixed: `_stream_file` takes the same `security` profile, `_serve_artifact_bundle` passes it, and `tests/test_gateway.py` now gates both the writer and the route (both fail if the fix is reverted).

Note this also required `response_capture._probe_handler`'s `_stub_stream` to grow the parameter — see the comment on `response_capture.py`.

package.json
```diff
@@ -35,5 +35,8 @@
   "scripts": {
     "test": "node tools/skills-installer/selftest.mjs"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "devDependencies": {
+    "playwright": "1.62.1"
```
Nobody-Zhang, line 40: **The new devDependency needs Node 20; this floor says 18, and `npm ci` only warns.**

`package-lock.json` records `"node_modules/playwright": {"engines": {"node": ">=20"}}`. A contributor on Node 18 follows the newly rewritten instruction in `CLAUDE.md` (`npm ci --ignore-scripts && ./node_modules/.bin/playwright install <engine>`): npm prints `EBADENGINE` and installs anyway — it does not fail without `engine-strict` — and then the driver aborts on the Node version. The previous `npm install --no-save playwright@1.54.1` was compatible with the declared floor.

CI is unaffected (all three browser jobs pin `node-version: "22"`), so this only bites locally, which is exactly where the instruction is aimed.

Left `engines.node` at `>=18`, since that floor belongs to the published Skill installer rather than the browser gates, and documented the Node 20 requirement in `CLAUDE.md` and `tests/README*.md` instead. Raising `engines` is the other reasonable call — your choice.

Nobody-Zhang:
**Finding 8 of 15 — `.github/workflows/publish-image.yml:65`.** Posted here rather than inline: that file is not part of this diff, so GitHub will not accept a line comment on it.

**The ghcr publication path still has the exact defect this PR closes in `release.yml`.**

`ref: ${{ steps.ref.outputs.tag }}` resolves `v0.2.0` at job-run time, minutes to tens of minutes after the `release: published` event. Anyone who can move the tag in that window gets their tree built and pushed as `ghcr.io/pku-yuangroup/openai4s:0.2.0` **and** `:latest` — exactly the window `release.yml` now closes with `"$TAG moved to $TAG_SHA after this run froze $SHA"`. `scripts/verify_release_tag.py` compares the tag *string* to declared versions, so a moved tag whose tree still declares `0.2.0` passes it; nothing here peels to a commit, requires an annotated tag, or asserts ancestry of `origin/main`.

This checkout also omits `persist-credentials: false`, so the `packages: write` token stays in `.git/config` while `scripts/container_smoke.sh` builds and runs a container from that tree — the hazard `finalize`'s own new comment names.

And no gate can see it: `grep -rn publish-image` returns zero hits under `tests/`, and all three new assertions read `_workflow("release.yml")`.

Not applied — it is a separate workflow outside this diff. Two smaller things while you are in there: the `case` guard `v[0-9]*.[0-9]*.[0-9]*` is an unanchored glob (it matches `v1.2.3;anything`), where `release.yml`'s new `^v[0-9]+\.[0-9]+\.[0-9]+$` is not.

openai4s/cli/main.py
```diff
@@ -1046,7 +1046,13 @@
...
     payload = result.as_dict()
     if args.json:
-        print(json.dumps(payload, ensure_ascii=False, indent=2))
+        # OnboardingResult.as_dict() is an explicit secret-free projection:
+        # has_api_key is a boolean and the credential value never reaches it,
+        # which tests/test_onboarding.py asserts against a live secret. The
+        # suppression has to ride the alert's own line -- a CodeQL suppression
+        # scope is that one line, so a comment above it suppresses nothing.
+        rendered = json.dumps(payload, ensure_ascii=False, indent=2)
+        print(rendered)  # codeql[py/clear-text-logging-sensitive-data]
```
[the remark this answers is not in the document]
Nobody-Zhang: Removed the alert at its source in `6f9c7c1` rather than suppressing it: reproduced it locally with the CodeQL CLI 2.26.4 and the flow is entirely name-based (`self.has_api_key` is classified "password" by its name while the value is a `bool` the whole way), so `OnboardingResult.as_dict()` now coerces with `bool(...)` — a rebuilt database confirms this alert is gone and the four unrelated pre-existing ones are unchanged. The `# codeql[...]` marker is deleted with it, since code scanning does not honour suppression comments at all.

Nobody-Zhang:
**Answers to the four points raised in the reviews, three of them verified rather than reasoned about.**

**1. Artifact CSP on PDFs — refuted; the profile *fixes* the PDF preview rather than breaking it.** Tested in Chromium against a probe server serving a real PDF into a same-origin iframe with no iframe `sandbox` attribute, exactly as `app.js:8661` builds it:

| response profile | result |
| --- | --- |
| pre-PR (`frame-ancestors 'none'`, `X-Frame-Options: DENY`) | `GET /pdf → 200 OK [FAILED: net::ERR_BLOCKED_BY_RESPONSE]`, console: *"Framing … violates … frame-ancestors 'none'"* |
| this PR (`frame-ancestors 'self'` + `SAMEORIGIN` + response `sandbox`) | loads; the Chrome PDF viewer renders the document, toolbar and thumbnail |

So the PDF iframe was already broken before this PR, by the UI shell's frame denial, and the `SAMEORIGIN` / `frame-ancestors 'self'` override is what makes it work. The response-level `sandbox` does not stop the built-in viewer.

**2. `freeze` looking the tag up locally — refuted.** That checkout keeps `fetch-depth: 0`, and at the pinned `actions/checkout` SHA (v7.0.1) `fetchDepth <= 0` takes `getRefSpecForAllHistory`, which is `['+refs/heads/*:refs/remotes/origin/*', tagsRefSpec]` with `tagsRefSpec = '+refs/tags/*:refs/tags/*'`. The `--no-tags` that v7 always adds only disables *automatic* tag following, not an explicit refspec — I ran the exact command checkout assembles against a repo with an annotated tag and all 7 tags arrived, after which `git rev-parse --verify "$TAG^{commit}"` and `git cat-file -t "$TAG"` both succeed. No publish fails closed.

**3. Prefix URL check in the retrosynthesis graph — fixed** in `6f9c7c1`. `text.startsWith("http")` before `a.href = text` also accepts `httpfoo:`; anchored to `/^https?:\/\//`.

**4. Playwright 1.62.1 vs `engines.node >=18` — real, documented rather than changed.** `engines` is the floor for the published Skill installer, not for the browser gates, so raising it would move a consumer-facing constraint to fix a contributor-facing one. The Node 20 requirement is now stated in `CLAUDE.md` and `tests/README*.md`, where the `npm ci` instruction lives. Happy to raise `engines` instead if you prefer that.

The failing **CodeQL** check is fixed in the same commit, at its source — see the reply on the inline thread.

tests/test_artifact_workbench.py
```diff
@@ -373,6 +375,33 @@
...
+    and it should not be interpolated into executable source.
+    """
+    import re
+
+    document = ketcher_document(
+        Config(roadmap_features=RoadmapFeatureFlags(stage9_artifact_workbench=True)),
+        {"artifact_id": ["art-1"]},
+    ).decode("utf-8")
+
+    assert re.search(r"<script(?![^>]*\bsrc=)[^>]*>", document) is None
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in `973e66a` by dropping the hand-rolled tag pattern entirely: the assertion now uses `test_security_headers._ScriptInventory`, the `HTMLParser` this repo already wrote for the same question, which lower-cases tag names by construction. Confirmed by putting `<SCRIPT>var sneaky = 1;</SCRIPT>` back into the wrapper — the old regex passed it, the parser fails on it — and a rebuilt CodeQL database reports zero `py/bad-tag-filter` results across the tree. The catch was a fair one: the assertion is an *absence*, so a pattern that misses a case variant would report the very thing it exists to catch as fixed.

PR #133, opened 2026-08-27
feat(webui): add host-bound interactive artifact previews

Nobody-Zhang:
Interactive HTML reports now run in the current Preact Artifact Viewer using a signed, scoped URL on the alternate loopback origin. A report cannot reuse that URL on the authenticated Workbench origin: the server rejects the Host before returning executable bytes.

The completed implementation includes:

- Grants bind a nonempty root frame, the exact primary Artifact version, the minting origin, the distinct spend origin, and a one-hour deadline. The response permits only the minting origin as a framing ancestor. Unsupported hosts, absent credentials, and unavailable snapshots keep the inert preview.
- Preview bytes come from verified immutable snapshots. Relative resources resolve within the granted frame, including when another session uses the same filename. The primary document remains pinned after edits; sibling resources use their current captured versions.
- The Preact viewer starts inert and upgrades only after validating the grant origin and path. Same-origin or arbitrary-origin overrides are refused, historical version links stay exact, and delayed responses cannot update a closed viewer. The frozen legacy shell keeps inert previews.
- Ketcher stays on the authenticated app origin. Real browser editing exposed a separate initialization failure in its pinned chemistry bindings; JavaScript string compilation is permitted only for `/static/vendor/ketcher/index.html`. The wrapper, Workbench, and Artifact policies retain their stricter script rules.
- The mint route's successful response is captured from real handler behavior in both response-contract artifacts. Security/API documentation describes the actual boundary, including the script-readable bearer URL and iframe self-navigation disclosure risk. Grant requests are redacted in request logs and granted bytes use `no-referrer` / `no-store`.

## Validation

- `npm test --prefix frontend`: 517 passed; committed dist rebuilt with `npm run build --prefix frontend`.
- `node tests/browser_sandbox_preview.mjs`: passed in both `127.0.0.1 → localhost` and `localhost → 127.0.0.1` directions, using the production Preact Viewer and a disposable daemon. Covers script/image execution, parent isolation, cross-frame refusal, actual iframe self-navigation to the app origin, historical versions, inert fallback, API refusal even with cookies on both hosts, and Ketcher molecule edits persisted through Save.
- Focused authorization/CSP/Gateway regressions, real successful response-contract coverage, and browser egress audit passed.
- Full offline suite through `capture_response_schemas.py --check`: **8,383 passed, 43 skipped**, with 1,163 captured response shapes covering **212/212 routes** and no breaking schema change.
- `pre-commit run --all-files`, `capture_response_contract.py --check`, all 38 PR-tier offline harness scenarios, bilingual directory coverage, source credential scan, and wheel/sdist build plus release-artifact verification passed.

- [GitHub CI on `a4e2c8bc`](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34017698214) passed: all 34 applicable PR checks are green, with 3 nightly-only checks skipped by configuration. This includes Python 3.10/3.12/3.13/3.14, Chromium/Firefox/WebKit, frozen response shapes, container smoke, and both Linux sandbox gates.

## Review notes

No hard third-party dependency was added to core. Gateway edits are surgical; vendored code and captured fixtures were not reformatted. The only removed built file is replaced by the new Vite asset. No test files were deleted. The branch incorporates current `main`.

Please focus security review on `sandbox_grants.py`, the grant/static-document adapters in `gateway.py`, and the exact Ketcher exception in `security_headers.py`. The remaining bearer disclosure/shared browser-origin limits are explicit in `docs/security.md`.


Nobody-Zhang:
@cursoragent  review this pr.

Nobody-Zhang (commented):
xhigh code review (10 finder angles, per-candidate verification, gap sweep). The six earlier cursor[bot] findings are closed at a4e2c8bc — origin swap (spend origin signed and Host-bound, ancestor = minting origin, browser test performs the swap), Ketcher on the app origin, empty `root_frame_id` refused, docs/tests, same-origin override — verified independently.

15 findings below, all fixed on this branch in 3b3a4bb1 (server/tests) and f4720f62 (frontend/harness/docs). Most severe: the exact-version inert fallback loading the raw API byte route (siblings 404, content type not forced), a top-level-open + cookie path that widened the grant's read scope, inert/granted sibling resolution disagreeing, and a Ketcher race that makes the new browser gate red on a fast host.

Not changed (deliberate, listed for the record): sibling resources resolve to their *current* captured version under a pinned primary — documented and tested, only the in-code comment overstated it (reworded); `sandbox_delivery` is a second `CompletionDeliveryService` because `runner.completion_delivery` is `None` unless Stage 1 is on (commented); the loopback pair is still a literal table on both sides (now one `LOOPBACK_PAIRS` on the server); mint-time + spend-time hashing of the primary and the missing `(root_frame_id, filename)` index are left as-is.

frontend/src/features/artifacts/preview.ts
```diff
@@ -0,0 +1,52 @@
...
+    override,
+  );
+}
+
+/** The inert preview remains usable if an interactive grant is unavailable. */
+export function renderHtmlPreview(content: HTMLElement, a: ArtifactRow): void {
+  const frame = el("iframe");
+  applyArtifactIframeSandbox(frame, "html-preview");
+  const exactVersion = a._exactVersion && a.version_id ? a.version_id : "";
+  frame.src = exactVersion ? artUrl(a) : htmlPreviewSrc("", a.id);
```
Nobody-Zhang: **[correctness, medium] Exact-version inert fallback loads the raw API byte route.**

For a deep link with `_exactVersion` this loads `/api/v1/artifacts/versions/<vid>`, which neither forces `text/html` nor resolves relative siblings: every `` becomes `/api/v1/artifacts/versions/figure.png` → 404 (`gateway.py` `_serve_artifact`, the `versions/` branch has no filename fallback), and a version row stored as `text/plain` renders as source under `nosniff`. That is the final state wherever the grant is refused (Docker `0.0.0.0`, LAN/https, no token). Before this PR the renderer always used `/preview/<id>`.

`/preview/<version-id>` already pins the bytes, forces HTML and lets siblings resolve as `/preview/figure.png`.

Fixed in f4720f62: the inert src is `htmlPreviewSrc(exactVersion || a.id)`; `preview.test.ts` pins `/preview/version-1`.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -14522,6 +14572,86 @@ def _stream_file(
                     self.wfile.write(chunk)
 
         # ---- artifact bytes --------------------------------------------
+        def _serve_sandbox_grant(self, path: str) -> None:
```
Nobody-Zhang, line 14599: **[correctness/security, low] A grant opened top-level on a cookie-bearing loopback name can read other frames' artifacts.**

Framed, this is safe: the frame is cross-site with the top-level app origin, so `SameSite=Strict` keeps `os_token` off every request it makes (the browser test's `api-fetch-refused` therefore measures `connect-src 'none'`, not the cookie). But if the user has logged in on both loopback names (the test sets exactly that up) and opens a grant URL as a *top-level* tab on the sandbox name, `">` is same-site, authenticates through the token gate, resolves across frames via `artifact_by_unique_filename`, has no `Cross-Origin-Resource-Policy`, is readable through an untainted canvas, and a document may always navigate itself (`source === target`) despite the missing `allow-top-navigation`. Broader than the documented one-frame bearer scope.

Fixed in 3b3a4bb1: `_serve_sandbox_grant` answers 404 whenever the request presents the daemon's own cookie (`os_token` / `os_user`) — a legitimate framed spend never carries one, and cookies of other local apps on the hostname don't count. Documented in `docs/security.md`; the harness comment now says what the probe actually proves.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -14522,6 +14572,86 @@ def _stream_file(
...
+            if not remainder.startswith("/preview/"):
+                self._json({"error": "not found"}, 404)
+                return
+            ident = unquote(remainder[len("/preview/") :])
+            if ident == grant.artifact_id:
+                meta = store.version_meta(grant.version_id)
+            else:
+                meta = store.get_artifact(ident)
+                if meta is None:
+                    matches = store.list_artifacts(
```
Nobody-Zhang: **[correctness, low-medium] Inert and granted previews resolve a sibling filename differently.**

This route resolves a sibling by exactly-one match *within the frame*; the inert `/preview/` route keeps store-wide uniqueness (`artifact_by_unique_filename`). With a `figure.png` in this frame and another in any other session, first paint (`/preview/figure.png`) is a broken image and the same `` loads once the grant lands — and stays broken wherever the upgrade never happens. Verified by driving both routes in the same store state (`test_a_sandbox_grant_is_minted_scoped_and_spendable` builds it but only asserts this route).

Fixed in 3b3a4bb1: `store.artifact_by_unique_filename(filename, root_frame_id=None)` is the one query both routes use; the inert route asks within the referring report's frame first (`Referer`, which `Referrer-Policy: same-origin` guarantees on same-origin subresource loads), then falls back to the old store-wide rule. The test now asserts both routes.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

tests/browser_sandbox_preview.mjs
```diff
@@ -0,0 +1,400 @@
...
+      const edited = await editor.evaluate(() => Promise.race([
+        window.ketcher.getMolfile(),
+        new Promise((_, reject) => setTimeout(() => reject(new Error("Ketcher serialization timed out")), 10000)),
+      ]));
+      return edited.includes(` ${atom} `);
+    }, 30000);
+    result.ketcher_edited_atom = atom;
+    await ketcherFrame.locator("#ketcher-save").click();
+    await waitUntil("Ketcher artifact save", async () =>
+      (await ketcherFrame.locator("#ketcher-status").innerText()).startsWith("saved "));
```
Nobody-Zhang: **[test-reliability] The Ketcher phase races the wrapper's deferred reload and accepts an unchanged save.**

`ketcher-page.js` calls `loadArtifact` twice — on the editor's `init` message and again 400 ms after the frame `load` — and both end with the same `loaded <id>` status, so the `startsWith("loaded ")` wait at L263 is satisfied by the first while the second `setMolecule(original)` lands after the edit. The save then posts the original molfile, the wrapper reports `saved <vid> (unchanged)`, `startsWith("saved ")` accepts it, and the gate goes red blaming persistence. 3/3 local runs failed both directions; CI happened to pass. An HTTP-only repro persists fine.

Fixed in f4720f62: let the deferred reload land, require the edit to survive four consecutive reads (re-applying if wiped), and refuse an `(unchanged)` status.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

frontend/src/features/artifacts/preview.ts
```diff
@@ -0,0 +1,52 @@
...
+  const suffix = exactVersion ? `?version_id=${encodeURIComponent(exactVersion)}` : "";
+  api(`/artifacts/${encodeURIComponent(a.id)}/sandbox-grant${suffix}`, { method: "POST" })
+    .then((grant) => {
+      // A user may close the viewer or switch artifacts before the POST ends.
+      if (!frame.isConnected || frame.parentNode !== content) return;
+      const src = grantedHtmlPreviewSrc(origin, grant, a.id);
+      if (!src) return;
+      frame.setAttribute("sandbox", INTERACTIVE_ARTIFACT_SANDBOX);
+      frame.src = src;
+      note.remove();
```
Nobody-Zhang: **[correctness, low] The caption is removed before the upgraded navigation lands.**

`note.remove()` runs synchronously on assigning the sandbox-origin `src`. If that spend 404s (snapshot altered or artifact deleted between mint and load, or a frame reload after the 1 h TTL) the JSON 404 goes out with the shell profile (`X-Frame-Options: DENY`, `frame-ancestors 'none'`), the browser refuses to render it, the inert document is already gone, and the user gets an unexplained blank frame. A `load` listener can't tell success from refusal on a cross-origin frame.

Fixed in f4720f62: the caption is retargeted (new `viewer.renderer.interactive` string) instead of removed.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/security_headers.py
```diff
@@ -62,6 +58,64 @@ def artifact_content_security_policy() -> str:
...
+            f"frame-ancestors {ancestors}",
+            # `allow-same-origin` is safe here and not on the app origin: this
+            # document's origin is the sandbox, so keeping it buys `'self'` for
+            # sub-resources without granting reach into anything.
+            "sandbox allow-scripts allow-same-origin",
+        ]
+    )
+
+
+def sandboxed_artifact_security_headers(
```
Nobody-Zhang: **[altitude, low] The plural `app_origins` signature still admits the reviewed origin-swap hazard.**

The first commit passed `_app_origins()` here (both loopback names → the HIGH finding). At HEAD the only caller passes a one-tuple, so the sequence form exists only to re-open the swap: `sandboxed_artifact_security_headers(_app_origins())` type-checks and emits both names with no `X-Frame-Options`; `(('',))` emits an empty `frame-ancestors` source list. `grant.app_origin` is already validated as exactly one loopback origin.

Fixed in 3b3a4bb1: singular `app_origin: str`, `ValueError` on anything but one whole `http://` origin, unconditional XFO pop; tests cover the two-origin, empty and injection shapes.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

tests/test_security_headers.py
```diff
@@ -165,37 +168,62 @@ def test_the_shell_itself_is_still_unframeable():
...
 
-def test_html_preview_iframe_does_not_reenable_artifact_capabilities(index_html):
+def test_legacy_preview_remains_inert(index_html):
     app_js = index_html.with_name("app.js").read_text(encoding="utf-8")
-    preview_line = next(
-        line for line in app_js.splitlines() if 'rendererId === "html-preview"' in line
-    )
+    body = _html_preview_renderer(app_js)
+    assert 'frame.setAttribute("sandbox", "")' in body
+    assert 't("viewer.renderer.noscript")' in body
```
Nobody-Zhang, line 206: **[test-coverage] The bilingual `viewer.renderer.noscript` assertion was dropped.**

The deleted test asserted both the zh and en rows exist in `app.js`; the replacement only checks the `t(...)` call. Deleting either row leaves this file and `test_webui_static_contract.py` green (verified by mutation).

Fixed in 3b3a4bb1: the two-marker loop is back inside `test_legacy_preview_remains_inert`.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

tests/test_egress_surface.py
```diff
@@ -380,14 +380,21 @@ def test_skill_egress_scan_recognizes_each_supported_client_family():
...
-#: a reason, so a third entry has to argue for itself.
+#: Absolute URLs the client may *name*, each with its bounded purpose. A
+#: string is not egress; a string handed to something that fetches it is.
+#: Most entries are inert. The loopback pair constructs the separately tested
+#: scoped preview origin; it cannot be replaced with an arbitrary destination.
 #:
 #: `vendor/` is excluded from the scan entirely: it is upstream minified code,
 #: and a URL inside a bundled library is not the client choosing to call it.
 _WEBUI_NAMED_HOSTS = {
+    "localhost": "constructs the exact alternate loopback preview origin at "
```
Nobody-Zhang, line 391: **[test-coverage] Loopback hosts are allowlisted for the whole webui tree.**

`_WEBUI_NAMED_HOSTS` is keyed by bare hostname and applied to every `.js/.html/.css` under `webui/`, so a port-less `const base = "http://127.0.0.1"` in `app.js` with a `fetch(base + ":8760/…")` on another line passes both egress tests at HEAD (verified by mutation) and failed at `origin/main`. Only the compiled `resolveSandboxOrigin` needs these two names.

Fixed in 3b3a4bb1: `_webui_host_accounted` admits them only under `dist/assets/` next to the `hostname===` comparison.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -14509,7 +14555,11 @@ def _stream_file(
                 # its own headers instead of going through _send, so it has to
                 # opt in explicitly — and it takes the same `security` profile
                 # as `_serve_file`, or the two writers of one fact drift.
-                profile = security if security is not None else security_headers()
+                profile = (
+                    security
+                    if security is not None
+                    else security_headers(frame_src=_app_origins())
```
Nobody-Zhang: **[altitude, low] Two of five default-profile writers still lack `frame-src`.**

The `frame_src` default was pasted into `_send`, `_serve_index` and this 200 branch, but `_send_static_bytes` (L14428) and this function's own 304 branch (L14532) still emit bare `security_headers()`, so one function answers 200 and 304 with different CSPs (empirically: a hashed >8 MiB script). Today only fingerprinted non-document assets reach the bare writers, so nothing user-visible breaks yet; the first hashed HTML entry would ship `frame-src 'self'` and have its preview upgrade refused.

Fixed in 3b3a4bb1: one `_default_security()` next to `_app_origins()`, used at all five sites.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -14522,6 +14572,86 @@ def _stream_file(
...
+                ident == grant.artifact_id
+                and meta.get("artifact_id") != grant.artifact_id
+            ):
+                self._json({"error": "not found"}, 404)
+                return
+            # Always serve a captured version. Never read mutable workspace
+            # bytes (or substitute a newer head) behind an already minted URL.
+            try:
+                body = sandbox_delivery.read_verified_snapshot(meta)
+            except DeliveryValidationError:
```
Nobody-Zhang: **[correctness, low] Snapshot validation failures on the grant routes leave no diagnostic.**

This handler and the mint-side `verify_snapshot` 409 swallow `DeliveryValidationError`, while the app-origin exact-version read records it (`record_diagnostic(..., surface="artifact:exact_version_read")`). A damaged snapshot behind a minted grant is then `http_request path=/sandbox/[redacted] status=404` — byte-identical to a forged or expired token — with nothing to pair via `X-Request-Id`.

Fixed in 3b3a4bb1: `artifact:sandbox_read` / `artifact:sandbox_grant` surfaces (daemon-side only; responses unchanged), asserted in the snapshot test.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/security_headers.py
```diff
@@ -90,11 +144,47 @@ def embeddable_security_headers() -> dict[str, str]:
...
+    """Allow the pinned first-party editor's generated chemistry bindings.
+
+    Ketcher 3.7.0 uses JavaScript string compilation as well as WebAssembly.
+    Without this permission its initialization raises EvalError and structure
+    import/export waits forever. Only the exact vendored editor document uses
+    this profile; its blob worker inherits it. The wrapper, Workbench, and
+    all untrusted Artifact responses keep their existing no-eval policies.
+    """
+    headers = embeddable_security_headers()
+    headers["Content-Security-Policy"] = "; ".join(
```
Nobody-Zhang: **[simplification] The Ketcher eval profile rewrites the CSP by string surgery.**

Splitting the finished policy on `"; "` and appending to whatever starts with `script-src ` depends on the builder's private join and yields the redundant `'wasm-unsafe-eval' 'unsafe-eval'` (`'unsafe-eval'` already permits wasm compilation), pinned literally by two tests; the wasm comment above the builder no longer describes this document. Also worth saying in the PR: this is a drive-by fix of a bug already on `main` (Ketcher was vendored on 2026-08-21 under the no-eval profile and its bundle uses `new Function`), whose only regression assertion is the chromium-gated Ketcher phase of the sandbox harness.

Fixed in 3b3a4bb1: `content_security_policy(allow_eval=)` / `embeddable_security_headers(allow_eval=True)`; the editor's directive is exactly `script-src 'self' 'unsafe-eval'`.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

frontend/src/features/artifacts/preview.ts
```diff
@@ -0,0 +1,52 @@
...
+  content.appendChild(note);
+
+  // Revalidate the store at use time; compatibility window exports can write it.
+  const origin = sandboxOrigin.value && resolveSandboxOrigin(
+    typeof location === "undefined" ? null : location,
+    sandboxOrigin.value,
+  );
+  if (!origin) return;
+  const suffix = exactVersion ? `?version_id=${encodeURIComponent(exactVersion)}` : "";
+  api(`/artifacts/${encodeURIComponent(a.id)}/sandbox-grant${suffix}`, { method: "POST" })
```
Nobody-Zhang, line 89: **[efficiency] Every render navigates the frame twice and re-mints a still-valid grant.**

The inert app-origin document and all its siblings are fetched, parsed and rendered, then discarded milliseconds later when the grant resolves; and every render of the same artifact/version (each tab switch back to the report wipes the pane) POSTs again, which server-side hashes the whole primary snapshot, although grants live 3600 s.

Fixed in f4720f62: the frame waits for the grant and navigates once (the inert document is loaded only when the origin is inert or the grant is refused), and a per-`id:version` cache reuses the granted URL until a minute before expiry.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

frontend/src/features/artifacts/preview.ts
```diff
@@ -0,0 +1,52 @@
...
+  resolveSandboxOrigin,
+} from "../../islands/frames";
+import { sandboxOrigin } from "../../stores/session";
+import { api, el, translate } from "./api";
+import { artUrl } from "./cache";
+import type { ArtifactRow } from "./types";
+
+/** Initialize the production workbench using the same verified origin policy. */
+export function initializeSandboxOrigin(target?: Record<string, unknown>): void {
+  const config = target?.__OPERON__;
```
Nobody-Zhang: **[simplification] Dead `sandboxOrigin` plumbing: override knob, ignored parameters, legacy comment.**

Nothing produces `__OPERON__.sandboxOrigin` (`git grep` hits only this file and its test), so `initializeSandboxOrigin`, the render-time revalidation, `resolveSandboxOrigin`'s override branch and five tests defend a channel that cannot be reached; `frames.ts` keeps `_origin` parameters every caller feeds `""`; `app.js` L111–117 describes an origin derivation the frozen shell no longer performs, next to a write-only `S.sandboxOrigin`.

Fixed in f4720f62: `resolveSandboxOrigin(location)` computed at render time, the override and the dead parameters removed, the orphaned comment and dead `try` deleted from `app.js`.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -13247,7 +13268,8 @@ def _send(
             self.send_response(code)
             self.send_header("Content-Type", _sanitize_header_value(ctype))
             self.send_header("Content-Length", str(len(body)))
-            self.send_header("Cache-Control", "no-cache")
+            if not security or "Cache-Control" not in security:
```
Nobody-Zhang: **[test-coverage] The `_send` `Cache-Control` special case is unverified on the wire.**

Making this `send_header` unconditional again leaves the targeted tests green (the new gateway tests stub `_send` and inspect the profile dict) while a granted response goes out with both `no-cache` and `no-store`; and `extra` is still written after the profile, so a caller passing `Cache-Control` through `extra` emits two contradictory headers.

Fixed in 3b3a4bb1: `_send` writes one header per name by dict merge (`{"Cache-Control": "no-cache", **profile, **extra}`), asserted on the wire in `test_send_emits_exactly_one_cache_control_per_response`.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

tests/browser_sandbox_preview.mjs
```diff
@@ -0,0 +1,400 @@
...
+async function main() {
+  const port = await allocateLoopbackPort();
+  assertion(port !== 8760, "the acceptance must not use the user's daemon port");
+  const dataDir = fs.mkdtempSync(path.join(os.tmpdir(), "openai4s-sandbox-preview-"));
+  const appOrigin = `http://127.0.0.1:${port}`;
+  const daemon = spawn(pythonPath,
+    ["-m", "openai4s", "serve", "--no-browser", "--port", String(port)], {
+      cwd: workspaceRoot,
+      env: {
+        ...process.env,
```
Nobody-Zhang: **[test-reliability] The disposable daemon inherits the shell environment and discards its logs.**

`...process.env` lets a real `OPENAI4S_*_API_KEY`, `TEAM_MODE`, `KERNEL_SANDBOX=enforce` or `UNATTENDED_APPROVAL` from the developer's shell drive the security-gate daemon (a live key is the documented way local browser gates time out), and `stdout/stderr.resume()` drops the only server-side evidence — the Ketcher failure above produced a SUMMARY with nothing from the daemon, and CI's "Show daemon log" step only reads the `:8760` daemon.

Fixed in f4720f62: `minimalChildEnvironment` and `boundedLogCollector` exported from `browser_auth.mjs`; the daemon log tail rides the SUMMARY on failure.

_🤖 Review by [Claude Code](https://claude.com/claude-code) (xhigh); fix applied on this branch._

openai4s/server/gateway.py
```diff
@@ -13545,7 +13624,63 @@ def _stream_file(
...
+                return
+            ctype = meta.get("content_type") or _guess_ctype(Path(target).name)
+            if str(ident) == str(meta.get("artifact_id") or "") and str(
+                ctype
+            ).startswith("text/html"):
+                ctype = "text/html; charset=utf-8"
+            self._serve_file(
+                Path(target),
+                ctype,
+                security=sandboxed_artifact_security_headers(_app_origins()),
```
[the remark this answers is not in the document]
Nobody-Zhang: Closed at a4e2c8bc (spend origin signed into the grant and verified against `Host`, `frame-ancestors` = the one minting origin, the browser gate performs the hostname swap and asserts 404), re-verified in this review; hardened further in 3b3a4bb1 — the header helper now takes exactly one origin by type and fails closed, and a spend that presents the daemon's own session cookie is refused so the top-level-open case cannot reach the authenticated API either.

Nobody-Zhang:
## Follow-up: three defects in the review fixes themselves (6adb14ca)

Before reporting the previous batch as applied I ran an independent read-only verification of each fix against the tree. Thirteen of fourteen were present and sound; the pass also found that three of the fixes had introduced problems of their own, and two class fixes had reached only one call site. All are repaired in [6adb14ca](https://github.com/PKU-YuanGroup/OpenAI4S/commit/6adb14ca).

**Defects introduced by the fixes**

- **Stale grant cache (user-visible).** The cache keyed on the exact-version *flag*, not on the version the grant pins. After a new capture, `syncArtifactVersion` rewrites the row's `version_id` without setting `_exactVersion`, so the head preview hit the cache and replayed superseded bytes for up to an hour under a banner naming the new version. Pre-fix behaviour was correct. Now keyed on the row's version identity.
- **Cookie guard failed open.** It used `SimpleCookie`, which abandons the remainder of a header it cannot parse (`SimpleCookie('a="unterminated; os_token=x')` yields no names). The previewed document runs `allow-same-origin` on that hostname, so it could set one malformed cookie and hide the real `os_token` behind it. The docstring asserted the opposite of what CPython does. Now scans the raw header for the two names.
- **Referer rule over-reached.** It lived in `_serve_artifact`, which also serves `/api/v1/artifacts/<name>`, so it widened that route's resolution as well. Now gated on `force_html`, which only the preview route sets, and an absent `Host` is refused instead of matching `""` against `""`.

**Class fixes that had reached one call site**

- `Cache-Control` one-header-per-name landed in `_send` only; `_send_static_bytes` and both `_stream_file` branches would still double-emit under a profile carrying one. All three resolve it identically now.
- `minimalChildEnvironment` / `boundedLogCollector` reached one of the two daemon-owning harnesses; `browser_stage1_trusted_delivery.mjs` kept private copies and now imports the shared ones.
- The daemon log was attached only when `main()` throws, but `runDirection` swallows its own assertions into `result.error` — the likeliest way this gate goes red, moments before the data directory is deleted. Now attached whenever the run is not green, and redacted, because `openai4s serve` prints its `?token=` URL. Verified by forcing that exact failure: exit 1 with a 299-byte log and the token replaced by `<redacted>`.

**Coverage added** for what nothing pinned: the default shell policy at every unprofiled writer including the 304 branch, the negative half of the egress host scoping, the cookie-evasion header, and the stale-grant regression. Each was mutation-checked — reverting its fix turns it red.

**Gates green at 6adb14ca:** full offline suite (0 failures), `pre-commit --all-files`, PR-tier harness 38/38, source secret scan, response contract 212/212, bilingual directory docs, frontend 509 tests + `tsc` + rebuilt `dist`, and `browser_sandbox_preview.mjs` in both loopback directions including real Ketcher edit/save.

One residual is now stated plainly in `docs/security.md` rather than implied away: the cookie refusal closes the reachable path (address-bar or bookmark open, where the cookie is sent), not every path — a cross-site top-level navigation withholds a `SameSite=Strict` cookie from the navigation while the loaded document's own same-origin subresources still carry it. Nothing in the Workbench links a grant URL that way and the response grants no popup or top-navigation permission, so the document cannot arrange it.

_🤖 Follow-up by [Claude Code](https://claude.com/claude-code)_

PR #144, opened 2026-09-02
release: next → main — auto-repair budget authority, faithful Skill import, Python 3.14 in the matrix, a no-progress circuit, and explicit remote cancel

Nobody-Zhang:
## Summary

The next-version report's full scope, landed on `next` across three waves of
parallel lanes. Five core items plus both capacity candidates and the P2
design freeze:

| Item | What it closes |
|---|---|
| **P0-1** | `repair_turn` was in `CONSUMERS` with a published limit, a projected `used`, and no reserve call anywhere — the field read as measured while measuring nothing. Now wired, and `inspect_budget_wiring()` fails closed on any unwired consumer, so Stage 12 can no longer report `ga_ready` over a hole. |
| **P0-2** | Importing a `SKILL.md` silently dropped `requirements`, `capabilities.network`, `license`, `category` and unknown nested keys. The raw document is now the rewrite seed, and an illegal network mode is a stable 400 with zero versions written. |
| **P0-3a** | Python 3.14 is the container's interpreter but was in no full matrix. Now in the offline matrix and the release gate; the six provenance fixtures are fixed with real venv prefixes rather than relaxed assertions; Linux evidence is split into `ci_attestation` and `release_reexecution`, which cannot contradict each other. |
| **P1-1** | A generic session could repeat the same action until `max_turns`. A durable no-progress circuit terminates as `no_progress`, rebuilt from the Action Ledger across restart and compaction. Completion truth is structural: the finish path forces `completion=None` on that branch. |
| **P1-4** | Task Center could see remote jobs but not cancel them. Explicit confirm-gated cancel; only a confirmed remote receipt maps to `cancelled`, and unreachable / unconfirmed / unsupported all stay `cancel_indeterminate` with "may still be running and billing". |
| **P1-2** | The dashboard loaded every project and the store did a COUNT/MAX per row. Keyset paging with search; team visibility enters SQL before `ORDER BY`/`LIMIT` so hidden rows cannot occupy page slots. |
| **P1-3** | No terminal-free way to diagnose. Passive status GET, explicit checks POST, admin-only deny-by-default bundle. |
| **P2-1** | Schema, limits, threat model, and a real old-importer probe. Conclusion: v1 optional files are safe to enable. Export stays off this version. |

Amends decision D8: it kept the full Poll/Harvest/Cancel/Reconcile workbench
at P2 with no public API this version. The report promotes the Cancel half to
core scope, so `POST /frames/{fid}/compute/tasks/{job_id}/cancel` is now a
public route. Harvest and Reconcile remain P2. `docs/plan-crosswalk.json` says
so rather than reading as if the panel were still read-only.

### Defects found while integrating, not in the plan

* **The cancel envelope's `status`.** `public_failure` adds the integer HTTP
  status and defers to a value the route set — a deliberate carve-out for one
  legacy route. P1-4's 409 set `status` to the job's terminal state, clobbering
  the 409 and making compute-cancel the second route declaring
  `["integer", "string"]`. `task["status"]` already carried it, so the envelope
  copy was removed.
* **A fourth response writer the contract prober did not stub.** P1-3 needs
  `Cache-Control: no-store`, which `_json` hard-wires as `no-cache`, so it
  sends through `_send_static_bytes` — which raises on the synthetic handler
  and gets swallowed. All three `/diagnostics/*` routes published bare 404s,
  `/bundle` included, which returns a zip. This is the same failure the
  `_stream_file` comment in `response_capture.py` records; the test written for
  that occurrence caught this one.
* **A frozen type nothing can observe.** `GET /orchestration/jobs/{id}` reports
  `allocation.reason` as `["null", "string"]`, but the route emits `allocation`
  only while it is active and a reason lands there only mid-cancel, so every
  recapture narrowed it to `"null"` — restored by hand four times this release.
  Now declared in `UNOBSERVABLE_WIDER_TYPES` and applied at write time.

## Area

- [x] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [x] Security-related change
- [x] Release / packaging

## Verification

Ran:

```text
uv run pytest -n auto --maxprocesses=4 --dist loadfile   # 8222 passed, 24 skipped, exit 0
uv run mypy                                              # clean
uv run pre-commit run --all-files                        # all hooks pass, no files modified
python3 scripts/source_secret_scan.py                    # 3753 files
uv run python3 -m harness.cli run --tier pr --offline    # 38/38
uv run python scripts/check_directory_readmes.py         # clean
uv run python scripts/capture_response_contract.py --check   # 211/211 routes
uv run python scripts/capture_response_schemas.py            # 211/211 coverage, regenerated
cd frontend && npx tsc --noEmit && npx vitest run         # 63 files, 445 tests
cd frontend && npm run build                              # dist rebuilt, no drift vs source
```

Not run:

```text
P0-3b: the real non-publishing release rehearsal (publish=false,
       macos_asset=omit workflow_dispatch on a candidate SHA)
node tests/browser_smoke.mjs and the rest of the browser E2E matrix
harness.smoke.linux_bwrap_interrupt
```

Reason:

```text
The rehearsal needs hosted-CI permissions and is a human step by design; this
PR delivers the schema, gates and tests that make its receipt verifiable, not
the run. Report section 12 keeps it in the definition of done.

The browser, container and Linux-sandbox gates are CI jobs: they need a
daemon on :8760, Docker, and Linux+bwrap respectively. macOS-local runs of
them prove less than the CI runs this PR will trigger.
```

## Risk and Reviewer Focus

Highest risk, in order:

1. **`openai4s/server/gateway.py`** — four lanes touched it in disjoint
   regions (`_loop()`, `cancel_compute_task()` + route, `GET /projects`
   handler, diagnostics route wiring). Each merged without conflict; worth
   confirming no region silently lost the other's edit.
2. **`openai4s/server/auto_budget.py` + `auto_repair.py`** — reserved counts
   against remaining exactly like committed, so every reserve needs a settling
   path that does not depend on another reservation existing. That coupling is
   what broke `review_rounds` in the previous release.
3. **`openai4s/server/response_capture.py`** — two changes to the thing that
   produces the frozen contracts: a new writer stub and a write-time widening
   table. A mistake here makes the contracts wrong rather than absent.
4. **`docs/plan-crosswalk.json`** — three closed rows were rewritten because
   P1-4 contradicts decision D8. That is a judgement call, not a mechanical
   re-audit.

Not covered by any gate here: the browser E2E matrix and container smoke run
in CI on this PR, not locally.

### Diff Readiness

- [ ] This PR is small and single-purpose.
      `host_dispatch.py`, `store.py`, `worker.py`, `manager.py`.
      source.
      reformatted.

Not small and single-purpose: this is a release branch carrying one planned
scope. Each work item is its own lane commit under a merge commit, so it
reviews per item.

### Tests

      a documented smoke test.

One test was rewritten rather than deleted:
`test_projects_route_has_no_pagination_semantics` asserted `?limit&offset` had
no effect, which report section 9 changes deliberately — `offset` is executed
during the compatibility window. The surviving half (no parameters means no
paging) is what the window exists for; `tests/test_project_paging.py` owns the
parameterized side.

### Core Dependency Policy

      stdlib web server.
      in-tree use site.

No new dependencies.

      verification notes.
      guarantees.

Security-relevant items carry their own tests: `raw_required` refused at load
and at every sink; diagnostics bundle canaries (secret / absolute path /
prompt / research) absent from the zip, temp files back to baseline on
complete, disconnect and error; a member's 403 is identical whether or not an
admin bundle is in flight; project paging applies team visibility inside SQL.

### Docs

      appropriate.
      affected.

---

> [!NOTE]
> **Low Risk**
> Changes are mostly CI expressions, release receipt metadata, and documentation or contract artifacts; no application runtime logic appears in this diff.
> 
> **Overview**
> **CI and release plumbing** fixes the offline-matrix `fetch-depth` expression so Python **3.12** alone gets full git history for `test_plan_crosswalk.py` (the previous `cond && 0 || 1` form always resolved to depth 1). Release workflows now pass **`--workflow-run-id`** and **`OPENAI4S_RELEASE_INPUTS_JSON`** into `release_receipts.py` for dist, macOS, Linux, Windows, staging, and finalize, binding artifacts to the dispatch run. **Dockerfile** and **docs/platforms.md** document **3.14** as the container interpreter aligned with the offline matrix and quality gates; **publish-image** and **release-validation** clarify that container smoke is not a substitute for the 3.14 suite and that Linux full bwrap boundary is attested via `ci-linux-sandbox-full`.
> 
> **Documentation and contracts:** **architecture.md** freezes a future optional **`artifact_observations.json`** v1 schema (export still off), limits, importer compatibility table, and threat model. **plan-crosswalk.json** rows are updated so explicit **P1-4 cancel** is reflected versus decision D8 (panel no longer read-only). **response-contract.json** adds **`/diagnostics/*`** routes and **`POST .../compute/tasks/.../cancel`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9d723d8e3755412ad756c6fc829ed9bc69b3afa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openai4s/server/gateway.py
```diff
@@ -10890,6 +11031,25 @@ def _auto_budget_chat(messages, cfg, **kwargs):
             return denied.reason
         st.last_engine_completion = result.completion
         st.last_model_prose = events.model_prose
+        if result.stop_reason == NO_PROGRESS_STOP_REASON:
```
Nobody-Zhang: **[correctness] `no_progress` is projected as a completed turn.** `run_message` initialises `status = "completed"` and only `loop_reason == "max_turns"` (line 9859) flips it to `failed`. A `no_progress` return from `_loop` falls through: the frame/turn record stays `completed` with no failure code, `gate_armed = status == "completed" and gate_requested` arms the Stage 4 review gate on a non-answer, and the Action Ledger's terminal for the same turn is `failed` -- two writers disagreeing about one turn. Same gap in `delegation.py:1622` (a `no_progress` child is `finish_done`) and `host_dispatch._delegate_result_status`.

Applied fix: a sibling `elif loop_reason == NO_PROGRESS_STOP_REASON` branch in `run_message` (status `failed`, code `no_progress`, the notice as `err_text`, streamed once), the notice block removed from `_loop`, and `no_progress` added beside `max_turns` in `delegation.py` / `_derive_task_status` / `host_dispatch`. Regression test: `tests/test_gateway.py::test_no_progress_stop_is_a_failed_turn_with_a_stable_code`.

openai4s/server/response_capture.py
```diff
@@ -693,6 +758,20 @@ def _stub_stream(file_path, ctype, extra=None, security=None):
...
+
+    # The fourth writer, arriving the same way the third did. P1-3's
+    # diagnostics routes need `Cache-Control: no-store`, which `_json`
+    # hard-wires as `no-cache`, so they send through `_send_static_bytes` --
+    # which, like `_stream_file`, builds headers straight on the
+    # BaseHTTPRequestHandler and raises inside `send_response` on this
+    # synthetic handler. Left unstubbed, the diagnostics routes publish
+    # whatever their *other* verbs produced, which is the exact lie the
+    # comment above describes. Mirrors the real method's signature.
+    handler._send_static_bytes = (
```
Nobody-Zhang: **[correctness/contract] The `_send_static_bytes` stub records JSON as opaque bytes, so no diagnostics 200 shape is frozen.** `observe_raw` only notes kind/status/content-type; `docs/response-schemas.json` has no `GET /diagnostics/status [ok]` or `POST /diagnostics/checks [ok]` key (only the wrong-verb `[error]` rows) while `docs/response-contract.json` lists the route as `json`. `install()` (the suite-side recorder) does not wrap `_send_static_bytes` at all, so a later change to `passive_status()`'s keys passes `--check` unnoticed.

Applied fix: both the probe stub and `install()` now decode an `application/json` body and call `recorder.observe`, falling back to `observe_raw` otherwise; the stubbed diagnostics/cancel tests are marked `stubbed_backend` so their fabricated bodies do not get frozen, and the schemas file is regenerated.

tests/test_compute_task_centre.py
```diff
@@ -286,3 +293,324 @@ def test_refresh_is_a_post_because_it_harvests(client):
...
+    )
+    assert status == 202
+    assert body["outcome"] == "cancel_indeterminate"
+    assert body["reason"] == "provider_cancel_unsupported"
+    assert "billing" in body["hint"]
+    assert client.store.get_compute_job("job-open")["status"] != CANCELLED
+    assert calls == [{"job_id": "job-open"}]
+
+
+def test_a_natural_completion_race_returns_the_real_terminal_state(client, monkeypatch):
```
Nobody-Zhang, line 489: **[conventions] Stub-driven route tests without `@pytest.mark.stubbed_backend` froze fabricated shapes.** CLAUDE.md: "A test that stubs a service must be marked `stubbed_backend` ... without it a stub's fabricated shape gets published as the route's contract". This test (and `test_unsupported_cancel_is_indeterminate` at 456, `test_cancel_without_confirm_does_not_contact_the_remote` at 340) replaces `build_dispatcher` with a hand-written `_Manager`; the recorder wraps `_api`, so `docs/response-schemas.json` now carries `outcome`/`task` on the cancel `[error]` shape and `hint`/`reason` on `[ok]` from these stubs. Same in `tests/test_diagnostics_routes.py` (`doctor.report` and `write_bundle_file` stubs at 416/516/573/594).

Applied fix: markers added on all seven tests; the real-manager tests (`_submit_ssh_job` + patched `Popen`) still exercise the 200/202 shapes through real code.

frontend/src/features/sessions/load.ts
```diff
@@ -35,12 +41,135 @@ import {
...
+    !canLoadMoreProjects({
+      loadingMore: !!_projectsLoadingMore.value,
+      hasMore: !!projectsHasMore.value,
+      cursor: projectsNextCursor.value,
+    })
+  ) {
+    return;
+  }
+  const gen = ++_projectsLoadGen;
+  const q = normalizeProjectQuery(String(projectsQuery.value || ""));
```
Nobody-Zhang: **[correctness] `loadProjects()` applies the dashboard search filter to every caller, and `projects.value` is now one 100-row page.** `projectsQuery` is read unconditionally, and nothing clears it except Escape in the search box. Type `alpha` on the dashboard, then create/open project `beta`: `openProject`/`createProject` (`projects.ts:292/311`), `conversation.ts:87`, `actions.ts:556` and `Wizard.tsx:331` all call `loadProjects()` → `GET /projects?limit=100&q=alpha` → `projName()` falls back to the placeholder name, the switcher omits `beta`, the session labels on the dashboard itself lose their project. Independently, with >100 projects every non-dashboard consumer (`projName`, switcher, `host.ts:86`, `Wizard.tsx:247`, `attention/api.ts:36`) now sees only the first activity-ordered page where the old route ignored `limit` and returned everything; the legacy `app.js:6766` shell sends `?limit=100&offset=0` and is silently truncated with no pager.

Applied fix (query leak only): `loadProjects({ q })` takes the query explicitly; only the dashboard passes `projectsQuery.value`, appends continue the query the page was loaded with, and the search input is debounced (150 ms). **Not fixed:** the >100-projects truncation for non-dashboard consumers and the legacy shell -- that is the P1-2 design (a first page of 100) and needs a product decision (a full-list name map for those readers, or a documented cap).

openai4s/compute/manager.py
```diff
@@ -3301,24 +3321,103 @@ def _terminate_sandbox(self, pid: str, sandbox_id: str) -> None:
...
+                        f"job {job['job_id']!r} has no recorded sandbox to "
+                        f"terminate; it may still be running and billing",
+                        "receipt_unconfirmed",
+                        indeterminate=True,
+                    )
+                try:
+                    self._terminate_sandbox(rest, sandbox_id)
+                except ComputeError as error:
+                    raise self._cancel_failure(error) from error
+        except ComputeError as error:
```
Nobody-Zhang, line 3404: **[correctness] `cancel()` maps only `ComputeError`; any other exception from the terminate path escapes unmapped.** A BYOC helper that exits before reading stdin (`BrokenPipeError` in `_spawn_helper`), a reply file that is not JSON, or an `OSError` from `Popen` is not a `ComputeError`, so this `except` is skipped: no `cancel_unconfirmed` event, no `reason`, the job stays `running`, and `_compute_guard` hands the kernel a bare error with no `error_kind`/`indeterminate` -- the SDK reads it as a definite rejection while the gateway answers 202 `cancel_indeterminate`.

Applied fix: a trailing `except Exception` wraps the failure as `ComputeError(..., "receipt_unconfirmed", indeterminate=True)`, records the unconfirmed cancel, and re-raises; `_compute_guard` now forwards `indeterminate` explicitly so the SDK no longer has to infer it from a cancel-only kind set (a `transient` submit was read as `indeterminate=False`).

frontend/src/components/customize/NestedEditor.tsx
```diff
@@ -169,33 +179,65 @@ function SkillImport() {
...
+            disabled={saving}
+            onClick={async () => {
+              if (!content.trim()) return;
+              setSaving(true);
+              try {
+                const r = await api("/skills/import", {
+                  method: "POST",
+                  body: JSON.stringify({ content }),
+                });
+                setReview(rec(r.review));
```
Nobody-Zhang, line 235: **[correctness] `dropSkillsCatalog()` moved to the Close button only.** Before the PR a 200 from `/skills/import` dropped the cached catalog immediately. Now it is dropped only by the new Close button; dismissing the review pane via the backdrop (line 34), the header X (44), Escape (`Customize.tsx:57`) or `actions.ts:18/30` leaves `skillsCatalog` populated with the pre-import list, so composer autocomplete (`composer.ts:62`), the palette (`palette.ts:186`) and send-time mention resolution (`send.ts:107`) do not know the new Skill until something else drops the cache.

Applied fix: `dropSkillsCatalog()` right after the successful POST (the write has already landed); the Close handler is unchanged.

openai4s/storage/frames.py
```diff
@@ -90,6 +91,137 @@ def visible_session_clause(
...
+        if payload.get("fp") != fingerprint:
+            raise ValueError("cursor filter mismatch")
+        project_id = payload.get("id")
+        if not isinstance(project_id, str) or not project_id:
+            raise ValueError("missing project id")
+        if "t" not in payload:
+            raise ValueError("missing last_active_at")
+        stamp = payload["t"]
+        if stamp is not None:
+            stamp = int(stamp)
```
Nobody-Zhang, line 153: **[correctness] An unbounded cursor `t` passes decoding and raises `OverflowError` at bind time → 500 instead of the documented 400 `invalid_cursor`.** The cursor is unsigned and trivially forgeable: `{"fp": sha256({"q":"","team_scope":""}), "id": "p1", "t": 10**30}` decodes fine, then `sqlite3` raises `OverflowError: Python int too large to convert to SQLite INTEGER`, which is not a `ValueError`.

Applied fix: the decoder refuses a stamp outside the signed 64-bit range as `ValueError("last_active_at out of range")`.

scripts/release_receipts.py
```diff
@@ -232,6 +442,20 @@ def verify_stage_attestation(
             f"stage attestation is for version {document.get('version')!r}, not "
             f"{version!r}"
         )
+    if not str(document.get("workflow_run_id") or document.get("source_sha") or ""):
```
Nobody-Zhang: **[correctness] The run-id requirement is vacuous.** `workflow_run_id or source_sha` is always truthy for a schema-2 attestation (it always carries a source SHA), so an attestation built with no `--workflow-run-id` and no `GITHUB_RUN_ID` is accepted -- while `verify_build_receipts` strictly refuses the same omission, and binding the workflow run is the property schema 2 was introduced to add.

Applied fix: both facts checked on their own (`workflow_run_id` required; `candidate_sha`/`source_sha` required); the two fixtures in `tests/test_release_gate_manifest.py` now pass a run id, and `tests/test_release_rehearsal.py::test_stage_attestation_without_a_workflow_run_id_is_refused` pins it.

.github/workflows/ci.yml
```diff
@@ -257,7 +259,7 @@ jobs:
           # depth-1 checkout does not contain those objects, so the check can
           # only skip. Deepening one job makes it a check CI actually performs
           # instead of one that is enforced on developer machines and nowhere
-          # else; deepening all three would pay for the same answer twice more.
+          # else; deepening all four would pay for the same answer twice more.
```
Nobody-Zhang, line 282: **[correctness, pre-existing] `fetch-depth: ${{ ... == '3.12' && 0 || 1 }}` is always 1.** In the expressions language `0` is falsy, so `true && 0` falls through to `|| 1`. The 3.12 job's checkout log on this PR's CI run shows `fetch-depth: 1` / `git fetch --depth=1`, so `tests/test_plan_crosswalk.py`'s baseline/ancestry assertions skip on every row (invisible under `-q`), and the comment this PR edits ("deepening all four would pay...") restates a deepening that has never happened.

Applied fix: `${{ matrix.python-version != '3.12' && 1 || 0 }}` (truthy middle operand). The crosswalk tests pass unskipped in a full clone at this merge ref, so the strict branch goes green rather than red.

openai4s/server/auto_budget.py
```diff
@@ -273,6 +269,15 @@ def inspect_budget_wiring() -> dict[str, Any]:
             continue
         if "auto_budget" not in source:
```
Nobody-Zhang: **[correctness] `inspect_budget_wiring()` proves a sink by the substring `auto_budget`, so `repair_turn` reads as wired against a method that never reserves it.** `AutoRepairService.run` passes because it contains `self._auto_budget()`; the only `reserve(... consumer="repair_turn")` is in `RepairTurnAdmission.admit_turn`, which runs only if an injected metered `repair_fn` chooses to call it. `model`/`native_tool` were likewise registered against `_loop`, which reserves neither. Deleting `UNWIRED_CONSUMERS["repair_turn"]` therefore turned a documented hole into a `ga_ready` claim.

Applied fix: the registry names the method that actually reserves each consumer (`_invoke_model_with_auto_budget`, `_invoke_control_with_auto_budget`, `RepairTurnAdmission.admit_turn`) and the check is structural (`consumer="<name>"` must appear in that method's source). **Still open:** whether a metered `repair_fn` calls `admit_turn` at all cannot be verified structurally; `_classify_repair_fn` now counts a third parameter only when it is required, so a legacy `fn(snapshot, findings, extra=None)` is no longer mis-classified as metered.

openai4s/server/response_capture.py
```diff
@@ -164,6 +164,67 @@ def route_for(path: str, patterns=None) -> str | None:
...
+
+    def apply(node: Any, path: tuple[str, ...]) -> Any:
+        if not isinstance(node, dict):
+            return node
+        result = dict(node)
+        properties = result.get("properties")
+        if isinstance(properties, dict):
+            result["properties"] = {
+                key: (
+                    {**apply(value, path + (key,)), **overrides[path + (key,)]}
```
Nobody-Zhang: **[altitude] `UNOBSERVABLE_WIDER_TYPES` replaces the observed `type` wholesale, so it permanently hides any real change to the field it names.** Both the frozen file and the fresh capture go through `document()`, so if `allocation.reason` regresses to an integer/object the capture is rewritten to `["null","string"]` and `--check` sees equal sets. The table was meant to restore a member the suite cannot see; as written it also erases a member the suite did see.

Applied fix: the declared members are unioned into the observed type (`_widened`), so `null` ∪ declared still yields `["null","string"]` while a genuinely new member surfaces as drift.

openai4s/server/gateway.py
```diff
@@ -10881,7 +11006,23 @@ def _auto_budget_chat(messages, cfg, **kwargs):
...
         )
+        from openai4s.agent.ledger import restore_progress_circuit
+        from openai4s.agent.progress_circuit import (
+            NO_PROGRESS_STOP_REASON,
+            ProgressCircuit,
+            attach_progress_circuit,
+        )
+
+        try:
+            restored_circuit = restore_progress_circuit(
```
Nobody-Zhang, line 11014: **[efficiency] This restore reads the whole branch ledger, JSON-decoding every event, on every Web message -- and is empty by construction.** `run_message` calls `action_ledger.append_user` (line 9748) before `_loop`, and `_epoch_groups` keeps only groups after the latest `user` group, so the rebuilt circuit is always `ProgressCircuit()` here; the read is O(session) per turn for nothing (the same happens on the CLI path in `loop.py`). A second latent issue for the day a mid-epoch resume exists: ledger `proposed` events store `_redact_value(call.arguments, keys)` while the live fingerprint hashes raw arguments, so a restored streak cannot chain with the next live identical call whenever a secret-named key is present.

Applied fix (cost only): `restore_progress_circuit` first scans the branch without events to find the last `user` ordinal and reads only the epoch's groups with events (`after_ordinal`), falling back to the fork-aware full read. The redacted-vs-raw fingerprint mismatch is reported, not fixed.

scripts/release_pipeline.py
```diff
@@ -2194,8 +2276,29 @@ def main() -> int:
...
+        "OPENAI4S_RELEASE_INPUTS_JSON", ""
+    )
+    if raw_inputs:
+        try:
+            inputs = json.loads(raw_inputs)
+        except ValueError as error:
+            print(f"::error::workflow inputs are not JSON: {error}", file=sys.stderr)
+            return 2
+
     pipeline = Pipeline(
```
Nobody-Zhang, line 2368: **[correctness] `Pipeline(...)` is built outside any `try`, and its constructor calls `normalize_workflow_inputs`, which raises `ReceiptError`.** `OPENAI4S_RELEASE_INPUTS_JSON='{"macos_asset": "signed"}'` (or a JSON list) parses as JSON, then dies with an uncaught traceback and exit 1 instead of the `::error::` annotation + exit 2 the adjacent `json.loads` failure gets.

Applied fix: the inputs are validated right after the parse with the same `::error::` / `return 2` path.

openai4s/server/diagnostics_routes.py
```diff
@@ -0,0 +1,251 @@
...
+        """``ok``, ``busy``, or ``rate``. ``ok`` must be paired with ``end``."""
+        now = self.clock()
+        with self._lock:
+            if self._in_flight:
+                return "busy"
+            last = self._last.get(principal)
+            if last is not None and (now - last) < self._cooldown_s:
+                return "rate"
+            self._in_flight = True
+            self._last[principal] = now
```
Nobody-Zhang, line 87: **[correctness] The per-principal cooldown is stamped before generation succeeds and never rolled back.** A 413 (`BundleTooLarge`) or a generation exception answers with no bundle, but `_last[principal]` is already `now`, so the admin who trims the logs and retries ten seconds later gets `429 wait 60 seconds between diagnostic bundle downloads` for a download that never happened.

Applied fix: `end(principal, delivered=...)` forgets the stamp when nothing shipped; a disconnect mid-stream keeps it (the generation cost was paid).

openai4s/server/gateway.py
```diff
@@ -3304,6 +3305,130 @@ def refresh_compute_task(self, root_frame_id: str, job_id: str) -> dict:
...
+            record_diagnostic(
+                error, surface="compute:provider", request_id=correlation_id()
+            )
+            raise GatewayError(
+                503, "remote compute is not available here", "no_provider"
+            ) from error
+
+        from openai4s.compute.manager import ComputeError
+
+        def _indeterminate_reason(error: BaseException) -> str:
```
Nobody-Zhang: **[reuse/simplification] `_indeterminate_reason` re-implements `ComputeManager._cancel_failure`, and the two `except` blocks below are identical.** `manager.cancel` already maps every non-`not_found` failure onto the closed cancel-reason set, so the gateway can read `error.error_kind` and build the envelope once; the closed set itself is now written three times (`manager._INDETERMINATE_KINDS`, `manager.CANCEL_INDETERMINATE_REASONS`, `sdk/compute._INDETERMINATE_CANCEL_KINDS`). Related cleanups applied in the same pass: the `/projects` handler no longer reaches into `store._frames` (a `Store.list_projects(...)`/`count_projects` facade plus a new `server/project_listing.py` own the parameter table and paging), `ProgressCircuit.would_block_action` and `_is_reasoning_insert` were mutation-verified dead and removed, `security_posture` builds on `passive_security_posture`, the dead `already_terminal` success branch in `island.ts` is gone, and `downloadDiagnosticsBundle` defers `revokeObjectURL` like every other object-URL site in the tree.

openai4s/compute/manager.py
```diff
@@ -3301,24 +3321,103 @@ def _terminate_sandbox(self, pid: str, sandbox_id: str) -> None:
...
+            "could not reach" in message
+            or "timed out" in message
+            or "timeout" in message
+            or "deadline" in message
+        ):
+            return ComputeError(str(error), "remote_unreachable", indeterminate=True)
+        return ComputeError(str(error), "receipt_unconfirmed", indeterminate=True)
+
+    def _record_unconfirmed_cancel(self, job: dict, error: ComputeError) -> None:
+        """Leave the job live and record that a cancel was asked, not granted."""
```
Nobody-Zhang: **[correctness] The unconfirmed-cancel note is never cleared, and it can overwrite a terminal row.** `_record_unconfirmed_cancel` writes `reason="cancel could not be confirmed; the job may still be running and billing"` through `_persist` (an unconditional UPDATE). The confirmed cancel that follows commits `CANCELLED` with `termination_reason` but no `reason=`, so the row -- and the Task Centre, which renders `task.reason` under the row -- keeps saying a cancelled job may still be running and billing. Race variant: if the poll thread commits `succeeded`/`failed` between the in-memory pre-check and the failed terminate, the terminal row's own `reason` is overwritten and a `cancel_unconfirmed` event is appended to a finished job.

Applied fix: the note is written only while the stored row is still live, and the confirmed-cancel commit passes `reason=""`. Regression test: `tests/test_compute_task_centre.py::test_a_confirmed_cancel_clears_the_earlier_unconfirmed_note`.

openai4s/skills_loader/capabilities.py
```diff
@@ -189,6 +199,48 @@ def legacy_capability(*, source: str = "inventory") -> NetworkCapability:
...
+        domains = []
+    if not isinstance(domains, list):
+        raise InvalidSkillCapability("capabilities.network.domains must be a list")
+    cleaned: list[str] = []
+    for item in domains:
+        if not isinstance(item, str) or not str(item).strip():
+            raise InvalidSkillCapability(
+                "capabilities.network.domains must be a list of domains"
+            )
+        cleaned.append(item)
```
Nobody-Zhang, line 240: **[correctness] The strict import gate blesses a flow-style `network: {mode: host_only, domains: [a.example, b.example]}` that the parser mangles.** `_parse_inline_network` splits the mapping on every comma before it looks at the `[...]` list, so the parts are `mode: host_only`, ` domains: [a.example`, ` b.example]`; the second domain is dropped (no `:`), the first keeps its `[`, and `validate_network_frontmatter_strict` only checks that each item is a non-empty string -- import returns 200 with `capabilities.domains == ['[a.example']`, and the runtime `host_only` allowlist matches neither intended domain. The block-style spelling of the same declaration parses correctly, so two YAML spellings yield different allowlists. Pre-existing in the tolerant reader, but the PR's 400 gate is what now certifies it.

Applied fix: the inline mapping is split on top-level commas only (`_split_top_level`, bracket-depth aware); `tests/test_skill_network_contract.py` pins both readers on the flow-style declaration.

tests/test_diagnostics_routes.py
```diff
@@ -0,0 +1,676 @@
...
+
+
+def _inject_canaries(cfg, monkeypatch):
+    monkeypatch.setenv("HOST_PRIVATE_SLOT_ORANGE", SECRET)
+    logs = Path(cfg.data_dir) / "logs"
+    logs.mkdir(parents=True, exist_ok=True)
+    (logs / "app.out").write_text(
+        json.dumps({"HOST_PRIVATE_SLOT_ORANGE": SECRET, "prompt": PROMPT_CANARY})
+        + "\n"
+        + f"{PATH_CANARY} {RESEARCH_CANARY}\n",
```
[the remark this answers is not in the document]
Nobody-Zhang: Renamed the planted canary in 023c1945: it is a canary the bundle test proves is scrubbed, not a stored secret, so it is now `PLANTED_CANARY` and no longer reads as clear-text storage.

tests/test_skill_frontmatter_edit.py
```diff
@@ -138,6 +138,74 @@
...
+    )
+    meta, body = _parse_frontmatter(out)
+    assert meta["name"] == "cryo-import"
+    assert meta["origin"] == "user"
+    assert meta["origin"] != "draft"
+    assert meta["requirements"] == "[gpu, cuda]"
+    assert meta["license"] == "CC-BY-4.0"
+    assert meta["category"] == "structural-biology"
+    assert "mode: host_only" in out
+    assert "api.openalex.org" in out
```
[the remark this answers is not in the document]
Nobody-Zhang: Rewrote the assertion in 023c1945 to check the preserved domain line whole (`"      - api.openalex.org" in out.splitlines()`) rather than as a URL substring; the test never sanitized anything.

tests/test_skill_frontmatter_edit.py
```diff
@@ -138,6 +138,74 @@
...
+    assert "doi.org" in out
```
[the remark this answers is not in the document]
Nobody-Zhang: Same as above in 023c1945: the `doi.org` line is asserted whole via `splitlines()`, not as a substring of a URL.

.github/workflows/ci.yml
```diff
@@ -300,6 +304,14 @@ jobs:
...
         if: matrix.python-version == '3.13' || matrix.python-version == '3.14'
         run: uv sync --locked --extra science
+      # The container's interpreter. Job-level `if:` cannot see `matrix`,
+      # so the extra install stays a step-level condition like the rows
+      # above. Science is not optional: without it the science-dependent
+      # modules importorskip and this job would stop validating the
+      # runtime the published image actually runs.
+      - name: Install dev and science dependencies
+        if: matrix.python-version == '3.14'
+        run: uv sync --locked --extra science
```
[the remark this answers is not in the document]
Nobody-Zhang: Removed the 3.14-only step in 59e4f4d0; the 3.13-or-3.14 step is the single science install now, with the comment explaining why one condition rather than two.

openai4s/server/gateway.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/server/gateway.py:9853` now has the sibling `elif loop_reason == NO_PROGRESS_STOP_REASON` branch (failure code at 9862), with the matching remaps in `delegation.py` and `host_dispatch.py`. Resolving so the required-thread-resolution gate can pass.

openai4s/server/response_capture.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/server/response_capture.py:508` decodes an `application/json` body before `recorder.observe`, and `install()` wraps `_send_static_bytes` the same way (578/610); `docs/response-schemas.json` now carries the diagnostics `[ok]` shapes. Resolving so the required-thread-resolution gate can pass.

tests/test_compute_task_centre.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `tests/test_compute_task_centre.py` carries 3 `stubbed_backend` markers and `tests/test_diagnostics_routes.py` 4; the fabricated cancel/diagnostics shapes were dropped from `docs/response-schemas.json`. Resolving so the required-thread-resolution gate can pass.

frontend/src/features/sessions/load.ts
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `frontend/src/features/sessions/load.ts:139` is `loadProjects(opts?: { append?: boolean; q?: string })`; only the dashboard passes `q`, every other caller loads unfiltered. The 100-row page for non-dashboard consumers stays a documented product limit. Resolving so the required-thread-resolution gate can pass.

openai4s/compute/manager.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/compute/manager.py:3435` has the trailing `except Exception` that wraps the failure as `ComputeError(..., "receipt_unconfirmed", indeterminate=True)` (3352), so the gateway and the SDK read the same indeterminate outcome. Resolving so the required-thread-resolution gate can pass.

frontend/src/components/customize/NestedEditor.tsx
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `frontend/src/components/customize/NestedEditor.tsx:145/209/234` call `dropSkillsCatalog()` right after the successful POST, so any dismissal path sees the fresh catalog. Resolving so the required-thread-resolution gate can pass.

openai4s/storage/frames.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/storage/frames.py:164` refuses an out-of-range stamp as `ValueError("last_active_at out of range")`; probed live against a daemon at this head: the forged `t=10**30` cursor answers `400 invalid_cursor`, not 500. Resolving so the required-thread-resolution gate can pass.

scripts/release_receipts.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `scripts/release_receipts.py:165` checks `workflow_run_id` on its own; `tests/test_release_rehearsal.py::test_stage_attestation_without_a_workflow_run_id_is_refused` pins it. Resolving so the required-thread-resolution gate can pass.

.github/workflows/ci.yml
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `.github/workflows/ci.yml:288` is `${{ matrix.python-version != '3.12' && 1 || 0 }}`; the merge of main (a9d723d8) kept this expression over main's `&& 0 || 1` form. Resolving so the required-thread-resolution gate can pass.

openai4s/server/auto_budget.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/server/auto_budget.py` registry is now structural (`consumer="<name>"` at the reserving method, see the note at line 55 and the check at 395), and `repair_turn` is back in `UNWIRED_CONSUMERS`. Resolving so the required-thread-resolution gate can pass.

openai4s/server/response_capture.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/server/response_capture.py:208` `_widened` unions the declared members into the observed type instead of replacing it. Resolving so the required-thread-resolution gate can pass.

openai4s/server/gateway.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/agent/ledger.py:849` `restore_progress_circuit` reads only the current epoch (`progress_circuit.py:341` `_epoch_groups`) instead of decoding the whole branch ledger per message. The redacted-vs-raw fingerprint mismatch stays a documented latent seam (every entry appends a user group first). Resolving so the required-thread-resolution gate can pass.

scripts/release_pipeline.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `scripts/release_pipeline.py:2306` validates the inputs with `normalize_workflow_inputs` inside the same `::error::` / `return 2` path as the JSON parse. Resolving so the required-thread-resolution gate can pass.

openai4s/server/diagnostics_routes.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/server/diagnostics_routes.py:90` `end(principal, delivered=...)` forgets the cooldown stamp when no bundle shipped (413 or a generation error); a mid-stream disconnect keeps it. Resolving so the required-thread-resolution gate can pass.

openai4s/server/gateway.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `_indeterminate_reason` has 0 occurrences left in `openai4s/server/gateway.py`; the envelope is built once from `error.error_kind`. The `/projects` handler goes through `Store.list_projects`/`count_projects` and `openai4s/server/project_listing.py`. Resolving so the required-thread-resolution gate can pass.

openai4s/compute/manager.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/compute/manager.py:3354` `_record_unconfirmed_cancel` now re-reads the row and returns without writing when it is already terminal, and the confirmed stop commits through `_commit_terminal(..., reason="")` (3452) so the "may still be running and billing" note is cleared once the cancel lands. Resolving so the required-thread-resolution gate can pass.

openai4s/skills_loader/capabilities.py
Nobody-Zhang: Fixed in 909b2e3b (the 'Applied fix' paragraph above) and re-verified at the current head a9d723d8, after the merge of main: `openai4s/skills_loader/capabilities.py:344` `_split_top_level` splits the flow mapping on commas outside `[...]`, so `{mode: host_only, domains: [a.example, b.example]}` yields both domains; `tests/test_skill_requirements.py` pins the flow-list spelling. Resolving so the required-thread-resolution gate can pass.

openai4s/agent/progress_circuit.py
```diff
@@ -0,0 +1,455 @@
...
+        if not calls:
+            return
+        if all(_call_is_malformed(call) for call in calls):
+            self.malformed_streak += 1
+            self._maybe_trip()
+            return
+        self.malformed_streak = 0
+        fingerprint = fingerprint_native_calls(calls)
+        if fingerprint == self.same_action_fingerprint:
+            self.same_action_streak += 1
```
Nobody-Zhang, line 256: **[correctness] The same-action streak counts a repeated poll whose result changed.** `fingerprint_native_calls` hashes tool name + canonical arguments only, and nothing here consults the result content, so `collect_children(timeout=300)` for children spawned with `wait=false`, `exec_peek(exec_id)` and `compute_result(job_id)` — tools designed to be re-called with identical arguments — reach `SAME_ACTION_THRESHOLD` on the third poll and the Engine returns `no_progress` with `completion=None`; `run_message` projects it as a failed turn ("Stopped repeating actions"). Reproduced in-process: three `collect_children` batches answering running-10% / running-55% / completed trip on the third, the one that delivered the child's output. The ledger reconstruct path (`_observe_native_group`) has the same blind spot.

Applied fix: a `same_result_digest` (name, error flag, whitespace-normalized content per row; no number collapsing, since "10%" → "55%" is the delta a poll exists to observe). The same call counts only when the answer did not move; a group without result rows keeps counting on the call alone. The persisted `result` rows carry content, so restore applies the identical rule. Pinned by three tests (moving results do not trip; identical results still trip; ledger restore applies the rule), mutation-checked.

openai4s/skills_loader/frontmatter_edit.py
```diff
@@ -39,10 +39,16 @@
...
 
 def _is_top_level_key(line: str) -> bool:
     """A `key:` at column 0. Indented lines belong to the block above."""
-    return bool(line) and not line[:1].isspace() and ":" in line
+    if not line or line[:1].isspace() or line.lstrip().startswith("#"):
+        return False
+    return ":" in line
 
 
 def _key_of(line: str) -> str:
```
Nobody-Zhang, line 54: **[correctness] The editor matches owned keys case-sensitively; the loader lowercases keys and keeps the last occurrence.** `_drop_field` compares `_key_of(line) == key`, so a document written with `Name:` / `Description:` / `Origin: personal` is not recognised as owning those fields: fresh `name:`/`description:`/`origin: user` lines are prepended and the capitalised originals are preserved after them — and `_parse_frontmatter` (`loader.py:156`, `key = k.strip().lower()`, last wins) reads the originals. Verified by executing `rewrite_import` + `_parse_frontmatter`: the result loads as `name='Old Name'`, `origin='personal'`. A Web Customize edit that does not take effect, and an import stamped `user` that loads as `personal` and then keeps `personal` on every later edit.

Applied fix: `_key_of` folds the way the loader folds (`.lower()`), so every spelling of an owned key is replaced. Pinned by `test_capitalised_owned_keys_are_replaced_not_duplicated`.

openai4s/server/project_listing.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def _first(query: Mapping[str, Any], key: str) -> str | None:
+    """The first value of a ``parse_qs``-shaped parameter, or None if absent/empty."""
+    values = query.get(key) or [None]
+    value = values[0] if values else None
+    return None if value in (None, "") else str(value)
+
+
+def _int_param(raw: str, *, name: str, minimum: int, code: str) -> int:
```
Nobody-Zhang: **[correctness] `_int_param` bounds `offset` only from below, so `?offset=9223372036854775808` reaches the SQLite bind and raises `OverflowError` — a 500 for a request the module's closed failure table promises a 400.** Reproduced against a live daemon at this head (`internal_error` + traceback at `frames.py:528`). It is the unguarded sibling of the cursor-`t` overflow 909b2e3b already closed in `decode_project_cursor`. `int()` also admits `1_0`, `+3`, ` 3` and full-width digits (`str.isdigit()` would not help: `'３'.isdigit()` is `True`).

Applied fix: an ASCII `-?[0-9]+` form check and an explicit `maximum` (SQLite's `2**63-1`) for `offset`; `limit` keeps its documented cap. `test_limit_and_offset_errors` now covers the overflow and the lenient forms, and fails at the pre-fix code.

frontend/src/features/sessions/dashboard.ts
```diff
@@ -77,7 +150,33 @@ export function renderDashProjects(): void {
...
+    hasMore: !!projectsHasMore.value,
+    loadingMore: !!_projectsLoadingMore.value,
+  });
+  if (view.kind === "error") {
+    const box = el("div", "dash-empty", projectCopy("error"));
+    const retry = el("button", "outline-btn small", projectCopy("retry"));
+    retry.type = "button";
+    retry.id = "dash-projects-retry";
+    retry.onclick = () => {
+      void loadProjects().then(() => renderDashProjects());
```
Nobody-Zhang: **[correctness] The error-state Retry reloads without the search query.** Every other dashboard load passes `{ q: projectsQuery.value }`; this one calls `loadProjects()`, so after a failed *filtered* load Retry paints the whole directory under a box that still shows the filter, `loadProjects` resets `_projectsLoadedQuery` to `""`, and every Load-more then walks the unfiltered set. The rename path in `projects.ts:353` has the same shape: `await loadProjects()` (no `q`) followed by `binds.renderDashProjects()` when the dashboard is visible.

Applied fix: Retry re-issues `searchProjectsNow()` (the exact request the box would make); the rename path loads with the box's query when the dashboard is visible. `openProject`/`createProject`/deep links keep loading unfiltered by design — they end in the workspace and need the whole directory for the header and switcher.

frontend/src/features/sessions/load.ts
```diff
@@ -35,12 +40,151 @@ import {
...
+ * project name in the header, the project switcher and the session labels.
+ * Reading the search signal here for all of them leaked a dashboard filter
+ * into places that had no search box to clear it from.
+ */
+export async function loadProjects(opts?: { append?: boolean; q?: string }): Promise<void> {
+  const append = !!opts?.append;
+  if (
+    append &&
+    !canLoadMoreProjects({
+      loadingMore: !!_projectsLoadingMore.value,
```
Nobody-Zhang: **[correctness] A Load-more is admitted while a replace (search) request is in flight, and the newer search reply is then discarded.** The gate only reads `_projectsLoadingMore`, which a replace never sets, and the list is not repainted until the reply — so the old Load-more button stays clickable during the 150 ms debounce + request. The click takes generation N+1 with the *old* `_projectsLoadedQuery` and cursor; the search reply hits `gen !== _projectsLoadGen` and is thrown away; page two of the previous filter is merged and rendered under the new box text. Escape has the mirror case.

Applied fix: a `_projectsReplaceInFlight` flag set for the life of a replace (cleared only by the newest generation's `finally`), read by both gates; `canLoadMoreProjects` keeps its signature. Pinned by `load.replace.test.ts` (search issued first, append clicked second, append resolved first → refused; the search reply lands), which goes red with the gate reverted.

scripts/release_pipeline.py
```diff
@@ -1460,6 +1488,33 @@ def step_verify(self) -> StepResult:
...
+            notary = macos_receipt.get("notary")
+            if isinstance(notary, Mapping):
+                try:
+                    notary = release_receipts.normalize_notary(notary, required=False)
+                except release_receipts.ReceiptError as error:
+                    raise ReleaseError(str(error)) from error
+            else:
+                notary = None
+            macos_asset = "omit"
+            if isinstance(macos_receipt.get("workflow_inputs"), Mapping):
```
Nobody-Zhang, line 1533: **[correctness] The workflow run id and the dispatch inputs are recorded on every schema-2 build receipt and never compared.** `verify_build_receipts` requires `workflow_run_id` non-empty; `step_verify` reads exactly one input from the receipt (`macos_asset`, preferring the receipt's copy over the live dispatch); `publish`/`pypi_only` are read nowhere; `verify_rehearsal` has no caller outside tests. So a receipt from an earlier dispatch, or one built under `publish=true` during a `publish=false` rehearsal, verifies as long as the SHA and the digests match — and `docs/release-validation.md` sells the binding as the reason a `publish=true` dispatch cannot masquerade.

Applied fix: `step_verify` refuses a build receipt whose `workflow_run_id` differs from this run's (when known) and one whose `publish`/`pypi_only` flags differ from this dispatch's. `macos_asset` is deliberately not compared whole (the fixtures record it per receipt kind). Two tests pin both refusals and the matching case; both go red with the checks reverted.

scripts/release_receipts.py
```diff
@@ -46,6 +49,152 @@ class ReceiptError(Exception):
...
+        "requested": bool(raw.get("requested")),
+        "submitted": bool(raw.get("submitted")),
+        "stapled": bool(raw.get("stapled")),
+        "stapler_returncode": stapler_code,
+        "spctl_returncode": spctl_code,
+        "post_staple_sha256": str(raw.get("post_staple_sha256") or ""),
+    }
+
+
+def notary_succeeded(notary: Mapping[str, Any] | None) -> bool:
```
Nobody-Zhang, line 109: **[correctness] Two definitions of "this DMG is notarized" gate the same release.** This predicate demands `spctl_returncode == 0`; the sidecar producer (`describe_macos_image.py:202`, `"notarized": stapler.returncode == 0`, spctl documented as *informational* because it is a Gatekeeper policy answer that depends on the runner's state) and the pre-existing `read_signature`/`signing_state` gate (`release_pipeline.py:698-705`, never reads `spctl_rc`) derive it from the staple + post-staple digest. A stapled, digest-bound image with a non-zero assessment is `verified` by one gate and refused by this one with "no notarization success receipt". Every fixture sets spctl in lockstep with stapler, so nothing pinned either reading; no commit or doc justifies the stricter one.

Applied fix: one definition — staple + post-staple digest; `spctl` stays recorded. Pinned by `test_a_stapled_image_with_a_nonzero_spctl_assessment_is_still_notarized`.

scripts/release_pipeline.py
```diff
@@ -94,26 +94,34 @@
...
             "the posture requested on the build runner; not a measurement of the "
             "sandbox on an end user's machine"
         ),
-        # The platform boundaries this release did not prove, and why. Carried
-        # because a bundle that simply omits them is indistinguishable from one
-        # where every platform passed, and that is the reading a reader defaults
-        # to. Named here so "we could not check Linux" survives into the
-        # evidence rather than living only in a workflow comment.
-        "unproven": dict(release_gates.PLATFORM_CHECKS_UNAVAILABLE),
+        "linux_boundary": linux_boundary or release_gates.build_linux_boundary([], []),
```
Nobody-Zhang, line 123: **[correctness] The sealed evidence bundle asserts `ci_attestation.status: "missing"` whenever the test step did not run in this process.** `report(sealing=True)` looks for `linux_boundary` only in `self.results` for a step named `test`; `--only evidence` (any member of `STEPS` is accepted) then falls through to `build_linux_boundary([], [])`. Reproduced: `release_pipeline.py --only evidence` exits 0 and seals `sandbox.linux_boundary.ci_attestation.status = "missing"` for a candidate whose `ci-linux-sandbox-full` gate passed and sits in the quality receipt in `assets_dir`. The base bundle could only omit (`"unproven"`); this one asserts a wrong fact. The workflow's `--stop-after reverify` runs are unaffected.

Applied fix: when no test result exists in this run, the sealing report reads the boundary from the quality receipt `verify_quality_receipt` proves (the same derivation `step_test` uses); no receipt, no seal. Pinned by `test_a_partial_run_seals_the_receipts_boundary_not_an_empty_default`.

openai4s/storage/frames.py
```diff
@@ -90,6 +91,143 @@ def visible_session_clause(
     return clause, [user_id, user_id, user_id]
 
 
+PROJECT_PAGE_DEFAULT = 100
+PROJECT_PAGE_MAX = 100
+PROJECT_Q_MAX_CODEPOINTS = 128
+
+
+def _like_contains(value: str) -> str:
```
Nobody-Zhang: **[reuse → correctness] `_like_contains` is a byte-identical third copy (`storage/artifacts.py:122`, `datapro_index.py:263 _like_pattern`), and only the DataPro copy is paired with a NUL escape.** SQLite `LIKE` stops at an embedded NUL, so `GET /projects?q=lab%00zzzz` matched every project named `lab` — a filter silently widened rather than applied (verified against a real Store: `list_projects(q="lab\x00zzzz")` returned the `lab` rows). The filename search in `artifacts.py` had the same hole.

Applied fix: one `storage/like.py::like_contains` (escapes `\` `%` `_`, maps NUL to a private-use stand-in so the rest of the needle stays in force) used by frames and artifacts; the `/projects` route refuses a NUL in `q` with `400 invalid_q`. Pinned at both layers.

openai4s/storage/frames.py
```diff
@@ -339,24 +477,73 @@ def delete_project(self, project_id: str) -> dict:
...
+            )
+        else:
+            order = " ORDER BY p.updated_at DESC"
+        sql = (
+            "SELECT p.*, COALESCE(a.conversation_count, 0) AS conversation_count, "
+            "COALESCE(a.last_active_at, p.updated_at) AS last_active_at, "
+            "a.last_active_at AS last_active_raw " + _ACTIVITY_JOIN + where + order
+        )
+        bind: list[Any] = list(params)
+        if limit is not None:
```
Nobody-Zhang: **[correctness] `OFFSET` is emitted only under `if limit is not None`, so the facade's `offset` without `limit` switches to keyset order and returns the whole table.** `paging` (line 501) is switched on by `offset`/`before` alone; `Store.list_projects` forwards every keyword unchecked. Reproduced on a 5-project store: `store.list_projects(offset=3)` → 5 rows. Unreachable through `GET /projects` today (`project_listing` always pairs offset with a resolved limit), but the repository contract admits the combination and answers it wrongly rather than raising.

Applied fix: `LIMIT ?` with SQLite's unbounded `-1` when only `offset` is given, so `OFFSET` is always honoured. Pinned by `test_the_repository_honours_offset_without_limit`.

openai4s/agent/progress_circuit.py
```diff
@@ -0,0 +1,455 @@
...
+import json
+import re
+from dataclasses import dataclass
+from difflib import SequenceMatcher
+from typing import Any, Mapping, Protocol, Sequence
+
+from .actions import Action, CodeCell, FinalizeAction, NativeToolBatch
+from .models import ExecutionOutcome, ModelReply, RunState
+
+NO_PROGRESS_STOP_REASON = "no_progress"
```
Nobody-Zhang, line 24: **[correctness] The new `no_progress` stop reason was hand-added to four stop-reason lists but not the fifth: telemetry's `_OUTCOME` (`openai4s/telemetry/emit.py:95`).** `max_turns` maps to `timeout`; `no_progress` has no row and takes the unknown → `error` fallback (`turn_outcome("no_progress") == "error"`, verified), so every circuit trip is counted in consented telemetry as an error — the bucket the table's own docstring reserves for "a failure it has not seen", and the exact drift (`submitted`/`plan`) that docstring records having inverted the metric once already. Four review angles converged on this line.

Applied fix: `no_progress → timeout` (the bounded-stop sibling of `max_turns`, which every other projection treats it as), plus a test that reads `NO_PROGRESS_STOP_REASON` from its owner so a rename fails there rather than drifting. A comment here also records that this token is an Engine stop reason, distinct from Auto Mode's `no_progress_turn_limit` budget field that the harness fixture's `loop_kind` mirrors.

docs/release-validation.md
```diff
@@ -234,12 +234,27 @@ therefore attests only the private-PID interrupt and persistence contract. On
...
+not a second status on the same fact. A receipt that flattens both into one
+`passed`/`unproven` field is refused.
+
+The container image runs Python 3.14. That series is in the full offline CI
+matrix and in the quality-receipt check-suite gates. Container smoke is not a
+substitute for that matrix.
+
+A candidate SHA is not publishable on the strength of a `publish=true`
+dispatch. Schema-2 build receipts and the stage attestation record the
+candidate commit, the workflow run id, the dispatch inputs, check-run ids,
```
Nobody-Zhang: **[docs-accuracy] "Schema-2 build receipts and the stage attestation record … check-run ids" — build receipts never receive them.** All four `release_receipts.py --kind <build>` invocations in `release.yml` (303, 453, 517, 582) pass only `--source-sha` and `--workflow-run-id`; `--check-runs-json` is passed nowhere, so `build_build_receipt(check_runs=())` writes `check_runs: []` on every build receipt. Only the stage attestation carries them (copied from the quality receipt). Same family, fixed in the same pass: `docs/webapp-api.md`'s `POST /skills/import` row still described the pre-`seed` behaviour ("writes a normalized `user` document") while the head keeps the pasted frontmatter and adds `400 skill_capability_invalid` + a `review` object absent from the closed code table; and the `frame_update` / `step_update` vocabularies omit the new `no_progress` code that `gateway.py:9862` emits.

Applied fix: the three passages now describe what the code does.

openai4s/agent/ledger.py
```diff
@@ -842,6 +846,69 @@ def restore_action_history(
...
+
+    from .progress_circuit import reconstruct_progress_circuit
+
+    list_groups = getattr(store, "list_action_groups", None)
+    if callable(list_groups):
+        try:
+            outline = list_groups(
+                root_frame_id,
+                branch_id=branch_id or root_frame_id,
+                include_events=False,
```
Nobody-Zhang: **[efficiency] The "outline" pass is still `SELECT *` over every action group on the branch — `wire_state`, `assistant_message`, `usage` blobs included — under the Store lock, on every Web message, for an epoch that is empty by construction.** `include_events=False` only skips the events sub-query (`storage/actions.py:539`, no column projection; `_normalize_group` decodes three JSON columns per row). At the only production call site (`gateway.py:11013`, inside `_loop`, reached after `append_user` at 9719) the fresh user row is the branch's last group, so the scan rebuilds `ProgressCircuit()` every time. Measured 7.0 ms per message on a 601-group branch with 3 KB blobs (2.7 ms of it in-lock), linear in session length.

Applied fix: two index probes — `latest_action_group_ordinal(kind="user")` and the branch's latest ordinal (both walk `ux_action_group_ordinal` backwards, ~0.008 ms) — and only a non-empty epoch is read with events. Repository + facade method added surgically; the fork-aware `branch_action_groups` fallback is unchanged. Pinned by a test that asserts zero `list_action_groups` reads for an empty epoch.

openai4s/sdk/compute.py
```diff
@@ -10,6 +10,20 @@
 # for from inside the kernel worker.
 from openai4s.compute.states import TERMINAL_STATES as _TERMINAL_STATUSES
 
+# Kinds a failed cancel may carry. The host's `_compute_guard` currently
+# forwards `error` + `error_kind` only, so this namespace infers
+# `indeterminate` from the kind when the wire omits the flag. A caller that
+# treats any exception as "the job is gone" is the failure this exists to
+# prevent.
+_INDETERMINATE_CANCEL_KINDS = frozenset(
```
Nobody-Zhang: **[reuse] The closed cancel-reason set is spelled three times (`manager._INDETERMINATE_KINDS`, `manager.CANCEL_INDETERMINATE_REASONS`, this `_INDETERMINATE_CANCEL_KINDS`) plus the gateway's own `else "receipt_unconfirmed"`, and the comment above justifies the copy by a limitation this PR removed** (`host_dispatch._compute_guard` now forwards `indeterminate` on every `ComputeError`, so the kind lookup only fires for replies without the flag). In `manager.cancel` the inner `try: _terminate_sandbox … except: raise _cancel_failure(error)` is immediately re-mapped by the outer `except` — the one input where it matters is a provider `not_found` on terminate, which the outer `not_found` short-circuit would otherwise re-raise as "no such job" for a job that exists and may be billing; no test pinned that.

Applied fix: `CANCEL_INDETERMINATE_REASONS` / `INDETERMINATE_KINDS` / `REASON_CANCEL_UNCONFIRMED` live once in `compute/states.py` (the kernel-side module the SDK already imports); manager keeps its names as bindings; the double mapping collapses to one unconditional `_cancel_failure` (idempotent for in-set kinds) and the corrected comment explains what the SDK fallback still covers. `test_a_provider_not_found_on_terminate_is_an_unconfirmed_cancel` guards the collapse. Same pass: `auto_budget._canonical_arguments` now aliases `progress_circuit.canonical_arguments` (verbatim copy; no pinned digest bytes change).

tests/test_diagnostics_routes.py
```diff
@@ -0,0 +1,682 @@
...
+                    len(_payload(raw)),
+                    headers.get("content-disposition", ""),
+                )
+            )
+        assert len(set(seen)) == 1
+    finally:
+        node.close()
+
+
+def test_member_still_gets_403_while_an_admin_bundle_is_in_flight(
```
Nobody-Zhang, line 263: **[conventions] This test stubs `write_bundle_file` and drives `POST /diagnostics/bundle` over a real daemon without `@pytest.mark.stubbed_backend`; its three siblings stubbing the same function are marked.** CLAUDE.md: "A test that stubs a service must be marked `stubbed_backend` … without it a stub's fabricated shape gets published as the route's contract." The admin thread's 200 here is built from the stub (`PK\x03\x04empty`), and the suite-side recorder wraps `_send_static_bytes` precisely so diagnostics responses are frozen.

Applied fix: the marker.

openai4s/agent/progress_circuit.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `observe_native_batch` now carries a `same_result_digest` beside the call fingerprint: the same call counts toward the streak only when the answer did not move (no number collapsing — "10%" → "55%" is the delta a poll exists to observe). A group with no result rows keeps counting on the call alone. The ledger persists result content, so the reconstruct path applies the identical rule. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/skills_loader/frontmatter_edit.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `_key_of` folds the key the way `_parse_frontmatter` folds it, so every spelling of an owned field is replaced rather than duplicated. Found alongside it and fixed in the same pass: a multi-line owned value was written back as a bare `key: value`, so a `description: |` block lost every line but the first and its second line became a real top-level field. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/server/project_listing.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `_int_param` now checks an ASCII `-?[0-9]{1,19}` form and an explicit maximum. The length bound is load-bearing twice: it keeps the value inside SQLite's INTEGER, and it keeps CPython's own 4300-digit `int()` limit from raising a `ValueError` the route would have answered as another 500. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

frontend/src/features/sessions/dashboard.ts
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). Retry re-issues `searchProjectsNow()` (the exact request the box would make) and the rename path loads with the box's query when the dashboard is visible. `showWorkspace()` also cancels a pending debounce and reloads the unfiltered directory the header and switcher read — restoring the previous list if that background reload fails, since an emptied switcher is the same fallback-name symptom. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

frontend/src/features/sessions/load.ts
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). A `_projectsReplaceInFlight` flag now spans the life of a replace and is read by both gates, so a Load-more cannot start while a search is pending. Pinned by `load.replace.test.ts`, which fails with the gate reverted. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

scripts/release_pipeline.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `step_verify` now refuses a build receipt whose `workflow_run_id` differs from this run's and one whose `publish`/`pypi_only` differ from this dispatch's; `verify_build_receipts` requires the recorded inputs to be present, so a receipt cannot escape the comparison by omitting them. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

scripts/release_receipts.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). One definition of notarized: the staple plus the post-staple digest, matching `describe_macos_image.py` and `signing_state`. `spctl` stays recorded as the assessment a first launch performs. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

scripts/release_pipeline.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). When no test result exists in this run, the sealing report reads the boundary from the quality receipt that step verifies; no receipt, no seal. A `--mode local` run whose own test step attested nothing keeps the honest `missing`, and a moved tag is now reported as a moved tag rather than as an unreadable receipt. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/storage/frames.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). One `storage/like.py::like_contains` for the project and artifact searches, with a private-use stand-in for NUL so the rest of the needle stays in force; the route refuses a NUL in `q` outright. The DataPro index keeps its own copy on purpose — its escape is reversible and applied to the stored text as well. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/storage/frames.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `LIMIT ?` is emitted whenever either bound is given, binding SQLite's unbounded `-1` when only `offset` is set, so the offset is always honoured. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/agent/progress_circuit.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `no_progress` maps to `timeout` beside `max_turns`, and a test reads the constant from its owner so a rename fails there rather than drifting back to `error`. A comment here now records that this token is an Engine stop reason, distinct from the Auto Mode budget field the harness fixture's `loop_kind` mirrors. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

docs/release-validation.md
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). The three passages now describe what the code does: build receipts carry no check-run ids (only the stage attestation does), the import row describes the preserved frontmatter and the `review` object, `skill_capability_invalid` is in the failure table, and `no_progress` is in both WebSocket vocabularies. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/agent/ledger.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). Two index probes (`Store.latest_action_group_ordinal`) find the epoch anchor and say whether anything follows it; only a non-empty epoch is read with events. A test asserts zero `list_action_groups` reads for the empty case. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

openai4s/sdk/compute.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). `CANCEL_INDETERMINATE_REASONS`, `INDETERMINATE_KINDS` and `REASON_CANCEL_UNCONFIRMED` live once in `compute/states.py`; the manager keeps its names as bindings and the double mapping collapses to one unconditional `_cancel_failure`. `test_a_provider_not_found_on_terminate_is_an_unconfirmed_cancel` guards the case the inner wrap was silently protecting. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

tests/test_diagnostics_routes.py
Nobody-Zhang: Fixed in 9482dd63 on `next` (subject: "fix: second review pass for the next→main release (PR #144)"). The marker is on, so the recorder pauses for this stub the way it does for the three siblings. Verified at that commit: the full offline suite is green (8254 passed, 23 skipped), the frozen-shapes gate reports 211/211 routes with no breaking change, and each fix was mutation-checked — reverting it turns its test red.

PR #145, opened 2026-09-03
fix(compaction): stable long-context runs — robust summaries, rolling chunks, kernel-readable outputs, and compaction on the Action Ledger

Nobody-Zhang:
## Summary

Long-context runs were unstable: the model forgot earlier work and compaction frequently failed. Real data from a production data dir showed why — `compaction_archives` had 0 rows, 0 archive JSONs existed, and only 2 externalize blobs — compaction was effectively not landing. Seven interacting defects were diagnosed and fixed in six isolated lanes across three waves, then integrated and reviewed here.

**Lane A — summary robustness** (`compaction.py`): the summary call gets a real output budget (`max(8192, cfg.llm.max_tokens)`, `OPENAI4S_COMPACTION_SUMMARY_MAX_TOKENS` override) instead of 1024, no `temperature`; an empty or truncated summary raises `CompactionSummaryError` rather than archiving the middle behind a `"- No reliable summary was produced."` placeholder the model then believed; the summarizer input drops `wire_state` and caps tool arguments.

**Lane B — CLI context budget** (`loop.py`): the CLI `CompactionPolicy` gets the same provider-derived `context_budget_provider` and `tool_schema_provider` the Web loop already had, so `openai4s run` compacts against the model's real window rather than 262144.

**Lane C — failure breaker + calibration** (`runtime.py`): two consecutive `compact()` failures open a circuit that reopens at 1.5× growth; the previous reply's real `input_tokens` calibrates the estimate (clamped to [0.5, 8]).

**Lane D — rolling chunk summaries** (`compaction.py`): a long middle is summarized as bounded rolling chunks (each `chat()` input under 0.35× budget, the previous handoff carried forward and excluded from the transcript), the retained tail is chosen by tokens without splitting an atomic pair, and CJK characters estimate as one token each.

**Lane E — kernel-readable large outputs** (`compaction.py`): an oversized output is copied to a workspace-relative `.openai4s/context/<sha256>.json` the kernel can `json.load`; the marker names only that relative path.

**Lane F — compaction on the Action Ledger** (`ledger.py`, `gateway.py`): compaction is a `kind="compaction"` group with `assistant_message=None` (old reducers skip it; zero migrations), so a daemon restart restores the compacted view instead of the full history.

**Integration seams** found by hand after each clean auto-merge: `compact()` now threads `workspace` into `externalize_large_outputs` (Wave 2), and the ledger-restored note carries the same `COMPACTION_NOTE_PREFIX` framing as the in-memory one (Wave 3).

**Wave 4 review** (`/code-review high`, 8 finder angles → 13 verifiers, 11 CONFIRMED with real-Store reproductions) — fixed in the last commit:
- The durable ledger record was written from inside `compact()` before the policy decided; a rejected compaction was recorded as applied and the next restore dropped the live tail. The sink is now buffered and replayed only for an adopted projection.
- Coverage was keyed by ordinal, which restarts at 0 on a forked branch behind the inherited parent prefix — a compaction deleted the child's own tail on restore. Coverage is now the last fully-covered `group_id` in the branch's spliced sequence, and any live/ledger misalignment covers nothing rather than guessing toward the tail.
- `append_compaction` was two autocommits; a group without its event restored as an empty system note that erased the previous handoff. It is one transaction now, and an orphan group is skipped.
- Summary `max_tokens` ≥ 8192 was rejected by `validate_model_request` on any model capped lower (e.g. a 4096 self-hosted overlay), which tripped the breaker and killed compaction for the run. It is clamped to the declared cap.
- Truncation was recognised only as `finish_reason == "length"`; Anthropic `max_tokens`, Gemini `MAX_TOKENS`, and the Responses wire's `provider_finish_reason == "incomplete"` were accepted and placeholder-padded. All are truncation now, with one retry at double budget.
- A zero `context_budget` (an entry whose max_output equals its window) triggered compaction every turn with a zero chunk budget; it now means unknown.
- A poisoned workspace `.openai4s/context` (regular file, chmod 000, symlink out) failed every compaction; the kernel copy is best-effort now.
- The Web `workspace_provider` was dead code (the Artifact branch never wrote the copy, so R cells had no read-back); the copy is written on every path, keyed on the kernel cwd (`st.workspace`).
- A reopened breaker kept `failure_streak` at its trip value and re-tripped on one failure.
- The `should_compact` identity shim (production branching on a test monkeypatch) is gone.
- `tests/README{,_zh}.md` list the nine new test files (the directory-README gate was red).

## Area

- [x] Server / gateway / API
- [x] Tests / harness / CI

## Change Type

- [x] Bug fix

## Verification

Ran (at the final commit):

```text
uv run pytest -n auto --maxprocesses=4 --dist loadfile   # 8144 passed, 24 skipped, 0 failed (12:03)
uv run pre-commit run --all-files                          # all hooks passed, no file modified
uv run mypy                                                # Success: no issues found in 8 source files
uv run python scripts/source_secret_scan.py                # passed (3749 files)
uv run python3 -m harness.cli run --tier pr --offline      # 38/38 passed, failed=0
uv run python scripts/capture_response_contract.py --check # contract up to date: 207/207 routes
uv run python scripts/capture_response_schemas.py --check  # no breaking change, 1136 shapes, 207/207 routes (pre-existing additive drift on GET /orchestration/jobs/…, not regenerated)
uv run python scripts/check_directory_readmes.py           # only the untracked local docs/.vitepress/ remains
```

Not run:

```text
node tests/browser_*.mjs, container_smoke.sh, linux_bwrap_* smokes
```

Reason:

```text
No web UI, container, or sandbox code changed; the Web path was exercised through
SessionRunner + a real Store in tests/test_compaction_ledger.py and
tests/test_compaction_wave4_review.py (fork, restart rebuild via _seed_messages).
```

## Risk and Reviewer Focus

- `openai4s/agent/ledger.py` — `_reduce_action_groups_annotated` / `_apply_compaction` / `compaction_cover_group_id`: positions are indexes into `branch_action_groups`' spliced sequence. Check the head rule (first `role=user`) and that a cover id pointing at or after the compaction group resolves to "nothing covered".
- `openai4s/agent/runtime.py` — the deferred `archive_sink` replay in `_prepare`: a sink failure returns the uncompacted messages and counts as a compaction failure.
- `openai4s/server/gateway.py` — three surgical lines in `_loop`'s `CompactionPolicy` construction plus the `_archive_compaction_record` signature; `workspace_provider` now reads `st.workspace` (the kernel cwd, which placement may repoint).
- Deliberately not changed (product calls for a follow-up): the Web loop builds a fresh `CompactionPolicy` per user message, so the breaker and calibration reset each turn (pre-existing lifetime; base had the same for the low-yield breaker); `compact()`'s internal chunk/tail budgets use the raw estimate rather than the calibrated one.

### Diff Readiness

- [x] Hotspot files were edited surgically if touched: `gateway.py`, `host_dispatch.py`, `store.py`, `worker.py`, `manager.py`.
- [x] Workbench changes were made in `frontend/` (not legacy `app.js`) and `openai4s/server/webui/dist/` was rebuilt and committed with the source. (no workbench change)

### Tests

- [x] Kernel / host-RPC / gateway changes include focused contract coverage or a documented smoke test.

### Core Dependency Policy

- [x] Optional science imports are guarded by `try/except ImportError` at every in-tree use site.
- [x] New dependencies, if any, are documented and justified. (none)

- [x] Security-sensitive changes include appropriate synthetic tests or manual verification notes. (workspace blob path confinement: `tests/test_compaction_externalize_retrieval.py`)

### Docs

- [x] `README.md` and `README_zh.md` stay in sync where translated sections are affected.

---

> [!NOTE]
> **Medium Risk**
> Changes core context compaction, ledger replay, and workspace archive I/O; mis-coverage or premature archive recording could drop tail history on restore, though Wave 4 tests target those seams.
> 
> **Overview**
> Fixes long-context runs where compaction rarely landed and summaries could replace real history with placeholders the model trusted.
> 
> **Compaction engine** (`compaction.py`) now uses **CJK-aware token estimates**, **token-based tails** (`keep_recent_by_tokens`) that still honor atomic segments, and **rolling chunk summaries** with chained handoffs. Summary calls get a real **`max_tokens` budget** (env/config, clamped to model caps), **no temperature**, and **fail closed** via `CompactionSummaryError` on empty or truncated handoffs; summarizer input drops **`wire_state`** and truncates tool args. **Large outputs** can be written to workspace-relative **`.openai4s/context/<sha256>.json`** with digest-verified exclusive writes and read-back hints (Artifact path included).
> 
> **Runtime policy** (`runtime.py`) triggers on **calibrated** context estimates from prior **`input_tokens`**, adds a **consecutive-failure circuit breaker** alongside low-yield breaking, and **defers `archive_sink`** until a compaction is actually adopted. **`workspace_provider`** threads kernel cwd into externalize/compact (best-effort if writes fail).
> 
> **Action Ledger** (`ledger.py`, `gateway.py`) records **`kind=compaction`** groups (transactional with events) so restarts **restore the compacted view** via `covered_through_group_id` on the branch’s spliced group sequence; restored notes share **`COMPACTION_NOTE_PREFIX`** with in-memory `compact()`.
> 
> **CLI** (`loop.py`) aligns with Web: provider **context budget**, **tool schemas**, and **workspace** on `CompactionPolicy`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26a600889e9b62fdf27a2b52fd8f511e926a65f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openai4s/agent/compaction.py
```diff
@@ -439,14 +596,59 @@ def _write_content_blob(
...
+
+def _write_workspace_content_blob(
+    workspace: Path | str,
+    digest: str,
+    payload: Mapping[str, Any],
+) -> str:
+    rel = _workspace_context_ref(digest)
+    path = _confined_workspace_blob_path(Path(workspace), digest)
+    _write_exclusive_json(path, payload)
+    return rel
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 26a60088: `_write_exclusive_json` no longer takes `FileExistsError` as a deduplicated write on its own — a pre-existing file at the digest path is verified the way `load_archived_content` verifies one (its `sha256` field and the digest recomputed from its `content` must both match), and a mismatch is a failed workspace write (archive-only marker, no `workspace_ref`) rather than an endorsement. The daemon-side archive blob gets the same check. Regression test `test_a_planted_file_at_the_digest_path_is_not_endorsed` plants poison at the exact path and asserts it is neither named nor rewritten, while an identical pre-existing file still dedups. A later overwrite of an already-endorsed file remains possible by design of an agent-writable workspace; the marker carries the honest sha256 for a reader to check.

Nobody-Zhang:
cursor review

openai4s/agent/compaction.py
```diff
@@ -655,34 +1069,51 @@ def compact(
...
+                {
+                    "role": "user",
+                    "content": _summary_user_content(batch, metadata, prev),
+                },
+            ],
+            cfg,
+            max_tokens,
+            cap,
+        )
+        prev = raw_summary
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 89efe7e6: segments are now priced as the serialized transcript text the model receives (JSON wrapping included) and batched on the fly against `chunk_budget - tokens(PREVIOUS_HANDOFF_PREFIX + prev)`, with a floor of a sixth of the chunk budget so an outsized handoff shortens chunks rather than zeroing them. Regression test `test_chunk_budget_leaves_room_for_the_previous_handoff` carries a 0.15×-budget handoff and asserts every request stays under 0.35× budget; with the reserve removed it reports a 16.4k-token request against a 14k cap.

openai4s/agent/compaction.py
```diff
@@ -439,14 +604,80 @@ def _write_content_blob(
...
+        raise ValueError("existing archive blob does not match its digest")
+    canonical = _json_text(existing.get("content")).encode("utf-8")
+    if hashlib.sha256(canonical).hexdigest() != digest:
+        raise ValueError("existing archive blob does not match its digest")
+
+
+def _write_exclusive_json(path: Path, payload: Mapping[str, Any]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
     try:
         with path.open("x", encoding="utf-8") as handle:
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness]** `_write_exclusive_json` writes straight into the content-addressed path (`open("x")` + streaming `json.dump`). A crash or ENOSPC mid-dump leaves a truncated `<sha>.json`; every later externalization of that content then raises `ValueError("existing archive blob is unreadable")` from `_verify_existing_blob` — forever, nothing repairs it. On the host archive path that error is caught by `_prepare` (output kept inline, only `workspace` neutralised), but `compact()` is still handed `archive_dir=cfg.compaction_dir`, re-externalizes, raises again, and `_compaction_failed` counts it: reproduced turn 1 `failure_streak=1`, turn 2 `circuit_open=True`, turn 3 skipped — with the summarizer never called, and a fresh run against the same data dir tripping identically. Two writers racing on one digest (delegate fan-out shares `compaction_dir`) hit the same path transiently.

Fix applied on the review branch: temp file in the same directory + `fsync` + `os.replace`, verify only a file that already existed; a torn file (not JSON) is replaced on both roots, a well-formed foreign blob is replaced only under the host-owned archive (the workspace planted-file contract is kept). Regression tests in `test_compaction_externalize_retrieval.py` and `test_compaction_failure_breaker.py` (red at 89efe7e6).

Nobody-Zhang:
**openai4s/server/session_package.py:3221** — **[xhigh review · CONFIRMED · correctness]** Session-package import remaps every `group_id` (`group_map`, ~3150/3205) but copies the compaction event's `result` verbatim (`result=event.get("result")`). The imported `covered_through_group_id` therefore names a group that does not exist on the imported branch; `_reduce_action_groups_annotated` resolves it to `-1`, `_apply_compaction` keeps every covered message *and* inserts the note claiming they were archived. Real-Store repro: source restores 4 messages after compaction, the import restores 8 (all covered observations back) plus the note; `archive_id` dangles too.

Fix applied: `_remap_compaction_result` in `_import_ledger` maps `covered_through_group_id` through `group_map` (raises `SessionPackageError` on an unknown non-`None` id, like the branch check) and nulls `archive_id` (the package carries no `compaction_archives` rows and the reducer never reads it). Tests added in `test_session_package.py`.

openai4s/agent/compaction.py
```diff
@@ -492,16 +726,34 @@ def externalize_large_outputs(
...
+        # ``.openai4s/context`` occupied, unreadable, or symlinked out, and
+        # losing the copy must not lose the compaction (the archive or the
+        # Artifact still holds the bytes).
+        workspace_ref: str | None = None
+        if workspace_root is not None:
+            try:
+                workspace_ref = _write_workspace_content_blob(
+                    workspace_root, digest, payload
+                )
+            except (OSError, ValueError):
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness]** `except (OSError, ValueError)` around the workspace copy misses the `RuntimeError("Symlink loop from …")` that `Path.resolve()` raises on CPython 3.10–3.12 (two of the three CI legs) when a cell leaves a self-referential symlink at `.openai4s/context` — exactly the "symlinked out" case the comment above anticipates. The exception escapes `externalize_large_outputs`; `_prepare`'s outer `except Exception` then keeps *every* oversized output inline for the turn and drops the workspace, so the whole externalization is lost, not just the kernel copy. Verified under 3.10.20 and 3.12.13 (3.13 raises OSError and is swallowed as designed).

Fix applied: catch `RuntimeError` there too (the repo already does this around `.resolve()` in `host/files.py`); regression test with an `os.symlink` loop in `test_compaction_externalize_retrieval.py`.

openai4s/agent/compaction.py
```diff
@@ -416,17 +581,17 @@ def _confined_blob_path(archive_dir: Path, digest: str) -> Path:
     return candidate
 
 
-def _write_content_blob(
-    archive_dir: Path,
+_WORKSPACE_CONTEXT_DIR = ".openai4s/context"
```
Nobody-Zhang, line 626: (re: line 571) **[xhigh review · CONFIRMED · correctness]** `_large_output_candidate` treats a user message as an output whenever the previous assistant message is a code fence (`index > 0 and _has_code_action(messages[index - 1])`). After a Stop-cancelled cell there is no observation between the code reply and the user's *next* prompt (`agent_run.py` returns `stop_reason="cancelled"` with no history message), and `_prepare` externalizes unconditionally before the trigger check — so a >16 KB pasted prompt is replaced by a preview marker the model never sees through, and a pinned figure (`[{"type":"image","data":…}]`, always >16 KB as JSON) is archived away from the vision model. The ledger restore inserts the synthetic cancelled observation, so the restored projection differs from the live one. Rule is pre-existing, but this PR rewrote the function around it and added the workspace copy on the same rule.

Fix applied: a user message is a candidate only when its content is a `str` starting with `[Observation]`/`[Tool Results]`/`[Tool result]` (every runtime-synthesized observation carries one of those; nudges are tiny); the code-action heuristic is dropped. 81 existing tests still pass; regression added.

openai4s/agent/runtime.py
```diff
@@ -231,13 +280,14 @@ def prepare(self, state: RunState) -> Sequence[Mapping[str, Any]]:
...
                 tool_schemas=tool_schemas,
+                context_budget=context_budget,
+                workspace=workspace,
             )
         except Exception as error:  # noqa: BLE001 - compaction cannot kill a run
-            state.metadata["last_compaction_error"] = str(error)[:500]
-            self.log(f"[compaction skipped] durable archive failed: {error}")
-            return messages
+            return self._compaction_failed(state, before, error, messages)
+        self.failure_streak = 0
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness]** `self.failure_streak = 0` runs immediately after `compact()` returns — *before* the deferred `archive_sink` replay a few lines down. A sink that keeps failing (ledger admission refused, store locked) therefore goes `_compaction_failed` → streak 1 → next turn compact() again → streak reset → sink fails → streak 1 … and the failure breaker can never open: probe shows six consecutive sink failures = six real summarization LLM calls with `circuit_open=False` throughout. That is precisely the repeated-futile-compaction the breaker exists to stop.

Fix applied: the reset moves to the adoption point (after the replay succeeds) and to the non-shrinking return; the sink-failure path keeps the live streak. Test `test_a_failing_archive_sink_trips_the_breaker` (red at 89efe7e6).

openai4s/agent/runtime.py
```diff
@@ -268,8 +318,96 @@ def prepare(self, state: RunState) -> Sequence[Mapping[str, Any]]:
...
+    def _calibration_ratio(self, state: RunState) -> float:
+        previous = _positive_float(state.metadata.get("context_estimate_calibration"))
+        if previous is None:
+            previous = 1.0
+        actual = 0.0
+        reply = state.last_reply
+        if reply is not None:
+            usage = reply.usage
+            if isinstance(usage, Mapping):
+                actual_value = _positive_float(usage.get("input_tokens"))
```
Nobody-Zhang, line 376: **[xhigh review · CONFIRMED mechanism · correctness]** `_calibration_ratio` reads the normalized `usage["input_tokens"]`, but on the Anthropic Messages wire that counter is the *uncached remainder only* — `normalize_usage` keeps `cache_read`/`cache_write` as separate keys and never folds them in (its docstring assumes providers include them, which OpenAI and Gemini do and Anthropic does not). With 93 % of a prompt cached the ratio collapses to the 0.5 floor, `calibrated_total` halves, and compaction fires only when the raw estimate exceeds 1.5× the window — after the provider has rejected the request. Dormant against api.anthropic.com today (the client never sends `cache_control`), live the moment any Anthropic-compatible endpoint auto-caches or `cache_control` is added. Note the naive `input + cache_read + cache_write` is *wrong* on OpenAI/Gemini (there `cache_read` is a subset of the input count).

Fix applied at the class level: `UsageMapping.input_excludes_cache=True` on `_ANTHROPIC_USAGE`, and `normalize_usage` folds the cache counters into `input_tokens` when set, so the canonical count means the whole prompt on every wire (this also fixes `calculate_usage_cost_usd`'s `regular_input` arithmetic on Anthropic). Frozen claude row in `test_llm_capabilities.py` updated (12→17, total 19→24); regression in `test_compaction_failure_breaker.py` built through `normalize_usage(..., "claude")`. Left as a product call: the 0.5 floor itself.

openai4s/agent/compaction.py
```diff
@@ -616,20 +901,155 @@ def _normalize_handoff(summary: str, metadata: CompactionArchiveMetadata) -> str
...
+            f"(finish_reason={finish_reason!r})"
+        )
+    return text
+
+
+def _summary_chunk(
+    request: list[dict[str, Any]], cfg: Config, max_tokens: int, cap: int | None
+) -> str:
+    """One chunk summary; a truncated reply gets one retry at double budget."""
+    reply = chat(request, cfg.llm, max_tokens=max_tokens)
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness]** Dropping `temperature=0.2` does not omit temperature at the wire: `providers/openai.py:35`, `anthropic.py:39`, `gemini.py:33` all build `"temperature": cfg.temperature if temperature is None else temperature`, and `LLMConfig.temperature` defaults to 0.7 and cannot be `None`. Wire probe through `_summary_chunk` → payload `temperature: 0.7`. So the summarizer now samples the handoff at the session's main-turn temperature, and the commit's stated intent ("omits temperature so reasoning endpoints that reject a non-default value keep working") is not achieved either — such an endpoint already receives 0.7 on every main turn. The four `assert kwargs.get("temperature") is None` tests certify the kwarg, not the wire.

Fix applied: explicit `_SUMMARY_TEMPERATURE = 0.2` on both summary calls; tests flipped to assert it. If real omission is wanted it needs a wire-level path (optional `LLMConfig.temperature`) shipped with a wire-level test.

openai4s/agent/compaction.py
```diff
@@ -616,20 +901,155 @@ def _normalize_handoff(summary: str, metadata: CompactionArchiveMetadata) -> str
...
+def _summary_max_tokens(cfg: Config, cap: int | None = None) -> int:
+    raw = (os.environ.get("OPENAI4S_COMPACTION_SUMMARY_MAX_TOKENS") or "").strip()
+    wanted = 0
+    if raw:
+        try:
+            wanted = int(raw)
+        except ValueError:
+            wanted = 0
+    if wanted <= 0:
+        wanted = max(8192, int(getattr(cfg.llm, "max_tokens", 0) or 0))
```
Nobody-Zhang, line 1290: **[xhigh review · CONFIRMED · correctness]** `max(8192, cfg.llm.max_tokens)` is clamped only when the model declares an output cap, and `get_provider_capabilities` deliberately sets `max_output_tokens=None` for every loopback *and private/LAN* endpoint (`_is_local_endpoint`), so a self-hosted deployment gets `max_tokens=8192` unclamped (the retry doubles it to 16384). vLLM/SGLang validate `prompt + max_tokens ≤ max_model_len` and answer 400, which is not retried; end-to-end with `OPENAI4S_CONTEXT_WINDOW=8192` and a fake vLLM: two turns of `HTTP 400 … requested 10761 tokens` then the breaker opens; the same run with 1024 compacts 23→8 messages. Bites exactly the operator who set the true window (<~12k); the default 262144 fallback never reaches it. `docs/configuration.md`'s own self-hosted recipe (`context_window_tokens=16_384`, no cap) is one step from it.

Fix applied: when the cap is unknown, bound the request by the room the window leaves after the chunk (`budget - chunk_budget - system - header`), used for the retry too; a declared cap is untouched, so the 8192 floor survives everywhere it fits. `OPENAI4S_COMPACTION_SUMMARY_MAX_TOKENS` is now documented.

openai4s/agent/loop.py
```diff
@@ -720,7 +721,16 @@ def bootstrap(kernel: Any) -> None:
...
-                        self.context_policy or CompactionPolicy(self.cfg, log=self._log)
+                        self.context_policy
+                        or CompactionPolicy(
+                            self.cfg,
+                            log=self._log,
+                            context_budget_provider=_child_context_budget(self.cfg),
+                            tool_schema_provider=lambda state: with_finalize_response(
+                                tool_catalog.specs_for(state.messages)
+                            ),
+                            workspace_provider=lambda _s: run_cwd,
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness / partial wiring]** These providers reach only the CLI *parent*: `_SteeringContextPolicy` (`delegation.py:1142`) builds a bare `CompactionPolicy(cfg, context_budget_provider=…)` and passes it as `context_policy=`, so the `or` here never runs for a delegated child. Real `DelegationRunner` child e2e: a 21 k-char native `read_text_file` result becomes `[Large output archived] … archive_ref: blobs/…` (a compaction-dir path no cell can open) with no read-back and no `ws/.openai4s/context` copy, while the parent on the same scenario gets `workspace_ref` + the read-back line; children also price their trigger with an empty schema tuple (4 253 schema tokens on the probe catalogue). The finder's stronger claim ("no marker at all") is refuted — `_prepare` always passes `cfg.compaction_dir` — but the kernel-readable half of Lane E and the CLI-budget half of Lane B skip the fan-out, which is where the longest runs happen.

Fix applied: `_SteeringContextPolicy.bind_runtime(**providers)` fills providers the base lacks (keeps the child's own budget), and `Agent.run` calls it on any supplied policy that offers it. Tests in `test_delegation_runtime.py`.

openai4s/agent/compaction.py
```diff
@@ -386,6 +429,128 @@ def safe_keep_recent(messages: list[dict], minimum: int = 4) -> int:
...
+    for index in range(len(messages) - 1, -1, -1):
+        cost = estimate_context([messages[index]]).total
+        if tokens + cost > token_budget:
+            break
+        tokens += cost
+        kept += 1
+    return safe_keep_recent(messages, kept)
+
+
+def _existing_handoff_text(messages: Sequence[Mapping[str, Any]]) -> str | None:
```
Nobody-Zhang, line 466: **[xhigh review · CONFIRMED · correctness]** `_existing_handoff_text` returns the previous note's full `content`, i.e. `COMPACTION_NOTE_PREFIX + handoff`, and `_summary_user_content` places it under "PREVIOUS HANDOFF (authoritative; carry every fact forward …)". A model that carries the banner forward (as instructed) returns text starting with it; `_normalize_handoff` passes it through when the titles are complete; `compact()` then builds `COMPACTION_NOTE_PREFIX + handoff` → banner twice in memory. The gateway records the bare handoff, and `_apply_compaction` prepends only `if not content.startswith(COMPACTION_NOTE_PREFIX)` → banner once on restore. Scripted: in-memory prefix count 2, restored 1, 128-byte delta — the exact byte-for-byte invariant the shared constant was introduced to hold. The existing second-compaction test never asserts the prefix is absent from `prev`.

Fix applied: strip the banner at the source (`removeprefix` in `_existing_handoff_text`) and make `compact()` idempotent like the ledger side; regression test drives `compact()` twice with an echoing fake and compares against `_apply_compaction`.

openai4s/agent/compaction.py
```diff
@@ -616,20 +901,155 @@ def _normalize_handoff(summary: str, metadata: CompactionArchiveMetadata) -> str
...
+        slim["arguments"] = _truncate_summary_arg(call["arguments"])
+    elif "arguments" in function_map:
+        slim["arguments"] = _truncate_summary_arg(function_map["arguments"])
+    if "raw_arguments" in call:
+        slim["raw_arguments"] = _truncate_summary_arg(call["raw_arguments"])
+    return slim
+
+
+def _summary_message(message: Mapping[str, Any]) -> dict[str, Any]:
+    slim = {key: message[key] for key in _SUMMARY_MESSAGE_KEYS if key in message}
```
Nobody-Zhang, line 1226: **[xhigh review · CONFIRMED · correctness]** `_summary_message` keeps `content` verbatim, so a pinned figure (`{"type":"image","data":<base64>,"mime":…}`, gateway ~9060) reaches the summarizer as base64 *text*. Measured for a 300 KB PNG: `estimate_context` prices the message at 1 102 tokens (the stored shape never contains the literals "base64"/"data:image" checked at :276), `_summary_pieces` prices the same segment at 75 199; `_next_summary_batch` ships it as a solo batch (68× the audit row's `summary_chunk_estimates` entry), and `chat()` sends a text transcript whose one string is the whole image — a normal turn would have sent an image block. 1 of 3 summarizer calls carried the full payload; at the sizes the gateway allows (4 MiB/image) that is ~1.4 M tokens, rejected, and the breaker opens after two turns.

Fix applied: `_summary_content` replaces image blocks with `{"type":"image","omitted":true,"chars":…,"mime":…}` and keeps text parts (dry run: 300 629 → 501 chars, image message joins a 41-message batch). Left as a product call: aligning `_content_estimate`'s image detection with the stored shape, since that moves the trigger.

openai4s/agent/compaction.py
```diff
@@ -492,16 +726,34 @@ def externalize_large_outputs(
...
+        workspace_ref: str | None = None
+        if workspace_root is not None:
+            try:
+                workspace_ref = _write_workspace_content_blob(
+                    workspace_root, digest, payload
+                )
+            except (OSError, ValueError):
+                workspace_ref = None
+        read_back = (
+            f"json.load(open({workspace_ref!r}))['content']"
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness of the instruction]** The read-back the marker hands the model is Python-only and cwd-relative, yet it is labelled "from any cell" and attached to R observations too (`format_observation` prefixes R cells with `[Observation]` and `_large_output_candidate` matches on that; `r_worker.R` has no `host` object). Run verbatim in a real R worker: `could not find function "json.load"`, `exists(host) → FALSE`. In a real Python kernel the snippet raises `NameError: name 'json' is not defined` (the worker namespace has no pre-imports — the PR's own test prepends `import json` that the marker omits), and after any `os.chdir` the relative path is `FileNotFoundError`. `OPENAI4S_WORKSPACE` is exported into both Python and R workers' env (`kernel/environment.py:216-217`), and env-anchored reads succeed in both.

Fix applied: one language-neutral hint ("JSON at '<ref>' under the kernel workspace ($OPENAI4S_WORKSPACE), key 'content' — Python: import json, os; json.load(open(os.path.join(os.environ['OPENAI4S_WORKSPACE'], ref)))['content']; R: jsonlite::fromJSON(file.path(Sys.getenv('OPENAI4S_WORKSPACE'), ref))$content"), `host.artifact_path` kept Python-scoped, no absolute path printed. The kernel test now runs the marker's own snippet verbatim after an `os.chdir('/')`.

openai4s/agent/runtime.py
```diff
@@ -208,20 +245,32 @@ def prepare(self, state: RunState) -> Sequence[Mapping[str, Any]]:
             # retry.  Without this the breaker permanently disables compaction
             # for the run and the context grows unbounded into a provider 4xx.
             if before.total < self.circuit_open_total * self.circuit_retry_growth:
-                self.log(
-                    "[compaction skipped] circuit breaker open after "
+                reason = (
```
Nobody-Zhang: **[xhigh review · CONFIRMED · correctness (diagnostics)]** The skip log guesses which breaker tripped from `if self.low_yield_streak else …`, but `low_yield_streak` is reset only by a high-yield compaction or the reopen block — not by `_compaction_failed`. Probe: low-yield, fail, fail → trip log says "2 consecutive failures", every later skip says "circuit breaker open after 1 low-yield attempts" (a count that cannot open it), and `state.metadata` carries `compaction_circuit_open` with no reason.

Fix applied: `circuit_reason` field set by one `_trip(reason, before)` helper at both trip sites, cleared on reopen, reported in the skip log and as `compaction_circuit_reason`.

openai4s/agent/compaction.py
```diff
@@ -492,16 +726,34 @@ def externalize_large_outputs(
...
+        # Artifact still holds the bytes).
+        workspace_ref: str | None = None
+        if workspace_root is not None:
+            try:
+                workspace_ref = _write_workspace_content_blob(
+                    workspace_root, digest, payload
+                )
+            except (OSError, ValueError):
+                workspace_ref = None
+        read_back = (
```
Nobody-Zhang: **[xhigh review · CONFIRMED · altitude]** Lane E is unreachable for the common case it was written for. `format_observation` → `_budgeted` (`runtime.py`, `OBSERVATION_SECTION_BUDGET = 12_000`, head 6 000 + tail 4 000) already trims every stdout/stderr section and spills the rest to `.openai4s/observations/obs-<kind>-<16hex>.txt` with its own `open(ref).read()` marker, so a single-section observation is at most 12 022 chars — below the 16 384 externalize threshold. Driven end-to-end (`openai4s run` against a mock provider): a 20 016-char print produced only the observation spill; `.openai4s/context/` was never created. What *can* reach Lane E: two-section observations, an unbudgeted `ERROR:` section, `role=tool` results up to `output_limit`, legacy `[Tool Results]`, assistant prose. On the Web path `_archive_context_output` (gateway ~10604) already writes the same `content` as a `schema_version 1` Artifact under `.openai4s-context/<sha>.json`, so one oversized output can now sit in a workspace in three formats. No test chains `format_observation` → `externalize_large_outputs`.

Not fixed here (facade + frozen Artifact schema): unify on one workspace spill primitive shared by `_budgeted` and compaction (one directory, one payload shape, one read-back phrasing), route the `ERROR:` section through `_budgeted`, and have the Web archiver register the existing blob instead of writing a second file. Worth a follow-up PR.

openai4s/agent/compaction.py
```diff
@@ -201,8 +233,19 @@ def _json_text(value: Any) -> str:
...
+
+    Empty input stays 0.  Any non-empty string is at least 1.  ASCII rounding
+    matches the historical ``(len + 3) // 4`` ceiling so 4000 Latin characters
+    still estimate as 1000 tokens.
+    """
+    if not value:
+        return 0
+    if value.isascii():
+        return max(1, (len(value) + 3) // 4)
+    cjk = len(_CJK_CHAR_RE.findall(value))
```
Nobody-Zhang: **[xhigh review · efficiency + cleanup, applied]** `len(_CJK_CHAR_RE.findall(value))` materialises one 1-char `str` per ideograph just to count them, on every `estimate_context` pass — and `prepare()` made three full-history passes per LLM call (`before`, `after`, then a third for `context_estimate_sent` that `_prepare` had already computed). Measured on 1.8 MB of Chinese: 84.5 ms per pass single-char vs 3.9 ms run-based. Applied: run-based `_CJK_RUN_RE` (also covering Hangul jamo, compatibility ideographs and the supplementary planes, which estimated at 4:1), `_prepare` returns its estimate so `prepare()` reuses it, `context_estimate` written only at adoption (no set-then-reset), the runtime header priced once per request instead of once per segment (~30 phantom tokens × segments against the chunk budget), the reserve and the request body built from one preamble, the three marker branches from one builder (the archive-only baseline text is byte-identical), one `_confine` for both blob roots, `_verify_blob_payload` shared by the read and dedup-write paths, `_truncate_summary_arg` keeps the marker inside the limit and renders with `_json_text`, `_archive` derives `summary_chunks` from the estimates, `_require_usable_summary` takes the reply mapping, dead `_should_trigger` params removed, gateway's `context_budget_provider` looks capabilities up once.

openai4s/agent/compaction.py
```diff
@@ -439,14 +604,80 @@ def _write_content_blob(
...
+
+    The workspace copy lives where a cell can write, so a file already at
+    ``<sha256>.json`` cannot be trusted for its name: a Skill could plant
+    different bytes there and the marker would then endorse them as the host
+    archive.  The digest is recomputed from the file's ``content`` the same
+    way ``load_archived_content`` checks it.
+    """
+    digest = str(payload.get("sha256") or "")
+    try:
+        existing = json.loads(path.read_text("utf-8"))
```
Nobody-Zhang: **[xhigh review · sweep · CONFIRMED · correctness/availability]** The new pre-existing-blob verification reads the file at an agent-writable, agent-enumerable path with a plain `path.read_text()`. A cell can `os.remove` + `os.mkfifo` every entry of `.openai4s/context/` (no digest prediction needed); the next externalization of the same content (dedup is the whole point, and any ledger restore re-externalizes) hits `FileExistsError` → `_verify_existing_blob` → `read_text` on the FIFO blocks forever with the Web turn lock held / the CLI process hung — no `OSError` ever fires, so the best-effort guard never runs. Probe: `externalize_large_outputs` did not return within 3 s. A symlink there would read whatever it points at; a huge regular file is swallowed whole.

Fix applied: `lstat` first (a non-regular object is a torn blob to replace, not something to read), size bounded by `2× expected + 1 MiB`, then an `O_NONBLOCK|O_NOFOLLOW` bounded read. Tests: a planted FIFO returns promptly and is replaced by the blob; a symlink to a host file is neither followed nor endorsed.

openai4s/agent/compaction.py
```diff
@@ -616,20 +901,155 @@ def _normalize_handoff(summary: str, metadata: CompactionArchiveMetadata) -> str
...
+        {
+            "finish_reason": finish_reason,
+            "provider_finish_reason": provider_finish_reason,
+        }
+    )
+    if not text:
+        raise CompactionSummaryError(
+            f"compaction summary was empty (finish_reason={finish_reason!r})"
+        )
+    if truncated and not _handoff_titles_complete(text):
```
Nobody-Zhang: **[xhigh review · sweep · CONFIRMED · correctness]** `_require_usable_summary` accepts a reply the provider flagged as truncated (`length`/`max_tokens`/`incomplete`) whenever all eight heading strings occur, and the 2× retry is gated on `truncated and not titles_complete`. A cut that leaves the headings in place lands in the *last*, append-only section — `## Key Artifacts`, the one most likely to be long — and every later chunk/compaction is told that handoff is "authoritative; Key Artifacts are append-only — never drop an item", so the artifact ids that were cut never come back. Probe: `titles complete: True / truncated: True / ACCEPTED … retry issued: False`.

Fix applied: any truncation earns the one retry at double budget (bounded by the cap/room); a reply still truncated after the retry is accepted only when its headings are complete. Test `test_a_truncated_summary_with_complete_headings_is_retried_once` asserts `max_tokens == [8192, 16384]`.

openai4s/agent/compaction.py
```diff
@@ -655,34 +1077,61 @@ def compact(
...
+    # sized against what is left of the budget after it -- never below a
+    # floor, so an outsized handoff shortens chunks rather than zeroing them.
+    chunk_budget = min(48_000, int(0.3 * budget))
+    chunk_floor = max(1, chunk_budget // 6)
+    chunk_estimates: list[int] = []
+    raw_summary = ""
+    cap = _summary_output_cap(cfg)
+    max_tokens = _summary_max_tokens(cfg, cap)
+    index = 0
+    while index < len(pieces):
```
Nobody-Zhang, line 1442: **[xhigh review · sweep · CONFIRMED · correctness]** The rolling loop issues up to N blocking `chat()` calls with no `should_cancel`, outside the engine's `ChatModel` (which is the only thing that carries `cancellation=EventCancellation(st.cancel)`, `_auto_budget_chat`, the LLM quota gate and `RuntimeActionLedger` usage accounting). Stop is therefore ignored until every chunk — including transport 429/5xx backoff sleeps, which poll `should_cancel` only when supplied — has returned; the PR multiplied that un-cancellable, un-metered spend from one 1024-token call to N × (8192, 16384 on retry). One transient failure on chunk k also discards chunks 1..k-1 and restarts from chunk 1 next turn.

Fix applied for cancellation: `CompactionPolicy.should_cancel` (CLI: the run's cancellation; Web: `st.cancel.is_set`; children via `bind_runtime`) is polled between chunks and handed to each summary `chat()`; a cancelled compaction raises `CompactionCancelled`, which `_prepare` returns from without counting a failure. Not fixed here (product/architecture): routing the summary calls through the same admission/quota/metering seam as the model port so they are budgeted and billed to the session — worth a follow-up.

openai4s/agent/ledger.py
```diff
@@ -792,10 +933,68 @@ def reduce_action_groups(groups: Sequence[Mapping[str, Any]]) -> list[dict[str,
...
+    n_middle = len(compacted_messages)
+    body = [
+        (index, message)
+        for index, message in annotated
+        if not message.get("compaction_handoff")
+    ]
+    if not body or n_middle <= 0:
+        return None
+    rest = body[1:]
+    if n_middle > len(rest):
```
Nobody-Zhang, line 995: **[xhigh review · sweep · PLAUSIBLE (latent) · correctness]** The docstring promises `None` when "the live history carried a message the ledger has no group for", but the only guard is `n_middle > len(rest)`: a live-only excess shorter than the tail is absorbed by the count and the bound slides into the tail compact() kept. Probe B2 (two live-only user messages in the middle, a 4-message live tail) returned `cover = c5`, whose assistant+observation the running context still holds — a restore would delete them. Latent today (every current Web append is paired with a ledger group; synthesized observations make the ledger longer, which under-covers safely), but any future ledger-less injection turns it into silent tail deletion on restart.

Fix applied: both ends of the compacted middle must be the messages the ledger reconstructs there (`_same_message`: same role and content, or the live marker's `content_archive.sha256` equal to `content_digest(reconstructed)` for an externalized message); otherwise `None`. Test: an injected live-only notice covers nothing; an aligned middle and an externalized last message still resolve.

openai4s/server/session_package.py
```diff
@@ -3218,13 +3223,39 @@ def _import_ledger(
...
-                    result=event.get("result"),
+                    result=result,
                     side_effect_class=event.get("side_effect_class"),
                     resource_keys=list(event.get("resource_keys") or []),
                     created_at=event.get("created_at"),
                 )
         return group_map, action_map, turn_map
 
+    @staticmethod
+    def _remap_compaction_result(
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 8d3bba43: `_remap_compaction_result` now runs the imported `handoff` through `_scan_untrusted_text`, the same injection screen every other imported message and cell gets, so a handoff that reads as instructions restores under the security banner rather than as a bare `role=system` note. Regression test `test_session_package_screens_the_imported_compaction_handoff` imports a package whose handoff says "ignore all previous instructions" and asserts the restored note carries the banner; with the scan removed it restores bare.

openai4s/agent/compaction.py
```diff
@@ -439,14 +647,158 @@ def _write_content_blob(
...
+    Only a complete file ever appears at the digest path: the bytes go to a
+    sibling temp file and are renamed into place, so a crash, ENOSPC, or a
+    concurrent writer can never leave a torn ``<sha256>.json`` that every
+    later externalization would then reject forever.  A file already there
+    is accepted only if it *is* this content.  A torn one (not JSON at all)
+    is replaced on both roots.  A well-formed file with different content is
+    replaced only under the host-owned archive; in the agent-writable
+    workspace it is left alone and the caller records no reference to it.
+    """
+    path.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in 58863e67: the workspace blob write no longer trusts the name `_confine()` checked. Every component below the blob root is opened relative to its parent descriptor with `O_NOFOLLOW|O_DIRECTORY`, the existing-blob check uses `stat(follow_symlinks=False)` plus a bounded no-follow read on that descriptor, the temp file is created with `O_EXCL` there, and `rename(2)` with dir fds pins the directory the check looked at — so swapping `.openai4s` or `context` for a symlink after the check is ELOOP and nothing is written (and no `workspace_ref` is returned). A platform without dir_fd support skips the kernel-readable copy rather than writing by name. Regression test: `test_a_component_swapped_for_a_symlink_after_the_name_check_is_refused` (both components, name check stubbed to answer as before the swap).

PR #147, opened 2026-09-04
fix(frontend): bind the composer once Shell has rendered it

WenyuLiang:
## Summary

Typing a message and pressing Enter in the workbench composer inserted a
newline and never sent anything; Cmd+Enter did nothing either, and the
composer has no send button. Only the dashboard example cards could start a
turn.

Cause: `frontend/src/features/send/index.ts` runs `installSend()` at module
import, which calls `bindComposer()`. That happens before `main.tsx` renders
`<App/>`, so `#composer`, `#plan-toggle` and `#explore-toggle` do not exist
yet; `bindComposer()` found nothing, bound nothing, and nothing ever tried
again. `bindComposerAutocomplete()` (`features/autocomplete/composer.ts`)
already handles exactly this ordering with a MutationObserver, which is why
`@`/`#`/`/` completion worked in the same textarea while Enter did not.

Fix: `bindComposer()` now mirrors that pattern. The first pass still binds
whatever is mounted; if `#composer` is not there yet, a MutationObserver on
`document.documentElement` (childList + subtree) retries on DOM changes and
disconnects once the composer carries the keydown handler. Repeat calls are
no-ops (module-level watch guard plus the existing `data-send-bound`
markers). No behaviour of the handler itself changed.

Introduced by the F-13 / F-11 port (`d28010ba` moved the composer into
`Shell.tsx`, `8ab17fb1` wired the send chain). No existing issue.

## Area

- [x] Web UI / frontend
- [ ] Server / gateway / API
- [x] Tests / harness / CI
- [ ] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Test / harness
- [ ] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

Ran (against the current `main`, `e9370ef7`):

```text
npm test --prefix frontend                 # 61 files, 439 tests passed (3 new in send/bind.test.ts)
npm run build --prefix frontend            # strict tsc + vite; dist rebuilt, two consecutive builds byte-identical
uv run pre-commit run --files <changed files>  # end-of-file/line-ending/whitespace/merge-conflict/mypy passed
```

Manual, macOS 27.0 (26A5416b), Chromium-based browser, against a daemon
started with a fresh `OPENAI4S_DATA_DIR`:

- Before: `document.getElementById('composer').dataset.sendBound` is
  undefined after the workbench loads; a synthetic `keydown` Enter on the
  composer is not `defaultPrevented`; typing and pressing Enter only inserts a
  newline.
- After: `dataset.sendBound === "1"` on `#composer`, `#plan-toggle` and
  `#explore-toggle`; Enter in a session composer clears the box and issues
  `POST /api/v1/frames/<id>/message → 202 Accepted`, and the turn starts.

Not run:

```text
Full `uv run pytest` on this host
```

Reason:

```text
Only frontend sources and the committed dist changed. This branch is rebased
onto the current main; none of the six commits since the prior base touch
`frontend/src/features/send/`. The subset that reads the web UI / dist passes.
```

## Risk and Reviewer Focus

- `frontend/src/features/send/send.ts::bindComposer` / `bindComposerNow`:
  the observer's completion signal is "#composer is bound". The two mode
  toggles are rendered as static siblings of the composer in `Shell.tsx`, so
  they bind in the same pass; if they were ever moved into a later or
  conditional render they would need their own signal.
- Until the composer exists the observer sees every DOM mutation and runs
  three `getElementById` calls per batch; it disconnects on the first
  successful bind, same as the autocomplete watcher.
- A separate cosmetic issue (static labels stay as raw i18n keys when the
  locale chunk resolves after `Shell` mounts) was deliberately left out of
  this PR; a first attempt to re-apply `applyStaticI18n()` after load could
  clobber `#conv-title`, which `setTitle()` writes directly, so it needs its
  own change.
- Regression test: `frontend/src/features/send/bind.test.ts` (vitest, node
  environment, hand-made document/MutationObserver fakes in the style of
  `features/artifacts/boot.test.ts`).

### Diff Readiness

      `host_dispatch.py`, `store.py`, `worker.py`, `manager.py`. (none touched)
      source.
      reformatted.

### Tests

      a documented smoke test. (n/a — none touched)

### Core Dependency Policy

      stdlib web server.
      in-tree use site. (n/a — none added)
- [x] New dependencies, if any, are documented and justified. (none)

      verification notes. (n/a)
      guarantees.

### Docs

      appropriate. (behavior now matches what the docs already describe)
      affected. (n/a — untouched)


frontend/src/features/send/bind.test.ts
```diff
@@ -0,0 +1,142 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
```
Nobody-Zhang, line 1: **[CI red] `frontend/src/features/send/README.md` / `README_zh.md` do not list `bind.test.ts`.**

CLAUDE.md: every maintained directory keeps a bilingual README pair that lists its direct files, checked by `scripts/check_directory_readmes.py` in its own CI job. At this head:
```
directory docs: frontend/src/features/send/README.md does not mention `bind.test.ts`
directory docs: frontend/src/features/send/README_zh.md does not mention `bind.test.ts`
```
That is the one red job on the PR. Fix: one table row in each file, next to `admission.test.ts` / `install.test.ts`.

frontend/src/features/send/send.ts
```diff
@@ -368,4 +390,5 @@ export function bindComposer(): void {
       }
     });
   }
+  return !!c && c.dataset.sendBound === "1";
```
Nobody-Zhang: **[correctness] The Enter handler (line 389, `void send(c.value)`) has no in-flight guard, so a double or held Enter sends twice and creates two sessions.**

`send()` only clears the composer at line 223, *after* `await api("/frames", POST)` + `await loadSessions()` on a fresh session (lines 178-188) and after `await loadSkillsCatalog()` for any `/skill` token. `running.value` / `enableComposer(false)` are also set after those awaits (217-218). Every keydown in that window runs synchronously to line 178, sees `currentId.value === ""`, and POSTs a second `/frames`. Keydown autorepeat is ~30/s, so holding Enter on the dashboard is enough.

Before this PR the handler was never attached, so this is newly reachable. Suggest dropping Enter while a dispatch is pending (a module-level in-flight promise in the keydown handler), with a unit test that a second Enter before the first dispatch resolves is not dispatched.

frontend/src/features/send/send.ts
```diff
@@ -327,8 +327,30 @@ export async function send(text?: string | null, opts?: { execute?: boolean }):
...
 export function bindComposer(): void {
   if (typeof document === "undefined") return;
+  if (bindComposerNow() || composerWatch) return;
+  // installSend() runs at module import, before Shell.tsx has rendered
+  // #composer, so this first pass finds nothing -- and nothing ever tried
+  // again: Enter inserted a newline and no typed message could be sent. Same
+  // ordering bindComposerAutocomplete already handles: bind when the node
+  // appears, then stop watching.
+  if (typeof MutationObserver !== "function") return;
+  composerWatch = new MutationObserver(() => {
```
Nobody-Zhang: **[altitude] Import-time bind + a document-wide MutationObserver, where a one-line post-render call already has a home.**

`main.tsx` calls Preact's synchronous `render(<App/>, mount)` and then `bootChrome()` / `bootAttention()` — both bind Shell-only nodes and work precisely because render is synchronous. `bindComposer()` belongs on that line. The artifacts lane solved the same import-before-render ordering with a DOM-free install + post-render finish (`artifacts/boot.ts`, pinned by `boot.test.ts` "keeps pre-render boot DOM-free"), not an observer.

What the observer costs: ~25 lines of lifecycle state, a module-level handle with no reset, a fourth copy of `observe(document.documentElement || document.body, {childList, subtree})` (`autocomplete/composer.ts:249`, `autocomplete/editor.ts:317`, `sessions/chrome.ts:163`), two observers racing for the same `#composer` node, and a completion predicate that must track Shell's DOM (see the toggle comment below). The observer also fires exactly once today — it exists only to defer the bind by one microtask past `render()`.

Suggested shape: remove `bindComposer()` from `installSend()` (keep install DOM-free), `import { bindComposer } from "./features/send"` in `main.tsx` and call it right after `render(...)`, delete `composerWatch` + the observer, and make `bindComposer` the plain binder again.

frontend/src/features/send/bind.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    onKey(ime);
+    expect(ime.preventDefault).not.toHaveBeenCalled();
+
+    (globalThis as { ac?: { open?: boolean } }).ac = { open: true };
+    const withAutocomplete = { key: "Enter", preventDefault: vi.fn() };
+    onKey(withAutocomplete);
+    expect(withAutocomplete.preventDefault).not.toHaveBeenCalled();
+    delete (globalThis as { ac?: unknown }).ac;
+
+    // Empty composer: send() returns early, but the key is still consumed.
```
Nobody-Zhang: **[test-coverage] This test is named "sends on Enter" but never asserts that `send()` ran.**

Mutation check: replacing `void send(c.value);` in `send.ts` with a comment leaves this file 3/3 green. The only assertion on the Enter path is `preventDefault`. It also invokes the *real* async `send("")` against a fake `document` with no `querySelector`; it stays inert only because `send()` bails at `if (!text && !anns.length)` before touching `$()` / `api()`.

Suggest giving `bindComposer` an injectable dispatch (`bindComposer(dispatch = send)`) so the test can assert `dispatch` was called with the composer text, and never runs the real chain.

frontend/src/features/send/bind.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    bindComposer();
+
+    expect(composer.dataset.sendBound).toBe("1");
+    expect(composer.listeners.keydown).toHaveLength(1);
+    expect(plan.dataset.sendBound).toBe("1");
+    expect(explore.dataset.sendBound).toBe("1");
+    expect(observers).toHaveLength(0);
+  });
+
+  it("waits for Shell to render #composer instead of giving up at import time", () => {
```
Nobody-Zhang: **[test-coverage] No browser gate presses Enter in `#composer`, so the outage this PR fixes was invisible to CI and stays invisible.**

`tests/browser_admission_fault.mjs:105-110` and `tests/browser_p1_controls.mjs:258` set `box.value` and call `window.send(...)` directly from `page.evaluate`; the only `page.keyboard.press("Enter")` in `browser_smoke.mjs` is the timeline overview. This unit test runs under vitest `environment: "node"` with a hand-made document and a stubbed `MutationObserver`, so a regression in real boot ordering, a composer remount, or an engine-specific keydown difference (IME `keyCode 229` / `isComposing`) keeps every gate green while typing+Enter is dead in a product with no send button.

`tests/browser_matrix.mjs` runs in all three engines against a real daemon; one check there — open the session, `fill` the composer, `press("Enter")`, observe the `POST /frames/<id>/message` — closes the class.

frontend/src/features/send/send.ts
Nobody-Zhang: **[correctness, latent] The completion signal covers `#composer` only, but this function is also the sole binder of `#plan-toggle` and `#explore-toggle`.**

Once `#composer` reports bound, the observer disconnects and nothing ever calls `bindComposer()` again, so a toggle that lands in a later mutation batch never gets its `onclick`. Today the three are siblings in one Preact commit, so it cannot happen; gate `.nb-foot` behind a conditional/lazy boundary or move the toggles into an island and Plan/Explore become silently dead buttons (also `sessions/actions.ts:179,184`, which reach them via `$("#plan-toggle")?.click()`). The test at `bind.test.ts:104` pins this: a document holding only `{ composer }` still asserts `disconnect` was called. A post-render call binds all three unconditionally and has no predicate to keep in sync.

frontend/src/features/send/send.ts
```diff
@@ -327,8 +327,30 @@ export async function send(text?: string | null, opts?: { execute?: boolean }):
...
+    if (!bindComposerNow()) return;
+    composerWatch?.disconnect();
+    composerWatch = null;
```
Nobody-Zhang: **[correctness, latent] One-shot watch + node-local `data-send-bound` marker: a future re-creation of the textarea loses Enter permanently.**

After this line nothing observes, nothing re-calls `bindComposer()`, and idempotence lives on the DOM node. `Shell.tsx` is static today (no key / conditional, one `render()` into `#app`, prefresh HMR reuses the node), so this needs a future keyed or conditional composer subtree — but when it happens the pre-PR outage returns silently and `bind.test.ts` stays green. Not something an observer fixes either; the honest guard is the real-browser assertion above.

frontend/src/features/send/send.ts
```diff
@@ -327,8 +327,30 @@ export async function send(text?: string | null, opts?: { execute?: boolean }):
   void loadSessions();
 }
 
+let composerWatch: MutationObserver | null = null;
+
 export function bindComposer(): void {
   if (typeof document === "undefined") return;
+  if (bindComposerNow() || composerWatch) return;
```
Nobody-Zhang: **[test-coverage] The `|| composerWatch` disjunct is unexercised — deleting it leaves all three tests green.**

Mutation check: `if (bindComposerNow()) return;` → 3/3 passed. `bind.test.ts:107` ("A later call is a no-op") runs after the composer was mounted and bound, so it takes the `bindComposerNow()` branch, never the watch guard. `installSend()`'s docstring justifies "Safe to call more than once" only by `registerUnlessPresent`, so a refactor that drops this disjunct would stack one document-wide observer per call with a green suite. (Moot if the observer goes.)

frontend/src/features/send/bind.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  const documentElement = {};
+  vi.stubGlobal("document", {
+    documentElement,
+    body: {},
+    getElementById: (id: string) => nodes()[id] ?? null,
+  });
+  return documentElement;
+}
+
+afterEach(() => {
```
Nobody-Zhang, line 56: **[test-coverage] `composerWatch` is module state in `send.ts` with no reset; this file is order-coupled.**

Vitest isolates per file, not per test, and this `afterEach` only clears `ac` and unstubs globals. The three tests pass in any order today (shuffle seeds 1/7/42/99/2026 all green) only because test 2 drives the mounted callback that nulls the handle. Reproduced off-tree: a test whose composer never mounts, followed by a copy of test 2, fails `expected [] to have a length of 1` because `send.ts:334` short-circuits on the stale handle; a mid-test failure in test 2 before line 101 cascades the same way. Either export a reset or drop the module state.

frontend/src/features/send/send.ts
```diff
@@ -327,8 +327,30 @@ export async function send(text?: string | null, opts?: { execute?: boolean }):
...
+  // again: Enter inserted a newline and no typed message could be sent. Same
+  // ordering bindComposerAutocomplete already handles: bind when the node
+  // appears, then stop watching.
+  if (typeof MutationObserver !== "function") return;
+  composerWatch = new MutationObserver(() => {
+    if (!bindComposerNow()) return;
+    composerWatch?.disconnect();
+    composerWatch = null;
+  });
+  composerWatch.observe(document.documentElement || document.body, {
```
Nobody-Zhang: **[efficiency, latent] Nothing disconnects this observer if `#composer` never mounts.**

No timeout, no teardown export, and the callback ignores its records. Today it always disconnects (only `dist/index.html` loads the bundle, `App` renders `Shell` unconditionally, `#composer` is static). Any future route that renders something other than `Shell`, a satellite page that loads the bundle, or a conditional composer keeps this and `autocomplete/composer.ts:249`'s identical observer attached to `documentElement` `{subtree: true}` through every streaming DOM batch. A synchronous post-render call carries no such state.

frontend/src/features/send/send.ts
Nobody-Zhang: **[simplification]** `return !!c && c.dataset.sendBound === "1"` is always `!!c` — the marker is set on every non-null `c` twelve lines up, so the comparison invites a reader to believe a "mounted but unbound" state exists. `document.documentElement || document.body` (line 346) is unreachable in a browser (documentElement is non-null once any script runs) and it propagated a dead `body: {}` into the test stub. The module-level handle + `composerWatch?.disconnect()` does what the sibling's local `const ob` does with no state and no optional chain. All three vanish with a post-render bind.

frontend/src/features/send/bind.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    // A later call is a no-op: no second listener, no second observer.
+    bindComposer();
+    expect(composer.listeners.keydown).toHaveLength(1);
+    expect(observers).toHaveLength(1);
+  });
+
+  it("sends on Enter and leaves Shift+Enter, IME composition and an open autocomplete alone", () => {
+    const composer = fakeComposer();
+    stubMutationObserver();
```
Nobody-Zhang: **[simplification]** `stubMutationObserver()` here stubs a constructor this test never reaches (the composer is present, so `bindComposerNow()` returns true at line 334) and its return value is discarded — unlike test 1, where `expect(observers).toHaveLength(0)` makes the same call load-bearing. Either drop it or assert on it. Likewise the `ac` global is set/deleted through two `globalThis as { ac?: ... }` casts plus a duplicated `afterEach` delete; `vi.stubGlobal("ac", { open: true })` is already undone by the existing `vi.unstubAllGlobals()`.

PR #151, opened 2026-09-08
fix: align Windows WSL startup, isolation, and scientific runtime

difficulttopickaname:
# fix: align Windows WSL startup, isolation, and scientific runtime

## Summary

Windows users with WSL2 could start a daemon yet encounter an unavailable workbench, unusable credential storage or scientific commands, and kernels that lost their state after a request ended. A detached server could also disappear when WSL stopped an otherwise idle distribution. This change repairs those paths and records a reproducible audit of the remaining differences from macOS.

- Keep the Windows-launched server alive without an open terminal, check the authenticated workbench before reporting ready, select a user distribution, and guide preparation of missing isolation support.
- Add a focused WSL sandbox adapter and Windows user-scoped DPAPI credential backend. Preserve an existing usable Linux keyring and keep the core standard-library-only.
- Preserve persistent kernel state across request-thread exit, page refresh and scoped interruption. Carry sandbox descriptors into environment probes and dynamic tools, use Bash explicitly, and filter Windows entries from the Linux tool search path.
- Install immutable payloads through an atomic pointer switch. Preserve a running analysis during an update, report its actual build identity, and keep status/stop available without the new ZIP payload.
- Repair shared Linux/macOS bundle contracts: include the default workbench assets, relocate Python console scripts, and prevent the current directory from shadowing the packaged application.
- Normalize a socket-timeout/watchdog race reproduced by the full regression run, with focused coverage in the shared HTTP deadline layer.

The existing scientific execution protocol and frontend are retained. WSL policy, encrypted storage, process creation and packaging remain in their owning layers.

## Area

- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Feature
- [x] Test / harness
- [x] Documentation
- [x] Security-related change
- [x] Release / packaging

## Verification

Passed on an isolated Ubuntu 24.04.4 WSL2 x86_64 distribution, as an ordinary user, with enforced isolation; launcher checks used Windows PowerShell 5.1 and browser checks used Chromium on Windows.

```sh
uv run pytest --tb=short -n 2 --dist loadfile \
  --basetemp="$TMPDIR/pytest" -o tmp_path_retention_policy=failed
```

- Full offline suite: **8345 passed, 107 skipped, 0 failed**, one expected unsupported-platform warning, 1052.39 seconds. The independent run used a private temporary filesystem and test-only additional swap, removed afterwards. An earlier concurrent run was OOM-killed and is not counted as a pass.
- Focused lifetime/MCP checks: **37 passed**; packaging/WSL checks: **113 passed**. PowerShell 5.1 launcher contracts passed. All-file formatting/lint/strict mypy, 160-directory bilingual coverage and source credential scanning passed; offline scenario harness: **38/38**.
- Real WSL sandbox and Python/R interruption checks passed, including Windows interop/mount/process aliases, Hyper-V sockets, confined environment probes and dynamic tools. R/jsonlite were test prerequisites, not additions to the base package.
- The candidate Linux bundle passed embedded-runtime/scientific-import/resource/relocation verification; the Windows ZIP passed native verification; wheel/sdist build and verification passed. The candidate reused the assembled scientific runtime and reran source staging, CLI generation, bytecode compilation and package verification with the product fixes.
- Actual installed-package browser checks passed: default workbench and assets, synthetic encrypted model-credential save/delete, CSV upload/download, calculation, plotted PNG, exact result table, refresh/interrupt with dataframe retention, and Notebook export. Notebook REPL was explicitly enabled; model/upload setup used authenticated API calls, not a certification of the setup wizard.
- The candidate remained reachable for 120 seconds after launcher exit without any terminal, WSL command or UNC read maintaining its lifetime. Update testing preserved the running PID, build identity and dataframe. After a WSL restart, saved results remained identical, inputs could be re-read and computed, and a synthetic encrypted secret was recovered and deleted.

### Known failing gate and unverified scope

**This PR does not establish a fully green Windows release gate.** The unmodified `tests/browser_smoke.mjs` still exceeded its 20-second queued-Agent admission limit without concurrent build load. A separate timed diagnostic observed normal handoff after 18.585 seconds; an earlier loaded run took 39.1 seconds. Those diagnostic runs are not substitutes for a passing full browser smoke.

Native macOS execution, Windows ARM, other distributions/network modes, real provider login/inference, full Windows reboot, Conda provisioning and every scientific recipe were not certified in this WSL x86_64 lab. Cross-browser/container jobs were not rerun here. HTTP response-schema/route recapture was not run because no gateway route or serializer changed.

## Risk and Reviewer Focus

- Review the WSL mount/PID/socket boundary in `security/wsl.py` and `security/sandbox.py`, descriptor ownership, and the Windows credential backend's stdin-only protocol and namespace binding.
- Review Linux process creation in `kernel/transport.py`: one persistent thread owns process creation so a short-lived request thread cannot terminate its kernel. Protocol readers and cancellation ownership remain unchanged.
- Review PowerShell 5.1 native argument quoting and hidden-process lifetime, plus atomic installation and running-build detection. One earlier normal stop exceeded the default five-second wait and returned 2 before exiting on its own; final-package stops returned 0, without force.
- Shared packaging changes affect macOS as well as Linux, although only WSL/Linux execution was exercised locally. The audit documents both baseline failures and later fixes; the baseline observations are not current-branch results.

Evidence: [English audit](https://github.com/PKU-YuanGroup/OpenAI4S/blob/feat/windows-wsl-parity/docs/windows-wsl-parity-audit.md), [Chinese audit](https://github.com/PKU-YuanGroup/OpenAI4S/blob/feat/windows-wsl-parity/docs/windows-wsl-parity-audit_zh.md), and the `fix_verification` section of [recorded evidence](https://github.com/PKU-YuanGroup/OpenAI4S/blob/feat/windows-wsl-parity/docs/windows-wsl-parity-evidence.json). The separate shell inventory records syntax/coverage, not certification of every workflow.

## Checklist

- [x] Full diff reviewed; no unrelated formatting or broad facade/frontend rewrite.
- [ ] Small diff: 44 files are changed; the generated shell inventory, audit evidence and regression coverage account for much of the added content.
- [x] Relevant checks and remaining failures are disclosed above.
- [x] No tests were deleted; default offline tests require no live provider credentials or external services.
- [x] No hard third-party dependency was added to the core.
- [x] No vendor or byte-exact fixture reformatting.
- [x] Security checks use synthetic data; no real model credentials, private research results or personal workstation paths are included. Evidence retains synthetic lab paths needed to demonstrate relocation.
- [x] User documentation and bilingual directory indexes reflect the changes; isolation and release-readiness claims are bounded by the recorded evidence.


scripts/windows/openai4s.ps1
```diff
@@ -654,6 +692,51 @@ function Test-OpenAI4SServing([string] $Distro, [string] $BootstrapLinux, [strin
...
+    Stop-WithGuidance 'the daemon is reachable but the OpenAI4S page could not be loaded.' @(
+        'The application is not ready to use. Run OpenAI4S.cmd doctor.',
+        'If an older version is still running, finish your work, run',
+        'OpenAI4S.cmd stop, and open this package again.'
+    )
+}
+
+function Show-RunningBuild([string] $Distro, [string] $BootstrapLinux, [string] $BundleDir, [string] $Expected) {
+    $result = Invoke-BootstrapCapture $Distro $BootstrapLinux @('cli', $BundleDir, 'status', '--json')
+    # Invoke-WslCaptureNative folds stderr into Output on purpose (wsl.exe's
```
Nobody-Zhang, line 725: **`Show-RunningBuild` parses the merged stdout+stderr capture as JSON, so one WSL warning line makes every launch claim a stale session.**

`Invoke-WslCaptureNative` folds stderr into `Output` *on purpose* — its own comment names the NAT localhost-proxy warning as the reason — so the capture is not a JSON document. `Get-AppUrl` guards the same capture with `Where-Object { $_ -match '^https?://' }`; this one joins every line and hands the result to `ConvertFrom-Json`.

Failure: a user with a Windows localhost proxy (the Clash/TUN audience this file exists for) launches against a daemon running *this exact build*. `wsl.exe` prints its warning, `ConvertFrom-Json` throws into the empty `catch`, `$running` stays `$null`, `$running.version` silently yields `$null` → `'unknown'`, and the launcher prints "update installed; your existing session is still running version unknown. Finish your work, run `OpenAI4S.cmd stop`…" on **every** launch, permanently.

Fixed: select the JSON line out of the capture before parsing, the way `Get-AppUrl` selects the URL. Covered by a new native PowerShell 5.1 contract in `tests/test_windows_launcher.ps1`.

scripts/windows/openai4s.ps1
```diff
@@ -654,6 +692,51 @@ function Test-OpenAI4SServing([string] $Distro, [string] $BootstrapLinux, [strin
     return $result.ExitCode -eq 0
 }
 
+function Assert-AppPage([string] $AppUrl) {
```
Nobody-Zhang, line 695: **`Assert-AppPage` routes through the machine's WinINET proxy and demands a byte-exact title, so it hard-refuses a daemon `Test-Serving` just proved healthy.**

Two independent triggers, both ending in `Stop-WithGuidance` (exit 1, browser never opened) after *both* `Test-Serving` and `Test-OpenAI4SServing` passed:

1. **Proxy.** Windows PowerShell 5.1's `Invoke-WebRequest` resolves `[Net.WebRequest]::DefaultWebProxy` from WinINET and has **no `-NoProxy`** (that is PowerShell 6+). Every other reachability check in this file deliberately uses a raw `TcpClient` for exactly this reason. With `localhostForwarding=false`, `$AppUrl` carries the WSL NAT address (`172.x`) — a dotted private literal that "bypass proxy for local addresses" does not cover — so the request is proxied, fails, and the launch is refused.
2. **Title.** `-match '<title>OpenAI4S</title>'` rejects a team daemon, whose HTML `GET` 303s to `/login` with the title `OpenAI4S — sign in` (`gateway.py` `_wants_html` branch, `webui/login.html`). That is the application answering.

Fixed: null `DefaultWebProxy` for the duration of the call (restored in `finally`), and match the title as a prefix. Both are asserted by a new PowerShell 5.1 contract.

openai4s/security/wsl.py
```diff
@@ -0,0 +1,187 @@
...
+            "/tmp",
+            "/var/tmp",
+            "/mnt/wsl",
+            "/mnt/wslg",
+            *windows_mounts(),
+        )
+    )
+
+
+def resolver_rebinds(masked: tuple[str, ...]) -> list[str]:
```
Nobody-Zhang, line 90: **`/run` and `/mnt/wsl` are masked with an *empty* tmpfs, which dangles `/etc/resolv.conf` and kills DNS in every network-enabled WSL Cell.**

`_bwrap_read_masks` turns a directory denial into `--tmpfs <dir>`, not a read-only bind. On WSL2 `/etc/resolv.conf` is a symlink either to `/mnt/wsl/resolv.conf` or — on the systemd Ubuntu 24.04 that `bootstrap.sh` tells the user to install — to `/run/systemd/resolve/stub-resolv.conf`. Both targets are now inside a fresh empty filesystem.

Failure: with `OPENAI4S_KERNEL_ALLOW_RAW_NETWORK=1` (the only posture where a Cell has network at all), every `requests.get("https://…")`, `pip install`, and R `download.file` fails with `Name or service not known`, while the identical Cell resolves fine on plain Linux — where `--ro-bind / /` leaves `/run` readable. Neither gate sees it: the self-test probe dials the literal `198.51.100.1`, and `harness/smoke/wsl_sandbox.py` runs with `forbid_raw_network=True`.

The repo already solves this exact problem: `security/byoc_confinement.py::_runtime_dns_rebinds()` rescues the resolved target with `--ro-bind-try`.

Fixed: `wsl.resolver_rebinds()` emits the same rescue, placed *before* the read-only remounts (a bind whose destination must be created inside an already-immutable tmpfs fails). Also dropped `--remount-ro` for `/tmp` and `/var/tmp`: a private tmpfs is already invisible to the host, so remounting those read-only bought no isolation and only made a hardcoded `/tmp/...` write fail on WSL alone.

scripts/windows/openai4s.ps1
```diff
@@ -737,13 +820,37 @@ if ((Test-LocalhostForwardingDisabled) -and
         Write-Host "  localhostForwarding=false detected; using WSL NAT address $ClientHost." -ForegroundColor Yellow
     }
 }
-$facts = Get-PackageFacts
 $packageLinux = ConvertTo-WslPath $distro $Here
 $bootstrap = "$packageLinux/wsl/bootstrap.sh"
+
+# Management remains available even when this ZIP's payload is missing or bad.
+# Exit 11 is bootstrap.sh's "nothing is installed in this data directory": on a
```
Nobody-Zhang, line 827: **The new early management-command exit targets `current`, so `doctor`, `--help`, `status` and `verify-package` all refuse on a fresh unzip.**

Previously every argument path ran `install` first, so those commands always had a bundle. This exit hands `'current'` to `resolve_app`, which on a machine with no prior install falls off the end and exits.

Failure: first-ever run on a clean machine. `OpenAI4S.cmd doctor` — the command `Assert-AppPage`'s own guidance sends a stuck user to, and which `Test-SandboxIndependentCli`'s comment promises "must remain reachable when bubblewrap is the thing the user is trying to diagnose" — prints `OpenAI4S is not installed in /home/u/.openai4s` and exits 1. `OpenAI4S.cmd --help` likewise.

Fixed: `bootstrap.sh` now reports "nothing is installed in this data directory" as exit **11**, distinct from a genuine failure, and the launcher falls through to the install path on that code only. Preflight still does not gate the management commands.

scripts/windows/bootstrap.sh
```diff
@@ -309,60 +427,86 @@ install)
...
   # The marker is the *last* step, because it is what makes the next launch
   # take the already-installed fast path. Written first, an install that then
   # failed to write pip.conf (unwritable tree, ENOSPC) left a tree marked
   # complete: every later launch skipped the extraction, re-entered
   # configure_network with FRESH_INSTALL=0, and failed the same way with no
   # route back to a working install.
-  printf '%s\n' "$EXPECTED" > "$MARKER"
+  printf '%s\n' "$EXPECTED" > "$STAGED_APP/.installed"
+  if [ -e "$APP" ]; then
+    # This path is the payload's own digest, so an existing tree here is the
```
Nobody-Zhang, line 453: **Refusing when `$APP` already exists turns a damaged install into a permanent dead end.**

The old `rm -rf "$APP"` meant a tree that failed the fast-path predicate (`.installed` matches *and* `bin/openai4s` is executable) was simply rebuilt. The replacement refuses — and it refuses *after* a full staged re-extract, i.e. exactly when reinstalling was the intended repair.

Failure: a WSL crash, VHDX corruption or a lost executable bit leaves `~/.openai4s/app/OpenAI4S-…-<digest>/bin/openai4s` non-executable. The fast path rejects it, the script re-extracts ~1 GB into staging, then refuses. Every subsequent launch repeats both and refuses again. The suggested remedy (`doctor`) does not delete app trees, and per the finding above it could not even start. The only exit is a manual `rm -rf` the user is never told about.

Fixed: `$APP` is the payload's own digest, so an existing tree there is the same bytes — replace it, unless the recorded daemon is still executing out of it (`readlink /proc/<pid>/exe`), in which case say so and name `OpenAI4S.cmd stop`. Regression added in `tests/test_wsl_parity.py`.

scripts/windows/bootstrap.sh
```diff
@@ -270,21 +275,134 @@ digest_of() {
...
+  RUNNING_PID="$(cat "$PIDFILE" 2>/dev/null || true)"
+  case "$RUNNING_PID" in ''|*[!0-9]*) return 0 ;; esac
+  readlink "/proc/$RUNNING_PID/exe" 2>/dev/null || true
+}
+
+# Immutable payloads are never overwritten, so without this every update left a
+# full bundle behind -- inside a WSL VHDX, which does not shrink when files are
+# finally deleted. Four releases is several GB, and the fifth install fails with
+# a disk-full message about a disk this scheme filled.
+prune_installs() {
```
Nobody-Zhang, line 354: **Content-addressed payloads are never pruned, so every update leaks a full bundle into a VHDX that does not shrink.**

`APP="$APP_ROOT/$DIRNAME-$EXPECTED"` makes each distinct payload digest a permanent directory, `select_install` only re-points a symlink, and nothing in the script ever removes an old tree (`rm -rf` survives only inside `cleanup_install` for `.staging-*`). `install_cli_link` points `~/.local/bin/openai4s` at the *versioned* path, so old trees stay reachable too — orphaned is not the same as unreferenced.

Failure: a user who takes four releases carries ~4 GB under `~/.openai4s/app/` inside the WSL VHDX, which does not reclaim space when files are eventually deleted. The fifth install hits ENOSPC during `tar -xzf` and the launcher reports "the usual causes are a full disk inside the distribution" — a disk this scheme filled.

Fixed: `prune_installs` reclaims every superseded tree after a successful selection, skipping the one the recorded daemon is executing from. Regression added.

scripts/windows/bootstrap.sh
```diff
@@ -309,60 +427,86 @@ install)
...
+  prune_installs "$APP"
   echo "installed $APP"
   ;;
 
 serve)
   DIRNAME="${1:?serve needs the bundle directory name}"
   HOST="${2:-127.0.0.1}"
   PORT="${3:-8760}"
-  APP="$APP_ROOT/$DIRNAME"
+  # The Windows launcher runs this action inside a hidden wsl.exe with no
```
Nobody-Zhang, line 477: **Every diagnostic the `serve` action emits before `exec` is discarded, and the log the guidance points at is empty.**

`Start-Bootstrap` uses `Start-Process -WindowStyle Hidden -PassThru` with no stream capture, and the `>>"$DATA_DIR/logs/app.out"` redirect is attached only to the final `exec`. So `resolve_app`'s refusals, `configure_fake_ip_dns`'s `exit 2`, a failed `mkdir`, and an unreadable `.installed` all go to a hidden console and nowhere else. The `catch` around `Start-Bootstrap` is effectively dead — `Start-Process` throws only if `wsl.exe` itself cannot be launched.

Failure: a user sets `OPENAI4S_WSL_FAKE_IP_DNS=maybe`. bootstrap exits 2 in ~50 ms with `invalid OPENAI4S_FAKE_IP_DNS_MODE` on a hidden stderr. `$server.HasExited` breaks the loop, and the launcher prints "the daemon started but http://127.0.0.1:8760/ never answered within 60s" plus "Read the log from inside WSL" — pointing at an empty `app.out`. The one message that says what happened is unreachable.

Fixed: the `serve` action redirects its own stdout/stderr into the log *before* it resolves anything, and `HasExited` now produces its own diagnosis ("the daemon exited before it started serving") instead of the localhost-forwarding guidance.

openai4s/security/wsl.py
```diff
@@ -0,0 +1,187 @@
...
+#: boundary on the strength of the host kernel's name.
+_WSL_MARKERS = (
+    "/run/WSL",
+    "/mnt/wsl",
+    "/proc/sys/fs/binfmt_misc/WSLInterop",
+    "/proc/sys/fs/binfmt_misc/WSLInterop-late",
+)
+
+
+def is_wsl() -> bool:
```
Nobody-Zhang, line 30: **`is_wsl()` is true inside any container running on a Windows host, so a documented Docker deployment silently inherits the WSL boundary policy.**

`"microsoft" in platform.release()` matches WSL1 and — more importantly — matches from *inside* a Docker/Podman container on Docker Desktop for Windows, whose kernel release is `…-microsoft-standard-WSL2`. Such a container has no `/init`, no `/run/WSL` and no DrvFs mounts, but this predicate now drives `--unshare-pid`, the seccomp filter, the `/run` `/tmp` `/var/tmp` masks, the loss of `_bwrap_daemon_environ_mask()`, and three `required_checks` in the sandbox self-test. `docs/docker.md` is a supported deployment; `scripts/container_smoke.sh` runs on a Linux runner and would never see it.

Fixed: require a surface WSL itself creates (`/run/WSL`, `/mnt/wsl`, the `WSLInterop` binfmt handler, or `WSL_DISTRO_NAME`) rather than the host kernel's name. Regression added in `tests/test_wsl_parity.py`.

openai4s/security/windows_dpapi.py
```diff
@@ -0,0 +1,153 @@
...
+
+import hashlib
+import json
+import os
+import subprocess
+from pathlib import Path
+
+from .wsl import powershell_path
+
+# Raw stdin/stdout streams rather than [Console]::InputEncoding/OutputEncoding:
```
Nobody-Zhang, line 18: **The two `[Console]::…Encoding` assignments sit *outside* the script's `try`, and their setter is exactly the call that fails without an attached console.**

`Console.InputEncoding`'s setter calls `SetConsoleCP`, which throws `IOException: The handle is invalid` when the process has no console — the normal case for a child spawned by a headless daemon over WSL interop, which is precisely how this module is reached. Because the assignment precedes `try {`, the `catch` that prints the sanitized message and `exit 1` is bypassed and PowerShell dies with a raw .NET stack trace.

Failure: `available()` still returns `True` (it only stats the exe), so `_self_test`'s `put` fails, the backend is disqualified as "present but unusable", and on a host with no Linux keyring `SecretBroker._resolve` raises `SecretStoreUnavailable` — "no secure secret store on this host" on a machine where DPAPI is present and this feature was supposed to work. Nothing in the offline suite executes the script (`subprocess.run` is stubbed).

Fixed: read stdin and write stdout as raw UTF-8 byte streams (`OpenStandardInput`/`OpenStandardOutput`). That is correct regardless of the host code page and removes the two assignments entirely, so nothing runs outside the `try` any more.

openai4s/security/windows_dpapi.py
```diff
@@ -0,0 +1,153 @@
...
+}
+"""
+
+
+#: Read in order. D-Bus keeps the second in step with the first and survives on
+#: images that ship no /etc/machine-id at all.
+_IDENTITY_FILES = ("/etc/machine-id", "/var/lib/dbus/machine-id")
+
+
+def machine_identity() -> str:
```
Nobody-Zhang, line 83: **A missing `/etc/machine-id` raises a bare `FileNotFoundError` straight out of the secret broker, past every `except SecretBrokerError`.**

`Path("/etc/machine-id").read_text("ascii")` sits outside the `try`, and `WindowsDPAPIBackend._request` converts only `RuntimeError`. `available()` checks nothing but PowerShell reachability, so on a `wsl --import`ed minimal distribution (systemd off, dbus absent — several Alpine/Arch WSL images ship no `/etc/machine-id` at all) `auto` selects this backend and then the first `get`/`put` escapes the broker with an errno. The `auto` fallthrough to `EnvInjectionBackend` never happens, and `secret_migration.py`'s `except SecretBrokerError` does not catch it either. The *empty*-file case was handled; the missing/undecodable one was not.

Fixed: `machine_identity()` reads `/etc/machine-id` then `/var/lib/dbus/machine-id`, converts every OS failure to `RuntimeError`, and `available()` now probes what `request()` actually needs. `_request` converts `OSError` too, since this backend crosses the interop boundary.

openai4s/security/windows_dpapi.py
```diff
@@ -0,0 +1,153 @@
...
+                if ([IO.File]::Exists($temp)) { [IO.File]::Delete($temp) }
+            }
+        }
+        'get' {
+            if ([IO.File]::Exists($path)) {
+                $bytes = [Security.Cryptography.ProtectedData]::Unprotect([IO.File]::ReadAllBytes($path), $entropy, $scope)
+                $value = [Text.Encoding]::UTF8.GetString($bytes)
+            }
+        }
+        # File.Delete is a no-op on a missing file but throws
```
Nobody-Zhang, line 60: **`delete` is not idempotent: removing a key that was never stored raises, where both sibling backends are no-ops.**

`.NET File.Delete` is a no-op on a missing *file* but throws `DirectoryNotFoundException` on a missing *parent*, and `%LOCALAPPDATA%\OpenAI4S\secrets` is created only inside the `'put'` branch. `KeychainBackend.delete` and `SecretServiceBackend.delete` both ignore the CLI's exit code precisely so a delete that matched nothing is not an error.

Failure: a data directory restored from another machine, or one whose LocalAppData was cleared, cannot remove a dead credential reference — `SecretBroker.delete(ref)` raises `SecretBrokerError` and the ref stays in the Store. `harness/smoke/wsl_sandbox.py` also calls `backend.delete(...)` from a `finally:`, so a genuine `put` failure would be masked by this one.

Fixed: guard the delete on the directory existing.

openai4s/sdk/bash.py
```diff
@@ -34,6 +35,20 @@
 _STDERR_BUDGET_CHARS = 8_000
 
 
+def _shell_executable() -> str:
```
Nobody-Zhang, line 38: **`executable="/bin/bash"` is right about the dialect and wrong about the path: it removes `host.bash` entirely on any host without bash.**

The underlying defect is real — `shell=True` alone means `/bin/sh`, which is dash on Debian/Ubuntu, so `[[ ]]`, `source` and process substitution failed as syntax errors in a capability the model is told is Bash. But POSIX guarantees `/bin/sh`, not `/bin/bash`. On Alpine/busybox, distroless and NixOS (which ships `/bin/sh` and `/usr/bin/env` and nothing else), every `host.bash(...)` now hits the `except (OSError, ValueError)` arm with `FileNotFoundError: '/bin/bash'` and is reported as `status="launch_failed"`, naming an interpreter the user never asked for. The container gate only covers `python:3.14-slim-bookworm`, which has bash.

Fixed: `_shell_executable()` resolves bash with `shutil.which` and falls back to `/bin/sh` — the shell `shell=True` would have used anyway — so a dialect mismatch cannot become total loss of the capability.

openai4s/kernel/transport.py
```diff
@@ -121,6 +123,47 @@ def _reset_inherited_signal_dispositions() -> None:
     signal.signal(signal.SIGQUIT, signal.SIG_DFL)
 
 
+# Linux's parent-death signal follows the creating thread. A request/job thread
+# ends after one Cell, so spawning bubblewrap there silently kills a persistent
+# kernel as soon as the request finishes. These daemon-lived threads own process
+# creation only; protocol readers, interrupts and cleanup stay unchanged.
+#
+# More than one worker because the identity of the thread does not matter, only
```
Nobody-Zhang, line 131: **The module-global single-worker executor serializes every Linux kernel spawn behind an unbounded wait, and raises instead of spawning during interpreter shutdown.**

The PDEATHSIG diagnosis is right — `--die-with-parent` follows the *creating thread*, so a request thread's exit kills a persistent kernel. The mechanism carries three costs:

1. `max_workers=1` funnels every Python and R kernel spawn in the daemon, for every session, through one thread, and `.result()` has no timeout. A `fork`/`execve` of bwrap stalled on a hung DrvFs/9p mount — precisely the WSL failure mode this PR is about — blocks every other tenant's kernel start *and* `kernel/supervisor.py`'s watchdog replacement spawn, with no bound and no error.
2. `concurrent.futures.thread` registers `_python_exit` via `threading._register_atexit`; once it runs, `submit()` raises `RuntimeError: cannot schedule new futures after interpreter shutdown` where a plain `Popen` succeeded.
3. `_worker` catches `BaseException` and exits the thread while leaving it in `executor._threads`, so `_adjust_thread_count` (which only checks `len(_threads) < max_workers`) never replaces it — every later spawn then blocks forever.

Fixed: build the pool lazily (so a macOS/Windows daemon never creates it), give it four workers — thread *identity* does not matter for PDEATHSIG, only that it outlives the child, and pool threads never retire — and fall back to a direct `Popen` when `submit` refuses at shutdown. Only `submit` is guarded; a `RuntimeError` out of `Popen` itself still reaches the caller.

Deeper alternative, not taken here: scope `--die-with-parent` off the persistent-kernel argv (the supervisor's ABA-safe watchdog and `kill()`'s group ladder already reap orphans) and leave it on the short-lived `subprocess.run` users, or hoist ownership into `KernelSupervisor`, which CLAUDE.md already names as the owner of worker identity/lifecycle.

openai4s/security/wsl.py
```diff
@@ -0,0 +1,187 @@
...
+IO_URING_SETUP_NR = 425
+
+#: AUDIT_ARCH_* plus the `socket(2)` syscall number, per ABI.
+_SECCOMP_ARCH = {
+    "x86_64": (0xC000003E, 41),
+    "aarch64": (0xC00000B7, 198),
+}
+
+
+def socket_filter() -> bytes:
```
Nobody-Zhang, line 151: **Two gaps in the Hyper-V socket filter: a bare `KeyError` on any other architecture, and `io_uring` walks straight past it.**

1. `{...}[platform.machine()]` has no fallback. Reached from `KernelSandbox.wrap_command` and `_run_self_test`, an unlisted `platform.machine()` raises `KeyError: 'armv7l'` — not `SandboxUnavailableError` — so `enforce` cannot report *why* it failed closed, and `_run_self_test` also leaks the `TemporaryFile` created on the line before.
2. seccomp classifies syscalls, not the operations submitted inside a ring. `IORING_OP_SOCKET` (Linux 5.19+, and the WSL2 kernel is newer) creates an `AF_VSOCK` fd without this filter ever being consulted — the exact escape it exists to close, and one `--unshare-net` does not cover either, as the docstring itself says. The self-test's `wsl_vsock_denied` check would still report `true`.

The program is otherwise byte-exact; I simulated all 11 instructions against `seccomp_data`'s layout and both jump targets land correctly.

Fixed: raise a diagnosable `UnsupportedArchitecture`, and deny `io_uring_setup` with EPERM (2 instructions, ahead of the socket rule — Docker and systemd block io_uring by default for the same reason). `tests/test_wsl_parity.py` now interprets the emitted BPF and asserts the verdict for `write`, `AF_INET`, `AF_VSOCK`, `io_uring_setup`, x32 and i386 `socketcall`.

scripts/windows/bootstrap.sh
```diff
@@ -270,21 +275,134 @@ digest_of() {
...
+  if [ -L "$APP_ROOT/current" ]; then
+    CURRENT="$(readlink "$APP_ROOT/current")"
+    case "$CURRENT" in
+      */*|''|*[!a-zA-Z0-9._-]*) echo "invalid installed application pointer" >&2; exit 1 ;;
+    esac
+    case "$REQUESTED:$CURRENT" in
+      current:*|"$REQUESTED:$REQUESTED-"*) printf '%s\n' "$APP_ROOT/$CURRENT"; return ;;
+    esac
+  fi
+  # Every payload is content-addressed as <bundle>-<digest>, so `$APP_ROOT/
```
Nobody-Zhang, line 293: **`resolve_app`'s recovery scan is unreachable, so a fully installed tree reports "not installed".**

`install` writes the digest-suffixed `$APP_ROOT/$DIRNAME-$EXPECTED`, so `$APP_ROOT/$DIRNAME` is a path nothing ever creates — yet that is what this returns whenever the `current` symlink is absent or points at another bundle. The `for CANDIDATE in "$APP_ROOT"/OpenAI4S-*` fallback is gated behind `REQUESTED = current`, and `openai4s.ps1` passes `$facts.BundleDir` on every `serve`/`cli` call.

Failure: `select_install` is interrupted, or the user deletes `~/.openai4s/app/current`. The payload sits complete at `…/OpenAI4S-0.2.0-linux-x86_64-<sha>/`, and both `serve` and `cli` print `not installed: …/OpenAI4S-0.2.0-linux-x86_64` and exit — with the recovery loop four lines below. `select_install` makes that reachable in one more way: `ln -s` without `-f` (unlike `install_cli_link`'s own `ln -sfn`) aborts under `set -e` on a `.current-<pid>` left by a killed run whose pid was later recycled, and `mv -T` is a GNU extension that a BusyBox userspace rejects — after `$APP` and its `.installed` marker are already in place, so every later launch fails identically.

Fixed: scan the requested bundle's own generations before giving up (and only fall back to *any* installation for the literal `current`, so `serve` can never silently start a different build); `ln -sf`; and a portable `rm -f && mv` fallback when `mv -T` is unavailable, taken under the install lock so no concurrent installer can observe the gap.

Nobody-Zhang:
## Max-effort review pass + branch update

`main` is merged in (clean; only `tests/README*` overlapped) and the branch is now 0 commits behind. **34/34 checks pass**, 3 nightly-only jobs skipping.

15 findings are posted inline above; all 15 are fixed in the four commits on top of the merge. The three that most change user-visible behaviour:

- `Show-RunningBuild` parsed a capture that `Invoke-WslCaptureNative` folds stderr into *on purpose*, so `wsl.exe`'s NAT localhost-proxy warning — on the very machines this launcher targets — made every launch print "your session is still running version unknown / run `OpenAI4S.cmd stop`", permanently.
- `Assert-AppPage` went through the WinINET system proxy (PowerShell 5.1 has no `-NoProxy`; every other reachability check in that file uses a raw socket for exactly this reason), so with `localhostForwarding=false` it hard-refused a daemon `Test-Serving` had just proved healthy.
- The WSL masks covered `/run` and `/mnt/wsl` with an *empty* tmpfs, which is where WSL2 points `/etc/resolv.conf` — every hostname lookup in a network-enabled Cell failed while the identical Cell resolved on plain Linux. `byoc_confinement` already had the `--ro-bind-try` rescue; it is now applied here too.

Fixed alongside them, without a separate inline comment:

- `status --json` gated the build identity on `_recorded_endpoint` (a pid-generation check needing a Linux start token), so a healthy current daemon reported `version: null, bundle_id: null` — the same stale-session banner by another route.
- `windows-dpapi` was missing from `diagnostics.py`'s backend allowlist, so a WSL support bundle silently dropped the one field naming the store in use.
- A `preinstall` probe that degrades to unconfined under `auto` still inherited the sandbox's descriptors.
- The Windows PATH strip moved from the CLI entrypoint into `build_kernel_environment`, so the Jupyter KernelSpec bridge, the harness and in-process embedders are covered too.
- `--remount-ro` dropped for `/tmp` and `/var/tmp`: a private tmpfs is already invisible to the host, so it bought no isolation and only made a hardcoded `/tmp` write fail on WSL alone.
- The WSL smoke's interop-socket assertions were vacuous when the glob came back empty.
- The DPAPI interop child no longer inherits the daemon's environment (`WSLENV` cleared) or its controlling terminal.

New coverage: PowerShell 5.1 contracts for the JSON-line selection and the proxy-free page check (they ran green on the Windows runner), a BPF interpreter that asserts the seccomp filter's verdict for `write`/`AF_INET`/`AF_VSOCK`/`io_uring_setup`/x32/i386 `socketcall`, and bootstrap regressions for install repair, pruning and the exit-11 contract.

### Found, deliberately not fixed here

- **`SecretBroker` resolution runs inside `Store`'s single lock.** On a WSL host with `secret-tool` present but no session keyring — the host in this PR's own evidence — that is 30 s of secret-service timeouts plus up to 60 s of PowerShell cold starts with every other Store operation blocked behind `self._lock`. Fixing it means restructuring locking in `store.py`, a protected facade.
- **`/etc/machine-id` rotation.** `wsl --export`/`--import` regenerates it, after which every derived key changes: `get` returns `None`, the broker reports the credential as *not configured* rather than unreadable, and the old blobs stay in `%LOCALAPPDATA%\OpenAI4S\secrets` forever with nothing to enumerate or reclaim them. A `/var/lib/dbus/machine-id` fallback is in, but the binding itself is a deliberate design choice.
- **The new WSL sandbox branches have no automated coverage on any platform.** `tests/test_kernel_sandbox.py`'s autouse fixture and `test_sandbox_credential_denies.py` both force `is_wsl()` to `False`, and no CI job runs `harness/smoke/wsl_sandbox.py`. `tests/test_wsl_parity.py` now covers `wrap_bwrap_command` directly, but `KernelSandbox._private_pid`, the seccomp file lifecycle and the two-fd `popen_pass_fds()` still need a WSL runner.
- **Altitude on the process spawner.** Four workers and a lazy build remove the head-of-line blocking, but the deeper fix is to scope `--die-with-parent` off the persistent-kernel argv (the supervisor's ABA-safe watchdog and `kill()`'s group ladder already reap orphans) while the short-lived `subprocess.run` users keep it — or to hoist ownership into `KernelSupervisor`, which CLAUDE.md already names as the owner of worker identity and lifecycle.
- `OPENAI4S_BUNDLE_ID` and `OPENAI4S_NONINTERACTIVE` are load-bearing for the shipped launcher and appear in no documentation.
- Four install-contract tests now skip on macOS, a consequence of `bootstrap.sh` requiring `flock` and `mv -T`.
- `host.bash` running under bash widens the grammar `security/shellcheck.py`'s literal-pattern blocklist was written against (`<(…)`, `$'…'`, `${!v}`, `&>`). That module is currently reached only by tests and the benchmark acceptance layer, not by the `host.bash` execution path, so nothing changed in enforcement — but the two have drifted.

PR #153, opened 2026-09-08
chore(deps): update Python image, formatting hooks, and CodeQL action

Nobody-Zhang:
## Summary

Update the container and development/CI dependencies together. This PR supersedes #148, #149, and #150.

| Dependency | Previous | Updated |
| --- | --- | --- |
| Python 3.14 slim-bookworm image, both stages | `416f0db…` | `9ab8d9c…` |
| isort | 8.0.1 | 9.0.1 |
| Ruff | 0.16.4 | 0.16.6 |
| CodeQL SARIF upload action | 4.37.8 | 4.37.9 |

The isort upgrade requires import formatting in five Python files. It also exposes an isort/Black loop on two commented alias imports: shorten the worker's inline comments while preserving the trusted-path explanation, and apply a documented, single-import `isort: skip` in the scenario test. All five Python files retain identical ASTs. Frozen dependency identities, the existing E/F lint rule selection, and the Python 3.10 formatting target remain intact.

## Area and Change Type

- [x] Tests / harness / CI
- [x] Dependency maintenance and formatter compatibility fix

## Verification

Passed locally:

```text
uv sync --locked
uv run --offline --no-sync pytest -n auto --maxprocesses=4 --dist loadfile
uv run mypy
git diff --check origin/main...HEAD
```

Full offline suite: **8,494 passed, 50 skipped, 2 warnings** on Python 3.13.13 (481.99 seconds).

The harness passed all 38 scenarios; the route contract matched all 212 routes. The complete pre-commit gate passed after the compatibility fixes, and the commit hooks passed again. A stdlib AST comparison confirmed no executable change in any of the five Python files.

`bash scripts/container_smoke.sh` was attempted but its preflight stopped because Docker is unavailable on this host. Container runtime, Linux sandbox, browser, and the other Python-version jobs are delegated to this PR's CI; their results are separate from the local checks. A real SARIF upload runs only on the Scorecard workflow's configured triggers.

## Risk and Reviewer Focus

Review `Dockerfile`, `.pre-commit-config.yaml`, `.github/workflows/scorecard.yml`, and the comment-placement changes in `openai4s/kernel/worker.py` and `tests/test_retrosynthesis_scenario_protocols.py`.

Two independent reviews verified the official image digest, unchanged Python series and architecture coverage, the exact upstream hook/action tag-to-SHA mappings, and isort/Ruff compatibility. The formatter loop found during review is fixed. No unresolved actionable findings remain in the reviewed diff. This review does not substitute for the repository's required code-owner approval before merge.

## Checklist

- [x] Full diff reviewed; scope is the requested consolidated dependency update and necessary formatting compatibility.
- [x] No unrelated formatting or broad rewrite. Worker and Store facade edits are limited to imports/comments.
- [x] No frontend source/dist change; vendored assets and captured fixtures were not reformatted.
- [x] Full offline suite and lint run; existing kernel, agent, and gateway tests are included in the suite.
- [x] No tests deleted and no live-resource tests or credentials added.
- [x] No hard third-party core imports or runtime dependencies added.
- [x] No secrets, private data, local paths, or unpublished research information added.
- [x] Documentation and translated behavior claims remain accurate; no user-facing API or security guarantee changed.

## Review follow-up (4 commits)

A max-effort review of the diff found one real defect in the compatibility fix and three structural gaps the bump exposed.

**`6df4911a` — the `isort: skip` workaround was wrong in three ways.**
`# noqa: E402; isort: skip` is not a directive ruff can parse: the `;` makes it reject the whole thing, so it suppressed nothing and printed `warning: Invalid # noqa directive` on every cold-cache run — the only invalid noqa in the tree. It is not merely inert: with the code selected and `--fix` on, ruff *deletes* the line, and `ruff-check` runs with `--fix` and tree-wide write access. Beyond the spelling, `isort: skip` created a section boundary (an appended `import zlib` was pinned under the third-party group at line 82 and `isort --check-only` still exited 0) and only held under isort 9 (isort 8 moved the statement below `yield_benchmark`, out of sorted order — reachable by reverting `657ea805` alone, or from a warm hook cache).

Fixed by dropping the trailing comment rather than re-spelling the directive: the statement then collapses to 84 characters and both formatters leave it alone. No suppression is lost — ruff exempts imports following a `sys.path` mutation, so the file is clean with `--select E402`. Verified byte-stable over four isort→black round-trips under isort 9.0.1 **and** 8.0.1.

**`ab7e94fa` — `[tool.isort]`.** `[tool.black]` and `[tool.ruff.lint]` each argue that a formatter's behaviour must be stated, not inherited; isort was the one of the three with no such block, and this is its *major* bump. `src_paths` defaults to `(cwd/"src", cwd)`, so first-party detection was a property of the process:

```
with the block, cwd=repo root ....   0 of 883 files reported unsorted
with the block, cwd=elsewhere ...    0
without,        cwd=repo root ....   0
without,        cwd=elsewhere ...  317   <- `openai4s` demoted to third-party
```

Zero output change, per the precedent that block records.

**`f857e7d8` — dependabot.** `hook-versions` excluded black because "a black bump is a style change that has to be read on its own, not merged as one row of a version table". isort has the same property; it was not excluded, which is how a major arrived batched with a ruff patch.

**`76a53cf9` — corrections to the above.** A sweep pass found the new comments' *explanations* wrong even where the code was right: the formatter trigger is not length alone (with `combine_as_imports` off, only *aliased* imports are collapsed — isolated with two imports of identical 81-character bodies), the two files are not symmetric (the kernel worker's suppressions **are** load-bearing, because its insert sits inside an `if`; the test's are inert), and `[tool.isort]` overclaimed that a bare `isort .` "agrees with the gate" — it still rewrites 318 of 408 pinned `skills/bioskills/` files. One disputed claim was re-measured and stands: the oscillation is unbounded, not a one-pass convergence (six rounds, `isort→b317cd8aabca`, `black→31dc7b2429e1`, never equal).

### Re-verified after the fixes

`pre-commit run --all-files` passes with no file modified · `uv run mypy` · full offline suite **8,494 passed, 50 skipped** · `harness.cli --tier pr --offline` 38/38 · `capture_response_contract.py --check` 212/212 · `check_directory_readmes.py` · `source_secret_scan.py` · 0 invalid-noqa warnings across all 883 checked files · all five reformatted files still AST-identical to `origin/main`.

### Reported, deliberately not applied

`base-image`'s `update-types` filter never matches a digest-only Docker update (proven by `7f45a740`'s trailers, which carry no `update-type:` and no `dependency-group:`) — but the fix is a product decision, since a single docker dependency cannot be batched down. Nothing verifies a `# frozen:` comment against its SHA, nor that the Dockerfile digest belongs to the tag beside it (Docker ignores the tag when a digest is present); `combine_as_imports` would remove 113 duplicated module paths across 63 files but rewrites 67. All are inline comments on the PR.

---

> [!NOTE]
> **Low Risk**
> Pin bumps and import formatting only; kernel and store edits are comment/import layout with identical AST per PR verification.
> 
> **Overview**
> Consolidates dependency maintenance: **Python 3.14 slim-bookworm** digest updates in both **Dockerfile** stages, **isort 9.0.1** and **Ruff 0.16.6** frozen SHAs in **`.pre-commit-config.yaml`**, and **CodeQL `upload-sarif` v4.37.9** in the Scorecard workflow.
> 
> **isort 9** only changes how imports are written in five Python files (collapsed multi-line `from … import` blocks). **`worker.py`** keeps the trusted-path comment with shorter `# noqa: E402` lines; **`test_retrosynthesis_scenario_protocols.py`** adds an **`isort: skip`** on one late import to stop an isort/Black reorder loop. No runtime or API behavior is intended to change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8fa1b6bfc6630a12544f884de6b169830950a4b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -52,6 +52,13 @@ updates:
       # read on its own, not merged as one row of a version table (black
       # 23.3.0 -> 26.5.1 reformatted 234 files).
       #
+      # isort is out for the same reason, learned the same way: 8.0.1 -> 9.0.1
+      # arrived as one row of a two-row version table, reformatted five files --
+      # two of them the `store.py` and `kernel/worker.py` facades CLAUDE.md says
+      # to edit surgically -- and needed a hand-written directive to settle a
+      # disagreement it opened with black. It is the other formatter `setup.sh`
+      # installs on every contributor machine.
+      #
       # `exclude-patterns`, not `update-types`, because this ecosystem cannot
       # evaluate `update-types`. Dependabot resolves pre-commit hooks as git
       # refs and does not semver-classify them: every hook bump it has opened
@@ -63,7 +70,9 @@ updates:
       # it was meant to stop kept arriving and kept being consolidated by hand.
       hook-versions:
         patterns: ["*"]
-        exclude-patterns: ["https://github.com/psf/black"]
+        exclude-patterns:
+          - "https://github.com/psf/black"
+          - "https://github.com/pycqa/isort"
     open-pull-requests-limit: 3
 
   - package-ecosystem: "github-actions"
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -53,7 +53,7 @@ jobs:
       # `tests/test_ci_gate_independence.py` pins for ci.yml, which is scoped
       # to that file and cannot see this one.
       - name: Upload to code scanning
-        uses: github/codeql-action/upload-sarif@db488ddef3bf6cb639b32c2e9a7c0a7ea8271d28 # v4.37.8
+        uses: github/codeql-action/upload-sarif@cdf488f595d80d6e07e03d4674febd5ab45fa938 # v4.37.9
         with:
           sarif_file: results.sarif
       # `always()` so the run that failed is the one whose SARIF you can still
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       - id: check-merge-conflict
         args: ['--assume-in-merge']
   - repo: https://github.com/pycqa/isort
-    rev: a333737ed43df02b18e6c95477ea1b285b3de15a  # frozen: 8.0.1
+    rev: 131f4adcd5582bfc53928ab0d740eceb8b506b6c  # frozen: 9.0.1
     hooks:
       - id: isort
         args: ["--profile", "black"]
@@ -31,7 +31,7 @@ repos:
   # arrived and a bare `ruff` became ambiguous. The rule set is pinned in
   # pyproject.toml's [tool.ruff.lint]; see the comment there for why.
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: aab412d509121cb5f7533134b7e67f9fab59c682  # frozen: v0.16.4
+    rev: 321478e58f4938179c6b86e4ddfa923d1547a49b  # frozen: v0.16.6
     hooks:
       - id: ruff-check
         args: [--fix, "--ignore=E501,F401,E722,E402,E741"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@
 # `.github/workflows/ci.yml`'s offline matrix and in
 # `scripts/release_gates.py` CHECK_SUITE_GATES; `tests/test_platform_support.py`
 # reads this FROM line rather than restating the series.
-FROM python:3.14-slim-bookworm@sha256:416f0db2a2b561945630cef9877a7ea0581b27449eb9fd9df42f03e1b74b5b63 AS builder
+FROM python:3.14-slim-bookworm@sha256:9ab8d9c8514b44f90cf0029dd42fdd7e9e211e639c8b995304cc04568dee900f AS builder
 
 WORKDIR /src
 # The build backend first, in its own layer: it changes only when the pin in
@@ -49,7 +49,7 @@ RUN python -m pip wheel --no-cache-dir --no-deps --no-build-isolation \
         --wheel-dir /wheels /src
 
 # --- runtime stage -----------------------------------------------------------
-FROM python:3.14-slim-bookworm@sha256:416f0db2a2b561945630cef9877a7ea0581b27449eb9fd9df42f03e1b74b5b63 AS runtime
+FROM python:3.14-slim-bookworm@sha256:9ab8d9c8514b44f90cf0029dd42fdd7e9e211e639c8b995304cc04568dee900f AS runtime
 
 # `science` installs numpy/pandas/matplotlib/scikit-learn so agent cells can do
 # actual work. Pass `--build-arg OPENAI4S_EXTRAS=` for the stdlib-only control
--- a/openai4s/kernel/worker.py
+++ b/openai4s/kernel/worker.py
@@ -58,15 +58,12 @@ if (
 ):
     sys.path.insert(0, _TRUSTED_PACKAGE_PARENT)
 
-from openai4s.kernel.protocol import (  # noqa: E402 - trusted path fixed above
+# The trusted package path must be fixed before importing the protocol.
+from openai4s.kernel.protocol import (  # noqa: E402
     JSON_WORST_BYTES_PER_CHAR as _JSON_WORST_BYTES_PER_CHAR,
 )
-from openai4s.kernel.protocol import (  # noqa: E402 - trusted path fixed above
-    MAX_FRAME_BYTES as _MAX_FRAME_BYTES,
-)
-from openai4s.kernel.protocol import (  # noqa: E402 - trusted path fixed above
-    MAX_OUTPUT_CHARS as MAX_OUTPUT,
-)
+from openai4s.kernel.protocol import MAX_FRAME_BYTES as _MAX_FRAME_BYTES  # noqa: E402
+from openai4s.kernel.protocol import MAX_OUTPUT_CHARS as MAX_OUTPUT  # noqa: E402
 
 _DISCARD_BUDGET = 8  # bounded discard for desync
 _HOST_CALL_WIRE_CAP = 15_000_000  # 15MB host_call payload cap
--- a/openai4s/kernel/worker.py
+++ b/openai4s/kernel/worker.py
@@ -52,13 +52,17 @@ from functools import partial
 _TRUSTED_PACKAGE_PARENT = os.path.dirname(
     os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 )
+# This must run before the protocol import below: those imports are
+# deliberately not at the top of the file, which is what their per-line
+# suppression comments record. Keep those comments short -- isort sizes such a
+# statement without its trailing comment while Black sizes it with, so a long
+# one makes the two rewrap the line forever.
 if (
     os.path.isfile(os.path.join(_TRUSTED_PACKAGE_PARENT, "openai4s", "__init__.py"))
     and _TRUSTED_PACKAGE_PARENT not in sys.path
 ):
     sys.path.insert(0, _TRUSTED_PACKAGE_PARENT)
 
-# The trusted package path must be fixed before importing the protocol.
 from openai4s.kernel.protocol import (  # noqa: E402
     JSON_WORST_BYTES_PER_CHAR as _JSON_WORST_BYTES_PER_CHAR,
 )
--- a/openai4s/kernel/worker.py
+++ b/openai4s/kernel/worker.py
@@ -52,11 +52,20 @@ from functools import partial
 _TRUSTED_PACKAGE_PARENT = os.path.dirname(
     os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 )
-# This must run before the protocol import below: those imports are
-# deliberately not at the top of the file, which is what their per-line
-# suppression comments record. Keep those comments short -- isort sizes such a
-# statement without its trailing comment while Black sizes it with, so a long
-# one makes the two rewrap the line forever.
+# This must run before the protocol imports below, which is why they are not at
+# the top of the file and why each carries a suppression comment. Unlike the
+# retrosynthesis test's, these are load-bearing: the insert sits inside an `if`,
+# so ruff's allowance for imports after a sys.path mutation does not apply and
+# stripping them really does raise E402.
+#
+# Keep them short. `combine_as_imports` is off, so isort renders each aliased
+# name as its own one-line statement and then decides to collapse on the code
+# alone, while Black measures the same line with the comment -- put the two on
+# opposite sides of 88 and they rewrap it against each other forever. Only the
+# aliased imports here are exposed, and only while they stay collapsed:
+# `_MAX_FRAME_BYTES` is 72 characters of code, so its trailing comment may be 16
+# and is 14. The first import is already past 88 on code alone, so it stays
+# wrapped and is not exposed at all.
 if (
     os.path.isfile(os.path.join(_TRUSTED_PACKAGE_PARENT, "openai4s", "__init__.py"))
     and _TRUSTED_PACKAGE_PARENT not in sys.path
--- a/openai4s/orchestration/bootstrap.py
+++ b/openai4s/orchestration/bootstrap.py
@@ -49,10 +49,7 @@ from openai4s.security.permissions import (
     FILE_MODE,
 )
 from openai4s.security.permissions import fsync_dir as _fsync_dir
-from openai4s.security.permissions import (
-    harden_dir,
-    harden_file,
-)
+from openai4s.security.permissions import harden_dir, harden_file
 
 #: Filename of the per-daemon signing secret, under the data dir.
 SECRET_FILENAME = "worker-bootstrap-secret"
--- a/openai4s/store.py
+++ b/openai4s/store.py
@@ -128,9 +128,7 @@ from openai4s.storage.model_capability_receipts import (
 from openai4s.storage.permissions import (
     DEFAULT_PERMISSION_RULES as _DEFAULT_PERMISSION_RULES,
 )
-from openai4s.storage.permissions import (
-    PermissionRuleRepository,
-)
+from openai4s.storage.permissions import PermissionRuleRepository
 from openai4s.storage.permissions import perm_match as _perm_match
 from openai4s.storage.plans import PlanRepository
 from openai4s.storage.recovery import RecoveryJournalRepository
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -227,11 +227,19 @@ line-length = 88
 # in the repository where the other two blocks already keep theirs.
 #
 # `profile` restates the hook's `--profile black` so that a bare `isort .` -- the
-# default an editor integration runs -- agrees with the gate instead of emitting
-# a diff the gate then reverts. `py_version` states the union-across-versions
-# default explicitly, which is the correct one for a `requires-python = ">=3.10"`
-# package that CI runs on 3.10 through 3.14; narrowing it to the floor would
-# misfile a module that only became stdlib later.
+# default an editor integration runs -- lays imports out the way the gate does,
+# instead of emitting a diff the gate then reverts. It does not make that run
+# safe, and nothing here can: the byte-exact trees are guarded only by
+# pre-commit's own `exclude:`, so a bare `isort .` still rewrites 318 of the 408
+# pinned files under `skills/bioskills/`. Mirroring those paths into a `skip`
+# here would not reach the gate either -- the hook's `args:` replaces the
+# upstream manifest's `--filter-files`, and without it isort ignores `skip` for
+# files named on the command line, which is the only way pre-commit invokes it.
+#
+# `py_version` states the union-across-versions default explicitly, which is the
+# correct one for a `requires-python = ">=3.10"` package that CI runs on 3.10
+# through 3.14; narrowing it to the floor would misfile `tomllib`, which this
+# repository imports and which only became stdlib in 3.11.
 #
 # Like the black block, this changes no output: the tree is byte-identical with
 # and without it.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -211,6 +211,34 @@ markers = [
 target-version = ["py310"]
 line-length = 88
 
+[tool.isort]
+# Same reason as the block above and the one below: state the setting instead of
+# inheriting whatever the pinned formatter happens to compute. isort was the one
+# formatter of the three with no section here -- its entire configuration was
+# `--profile black` on the hook's command line -- so everything the profile does
+# not name was whatever the pinned version shipped, across an 8.0.1 -> 9.0.1
+# major that moved exactly those defaults.
+#
+# `src_paths` is the load-bearing entry. It defaults to `(cwd/"src", cwd)`, so
+# first-party detection was a property of the process rather than of the
+# repository: run isort from anywhere but the repo root and `openai4s` is
+# demoted to third-party, which moves the section breaks. A relative path in a
+# config file resolves against that file's directory, which puts the answer back
+# in the repository where the other two blocks already keep theirs.
+#
+# `profile` restates the hook's `--profile black` so that a bare `isort .` -- the
+# default an editor integration runs -- agrees with the gate instead of emitting
+# a diff the gate then reverts. `py_version` states the union-across-versions
+# default explicitly, which is the correct one for a `requires-python = ">=3.10"`
+# package that CI runs on 3.10 through 3.14; narrowing it to the floor would
+# misfile a module that only became stdlib later.
+#
+# Like the black block, this changes no output: the tree is byte-identical with
+# and without it.
+profile = "black"
+py_version = "3"
+src_paths = ["."]
+
 [tool.ruff.lint]
 # State the rule set instead of inheriting whatever the pinned ruff calls its
 # default. Until now there was no [tool.ruff] section at all, so the rule set
--- a/tests/test_retrosynthesis_scenario_protocols.py
+++ b/tests/test_retrosynthesis_scenario_protocols.py
@@ -74,7 +74,9 @@ from retrosynthesis_planning.reproducibility_bundle import (  # noqa: E402
     build_reproducibility_bundle,
 )
 from retrosynthesis_planning.route_review import route_similarity  # noqa: E402
-from retrosynthesis_planning.scenario_benchmark_cli import (  # noqa: E402
+
+# Keep the noqa here: isort 9 and Black otherwise move it back and forth.
+from retrosynthesis_planning.scenario_benchmark_cli import (  # noqa: E402; isort: skip
     main as scenario_cli_main,
 )
 from retrosynthesis_planning.yield_benchmark import (  # noqa: E402
--- a/tests/test_retrosynthesis_scenario_protocols.py
+++ b/tests/test_retrosynthesis_scenario_protocols.py
@@ -75,10 +75,12 @@ from retrosynthesis_planning.reproducibility_bundle import (  # noqa: E402
 )
 from retrosynthesis_planning.route_review import route_similarity  # noqa: E402
 
-# Keep the noqa here: isort 9 and Black otherwise move it back and forth.
-from retrosynthesis_planning.scenario_benchmark_cli import (  # noqa: E402; isort: skip
-    main as scenario_cli_main,
-)
+# This one import carries no trailing suppression comment, unlike its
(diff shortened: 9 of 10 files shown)


Nobody-Zhang (commented):
Reviewed the combined diff against `main`, including independent review of the container/CodeQL updates and the isort/Ruff upgrade.

One issue was found and fixed before this review: isort 9 and Black repeatedly rewrote `noqa` placement in two alias imports, so rerunning the original dependency change never produced a clean lint gate. The final change preserves the worker's trusted-path explanation and uses one documented import-local isort exception in the scenario test. The other required import edits are formatting only. All five Python files have identical ASTs before and after the update.

The official Python image index matches the pinned digest and preserves the Python series and architecture coverage. The isort, Ruff, and CodeQL version tags resolve to the exact frozen commits. The configured Ruff hook does not adopt Markdown formatting, and the existing E/F rule set stays unchanged.

Validation: **8,494 passed, 50 skipped** in the full offline suite on Python 3.13.13; complete pre-commit, mypy, 38 harness scenarios, 212-route contract check, bilingual directory coverage, source credential scan, and diff whitespace checks all passed.

No unresolved actionable findings in the final diff. The container smoke could not run locally because Docker is unavailable; this PR's hosted CI must supply that runtime evidence, together with Linux/browser and other Python-version checks. Real SARIF upload is limited to the Scorecard workflow's configured triggers. This is a code-review comment, not code-owner approval or authorization to merge.

tests/test_retrosynthesis_scenario_protocols.py
```diff
@@ -74,9 +74,13 @@
...
-from retrosynthesis_planning.scenario_benchmark_cli import (  # noqa: E402
-    main as scenario_cli_main,
-)
+
+# This one import carries no trailing suppression comment, unlike its
+# neighbours. isort sizes the collapsed statement at 84 characters because it
+# does not count a trailing comment, while Black sizes the line it emits at 98
+# and wraps it again -- so the pair never reaches a fixed point. None is needed
+# here: ruff exempts imports that follow the sys.path mutation above.
+from retrosynthesis_planning.scenario_benchmark_cli import main as scenario_cli_main
```
Nobody-Zhang, line 91: **1. `# noqa: E402; isort: skip` was not a directive ruff can parse — it suppressed nothing.** *(fixed in 6df4911a)*

The `;` makes ruff reject the whole directive rather than terminate the code list:

```
warning: Invalid `# noqa` directive on tests/test_retrosynthesis_scenario_protocols.py:79:
         expected code to consist of uppercase letters followed by digits only (e.g. `F401`)
```

A/B on a file where E402 actually fires — `# noqa: E402` → `All checks passed!`; `# noqa: E402; isort: skip` → **E402 reported**. It was the only invalid noqa in the tree (origin/main: 0, this PR: 1). Three things hid it: E402 is on the hook's `--ignore`, ruff exits 0, and pre-commit swallows a passing hook's output.

It is not merely inert. With the code selected and `--fix` on — and `ruff-check` runs with `--fix` and write access to the whole tree — ruff **deletes the line**:

```
before: from json import loads as _loads  # noqa: F401; isort: skip
        from json import dumps as _dumps  # noqa: F401  # isort: skip
$ ruff check --isolated --select F401 --fix fix.py   → "1 fixed"
after:  from json import dumps as _dumps  # noqa: F401  # isort: skip
```

The semicolon form was deleted; the conventional form survived. Note the repo's existing `# noqa: E402 - trusted path fixed above` and `# noqa: BLE001 — …` spellings parse fine — whitespace and the em dash terminate the list cleanly, `;` uniquely does not.

Nobody-Zhang, line 91: **2. `isort: skip` made this a section boundary, and the gate then blessed the mis-sorted result.** *(fixed in 6df4911a)*

isort 9 treats a skipped statement as a section break. Appending `import zlib` below it:

| | landing line | `isort --check-only` |
|---|---|---|
| with the skip | **82** — under the third-party group | exit 0 ✅ |
| without it | **18** — the stdlib group | exit 0 ✅ |

So a future contributor's stdlib import gets pinned into a second, wrongly-placed group and isort certifies it as sorted — a gate answering an easier question than the one it appears to answer.

Worth recording: the import *ordering* that actually matters in this file is protected by the `sys.path.insert` at line 16, which isort never sorts across. I verified nothing moves above it either way. The skip was purely a formatter-loop device, not an ordering guard.

tests/test_retrosynthesis_scenario_protocols.py
```diff
@@ -74,9 +74,13 @@
     build_reproducibility_bundle,
 )
 from retrosynthesis_planning.route_review import route_similarity  # noqa: E402
-from retrosynthesis_planning.scenario_benchmark_cli import (  # noqa: E402
-    main as scenario_cli_main,
-)
+
+# This one import carries no trailing suppression comment, unlike its
```
Nobody-Zhang: **3. The comment named the wrong mechanism, and its premise was wrong.** *(fixed in 6df4911a)*

> `# Keep the noqa here: isort 9 and Black otherwise move it back and forth.`

The noqa is the **cause**, not the cure — `isort: skip` was the load-bearing half. Measured: noqa-only oscillates; `isort: skip` alone converges; no directive at all converges. "Keep the noqa here" reads as an instruction to preserve the noqa's position, which is precisely the framing that produced the spliced directive in finding 1.

The root cause is a measurement disagreement: **isort sizes the collapsed statement at 84 characters because it does not count a trailing comment; Black sizes the line it emits at 98 and wraps it again.** (ruff 0.16.6 shipped the same class of fix upstream — *"[isort] Exclude pragma comments from line length calculation (I001)"*.)

The replacement comment now states that threshold, so the next person can see why the line is shaped as it is.

tests/test_retrosynthesis_scenario_protocols.py
Nobody-Zhang, line 91: **4. The `isort: skip` remedy only held under isort 9; a partial revert silently unsorted the block.** *(fixed in 6df4911a)*

The pin bump (`657ea805`) and the reformat (`f8fa1b6b`) are separate commits, so reverting the pin alone is easy — and a contributor with a warm `~/.cache/pre-commit` is in the same position.

| file | isort 9.0.1 | black | isort 8.0.1 |
|---|---|---|---|
| origin/main | MODIFY | clean | clean |
| this PR (before fix) | clean | clean | **MODIFY** |

Under isort 8 the whole skipped statement was moved below `yield_benchmark`, out of sorted order.

The fix drops the trailing comment instead of re-spelling the directive, so the statement collapses to 84 characters and both formatters leave it alone. Verified byte-stable over four isort→black round-trips under **both** isort 9.0.1 and isort 8.0.1, and clean with `--select E402` (ruff exempts imports that follow a `sys.path` mutation, so no suppression is lost).

openai4s/kernel/worker.py
```diff
@@ -52,21 +52,22 @@
...
+from openai4s.kernel.protocol import (  # noqa: E402
     JSON_WORST_BYTES_PER_CHAR as _JSON_WORST_BYTES_PER_CHAR,
 )
-from openai4s.kernel.protocol import (  # noqa: E402 - trusted path fixed above
-    MAX_FRAME_BYTES as _MAX_FRAME_BYTES,
-)
-from openai4s.kernel.protocol import (  # noqa: E402 - trusted path fixed above
-    MAX_OUTPUT_CHARS as MAX_OUTPUT,
-)
+from openai4s.kernel.protocol import MAX_FRAME_BYTES as _MAX_FRAME_BYTES  # noqa: E402
```
Nobody-Zhang, line 78: **5. This line sits 2 characters from the band where the two formatters stop converging.** *(documented in 6df4911a; the structural fix is out of scope)*

Same mechanism as finding 3, swept across lengths for a 14-char trailing comment:

```
code-only 76,78,80,82,84,86,88 -> *** PING-PONG ***
code-only 89,90                -> converged
```

Band = code-only ∈ (74, 88]. This line's code-only length is **72** — headroom 2. Mutation-verified: a 3-character alias rename puts it back into ping-pong, and the state Black leaves behind parks `# noqa: E402` on the closing paren, where **ruff does not honour it** (E402 is reported at the statement's opening line). So the churned form kills the suppression silently too.

That makes this PR's shortening of the noqa text load-bearing and, as merged, uncommented. `6df4911a` records the constraint on the guard above. The real fix is `combine_as_imports` (see the comment on `pyproject.toml`), which collapses these three statements into one and removes the trailing comments entirely — 67 files repo-wide, so it belongs in its own PR.

openai4s/kernel/worker.py
```diff
@@ -52,21 +52,22 @@
 _TRUSTED_PACKAGE_PARENT = os.path.dirname(
     os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 )
+# This must run before the protocol import below: those imports are
```
Nobody-Zhang: **6. The relocated comment was hoistable; an added import would strand it.** *(fixed in 6df4911a)*

The PR replaced three per-line `# noqa: E402 - trusted path fixed above` justifications with one floating comment directly above the first import. Attached to an import, isort moves it: adding `from openai4s.kernel.aaa_new import THING  # noqa: E402` to that block hoists the new import **above** the comment, leaving an unexplained `sys.path`-dependent import at the top and the explanation three lines down.

Re-anchored to the `if` guard, where isort never sorts, so it cannot be separated from what it describes. Nothing in the repo referenced the removed text (`grep -rn "trusted path fixed above"` → 0 hits), so no gate was affected — this is readability plus the length constraint from finding 5.

Related and **not** fixed here: nothing tests that these three imports stay *below* the `sys.path.insert`. A future `float_to_top` or formatter bump could hoist them, lint would stay green, and the remote source-checkout worker would lose its trusted-path hardening — failing only at runtime.

pyproject.toml
```diff
@@ -211,6 +211,34 @@ markers = [
...
+# a diff the gate then reverts. `py_version` states the union-across-versions
+# default explicitly, which is the correct one for a `requires-python = ">=3.10"`
+# package that CI runs on 3.10 through 3.14; narrowing it to the floor would
+# misfile a module that only became stdlib later.
+#
+# Like the black block, this changes no output: the tree is byte-identical with
+# and without it.
+profile = "black"
+py_version = "3"
+src_paths = ["."]
```
Nobody-Zhang, line 248: **7. isort was the one formatter of the three whose behaviour was inherited rather than stated.** *(fixed in ab7e94fa)*

`[tool.black]` and `[tool.ruff.lint]` each argue the same point at length — and this PR is a **major** isort bump.

> "State the setting instead of inheriting whatever the pinned formatter happens to compute … not a side effect of Dependabot moving a pin."

There was no `[tool.isort]`, no `.isort.cfg`, no `setup.cfg`. The entire configuration was `--profile black` on the hook line, which names seven keys; the rest was whatever the pinned version shipped.

`src_paths` is the load-bearing entry, and it is the same defect the black block describes — it defaults to `(cwd/"src", cwd)`, so first-party detection was a property of the *process*, not the repository. Measured over the 883 files the hooks check:

```
with the block, cwd=repo root ....   0 files reported unsorted
with the block, cwd=elsewhere ...    0
without,        cwd=repo root ....   0
without,        cwd=elsewhere ...  317   <- `openai4s` demoted to third-party
```

A relative path in a config file resolves against that file's directory, which puts the answer back in the repo. `py_version` states the union-across-versions default explicitly rather than the `requires-python` floor — narrowing it to 3.10 would misfile `tomllib`, which this repo imports and which only became stdlib in 3.11.

Zero-output-change, as the black block was: `pre-commit run --all-files` passes with no file modified. Landed as its own commit, per the precedent that block records.

.github/dependabot.yml
```diff
@@ -63,7 +70,9 @@ updates:
       # it was meant to stop kept arriving and kept being consolidated by hand.
       hook-versions:
         patterns: ["*"]
-        exclude-patterns: ["https://github.com/psf/black"]
+        exclude-patterns:
+          - "https://github.com/psf/black"
+          - "https://github.com/pycqa/isort"
```
Nobody-Zhang, line 75: **8. The group excluded black for a reason that applies identically to isort.** *(fixed in f857e7d8)*

> "deliberately keep black out of the group: a black bump is a style change that has to be read on its own, **not merged as one row of a version table**"

isort has the same property and was not excluded, so 8.0.1 → 9.0.1 — a major — arrived batched with a ruff patch. Its reviewable diff was two SHA lines; what it actually did was reformat five files, two of them the `store.py` and `kernel/worker.py` facades CLAUDE.md says to edit surgically, and open a formatter disagreement that needed a hand-written directive to settle (findings 1–4).

The `workflow-actions` group below already records this lesson from the other direction — *"a major went through as one row of a two-row version table"* — and fixed it with `update-types`. This ecosystem cannot express `update-types`, as the comment above explains, so `exclude-patterns` is the only lever.

Note the sharper version of the risk: isort 9.0.0's *"Consider private stdlib modules to be stdlib"* would have **re-sectioned imports with no diff in the chore PR at all**, then reddened the next unrelated PR that touched those files. (This repo has no `import _xxx`, so it did not fire here — I checked.)

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
       - id: check-merge-conflict
         args: ['--assume-in-merge']
   - repo: https://github.com/pycqa/isort
-    rev: a333737ed43df02b18e6c95477ea1b285b3de15a  # frozen: 8.0.1
+    rev: 131f4adcd5582bfc53928ab0d740eceb8b506b6c  # frozen: 9.0.1
```
Nobody-Zhang, line 21: **9. Nothing verifies a `# frozen:` comment against its SHA — while Action pins get two gates.** *(not fixed — needs its own PR)*

- `.github/workflows/*` → `tests/test_governance.py` globs every workflow for the SHA+comment shape, **and** `ci.yml:72` runs `pinact-action` with `verify: "true"`, which dereferences each tag through the API.
- `.pre-commit-config.yaml` → nothing. `grep -rn "pre-commit-config" tests/ scripts/ harness/` returns no matches; pinact's file set is workflows only, and it parses `uses:`, not `rev:`.

Yet this file's own header (lines 1–9) argues these pins matter **more**: *"none of these tools is in `uv.lock` … `ruff-check` runs with `--fix` (write access to the whole working tree) … a re-pointed tag would rewrite source on developer machines, not just fail a CI check."*

Aggravating: the frozen envs cannot self-identify — `isort --version` inside the installed hook env reports `VERSION 0.1.dev1` (shallow clone, no setuptools-scm tag). The SHA in this file is the *only* identity record, and it is unchecked.

All three bumped pins in this PR **do** verify, by hand: isort `9.0.1` → `131f4adc…`, ruff-pre-commit `v0.16.6` → `321478e5…`, codeql-action `v4.37.9` → annotated tag `a35ac6e6…` → `cdf488f5…`. Not fixed here because the real check needs network and a new gate.

Dockerfile
```diff
@@ -35,7 +35,7 @@
 # `.github/workflows/ci.yml`'s offline matrix and in
 # `scripts/release_gates.py` CHECK_SUITE_GATES; `tests/test_platform_support.py`
 # reads this FROM line rather than restating the series.
-FROM python:3.14-slim-bookworm@sha256:416f0db2a2b561945630cef9877a7ea0581b27449eb9fd9df42f03e1b74b5b63 AS builder
+FROM python:3.14-slim-bookworm@sha256:9ab8d9c8514b44f90cf0029dd42fdd7e9e211e639c8b995304cc04568dee900f AS builder
```
Nobody-Zhang, line 38: **10. Nothing checks that this digest belongs to `python:3.14-slim-bookworm`.** *(not fixed — needs its own PR)*

When a digest is present Docker resolves by digest and **ignores the tag text** — so the half the tests assert is the decorative half:

- `tests/test_container_deployment.py:103` → `re.fullmatch(r"python:3\.14-slim-bookworm@sha256:[0-9a-f]{64}", …)` — tag text + digest *shape*.
- `tests/test_platform_support.py:211` → `^FROM python:(\d+\.\d+)-` — tag text only.

So "the container ships 3.14", which the release-gate manifest renders job names from, rests on a string that Docker never reads. This is exactly the identity check pinact provides for Actions, absent for the base image. Only `container_smoke.sh:154` reconciles the series at runtime, in a separate Docker-requiring job.

**This PR's digest is correct** — verified against the registry, and the PR description's architecture claim holds:

```
tag python:3.14-slim-bookworm  docker-content-digest: sha256:9ab8d9c8…dee900f   (== the new pin)
mediaType: oci.image.index.v1+json — linux/amd64, arm/v7, arm64/v8, 386, ppc64le  (identical set to 416f0db2)
PYTHON_VERSION=3.14.7, PYTHON_SHA256 identical, same PATH/Cmd/User, same Debian snapshot
only behavioural delta: CXX/LDCXXSHARED gcc -> g++ (2 of 1105 sysconfig vars; neither compiler survives into runtime)
```

.pre-commit-config.yaml
```diff
@@ -31,7 +31,7 @@ repos:
   # arrived and a bare `ruff` became ambiguous. The rule set is pinned in
   # pyproject.toml's [tool.ruff.lint]; see the comment there for why.
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: aab412d509121cb5f7533134b7e67f9fab59c682  # frozen: v0.16.4
+    rev: 321478e58f4938179c6b86e4ddfa923d1547a49b  # frozen: v0.16.6
```
Nobody-Zhang, line 34: **11. The "never reformat the pinned trees" guarantee has exactly one call site.** *(not fixed — needs its own PR)*

`exclude:` on line 55 is the only place it lives. There is no `[tool.ruff] exclude`, no `[tool.black] force-exclude`, no `[tool.isort] skip` — so the tools themselves do not refuse:

```
$ ruff check --ignore=E501,F401,E722,E402,E741 --statistics skills/bioskills/
92 E702 · 62 F541 · 7 F841 · 1 E401 → Found 162 errors.  [*] 63 fixable with --fix
$ black --check skills/bioskills/   -> 408 files would be reformatted.
```

This is reachable from ordinary review work: validating this very bump by running `ruff check --fix` at repo scope rewrote **34 pinned upstream files / 64 lines** before they were restored. `pre-commit run --all-files` is blind to it (it excludes the tree) and so is the docs gate; it is caught only by `uv build` + `verify_release_artifacts`, as a payload-digest mismatch rather than "you reformatted a pinned tree", and only for files that ship.

Given this PR moves two formatter pins, adding the exclusion to the three tool configs would put the refusal where the tool can act on it.

openai4s/orchestration/bootstrap.py
```diff
@@ -49,10 +49,7 @@
     FILE_MODE,
```
Nobody-Zhang, line 49: **12. Leftover three-line block for a 51-character name, directly above the sibling this PR just collapsed.** *(not applied — see below)*

```python
from openai4s.security.permissions import (
    FILE_MODE,
)
from openai4s.security.permissions import fsync_dir as _fsync_dir
from openai4s.security.permissions import harden_dir, harden_file   # <- collapsed by this PR
```

`from openai4s.security.permissions import FILE_MODE` is 51 chars against a limit of 88, and the collapsed form is stable under isort 9.0.1 + black 26.5.1 (verified). isort leaves it split only because the magic trailing comma tells it to.

Not applied: it is formatting this PR did not otherwise touch, and the PR checklist states "No unrelated formatting or broad rewrite." Worth a follow-up. (`store.py:128-130` genuinely cannot collapse — that one-liner would be 94 chars.)

openai4s/store.py
```diff
@@ -128,9 +128,7 @@
 from openai4s.storage.permissions import (
     DEFAULT_PERMISSION_RULES as _DEFAULT_PERMISSION_RULES,
 )
-from openai4s.storage.permissions import (
-    PermissionRuleRepository,
-)
+from openai4s.storage.permissions import PermissionRuleRepository
```
Nobody-Zhang, line 131: **13. `combine_as_imports` is the root-cause fix for the duplicated module paths.** *(not applied — 67 files, own PR)*

These repeated `from openai4s.storage.permissions import …` statements exist *only* because isort's default splits aliased names into their own statement. With `combine_as_imports`, verified against the pinned tools:

```diff
 from openai4s.kernel.protocol import (  # noqa: E402
     JSON_WORST_BYTES_PER_CHAR as _JSON_WORST_BYTES_PER_CHAR,
+    MAX_FRAME_BYTES as _MAX_FRAME_BYTES,
+    MAX_OUTPUT_CHARS as MAX_OUTPUT,
 )
-from openai4s.kernel.protocol import MAX_FRAME_BYTES as _MAX_FRAME_BYTES  # noqa: E402
-from openai4s.kernel.protocol import MAX_OUTPUT_CHARS as MAX_OUTPUT  # noqa: E402
```

It removes **113 redundant module-path restatements across 63 files** (`gateway.py` 12, `host_dispatch.py` 6, `store.py` 4), makes the retrosynthesis import stable with a plain trailing comment and no directive at all, and closes the length band in finding 5 at its source rather than per-site.

Honest cost: it rewrites **67 files**, so by the same rule this repo applies to black and to ruff rule families, it belongs in its own PR where the diff can be read.

.github/workflows/scorecard.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       # `tests/test_ci_gate_independence.py` pins for ci.yml, which is scoped
       # to that file and cannot see this one.
       - name: Upload to code scanning
-        uses: github/codeql-action/upload-sarif@db488ddef3bf6cb639b32c2e9a7c0a7ea8271d28 # v4.37.8
+        uses: github/codeql-action/upload-sarif@cdf488f595d80d6e07e03d4674febd5ab45fa938 # v4.37.9
```
Nobody-Zhang, line 56: **14. This file's pins and step ordering are ungated, as its own comment admits.** *(not fixed — own PR)*

The comment three lines above the changed line states the invariant — the SARIF upload must precede the artifact upload so an artifact 5xx cannot skip code scanning — and then says the enforcing test *"is scoped to that file and cannot see this one."* `tests/test_ci_gate_independence.py:31` hard-codes `CI = …/ci.yml`.

So a future edit reordering these two steps leaves the Security tab holding stale Scorecard alerts with nothing red. The repo's own canonical failure shape: the right guard wired to one call site of several.

The bump itself is clean — `codeql-action` appears exactly once in the tree, so no `init`/`analyze` version skew, and `v4.37.9`'s annotated tag does dereference to `cdf488f5…`.

.pre-commit-config.yaml
Nobody-Zhang, line 1: **15. `CLAUDE.md` documents three excludes; this line has four.** *(pre-existing, not fixed)*

CLAUDE.md:117 — *"pre-commit excludes `openai4s/server/webui/vendor/`, `skills/bioskills/`, and `tests/fixtures/` — never reformat those"* — omits `openai4s/server/webui/dist/`.

Confirmed pre-existing on `origin/main`, so this PR did not introduce it. Flagging because a review of a formatter-pin change is exactly when someone audits this list against the docs and concludes `dist/` is reformattable when it is not.

.github/dependabot.yml
Nobody-Zhang, line 1: **16. The `base-image` group's `update-types` filter matches nothing, so the group has never batched anything.** *(not applied — the right fix is the maintainer's call)*

This is the identical defect the same file documents at length one block below, for `hook-versions`:

> "`exclude-patterns`, not `update-types`, because this ecosystem cannot evaluate `update-types`. Dependabot resolves pre-commit hooks as git refs and does not semver-classify them … A group filtered only on `update-types` therefore matched nothing, and the group batched zero updates from the day it was added."

The same is true of a digest-only Docker update, and this PR's own commits prove it:

```
7f45a740 (docker)          dependency-name: python
                           dependency-version: 3.14-slim-bookworm
                           <- no update-type:, no dependency-group:
8cb122e2 (github-actions)  update-type: version-update:semver-patch
                           dependency-group: workflow-actions
657ea805 (pre-commit)      dependency-group: hook-versions
```

`7f45a740` carries neither trailer, so `update-types: ["minor", "patch"]` never matched it and the group is inert — while the comment above it says the group exists because "upstream rebuilds it for security patches roughly weekly, so this would otherwise open a PR per rebuild."

Not applied, for two reasons worth stating rather than glossing. First, removing the filter changes what Dependabot does, and the series guard is elsewhere anyway — `tests/test_container_deployment.py:103` fullmatches `python:3\.14-slim-bookworm@…`, and its own comment says naming the minor there "is what makes such a bump arrive as a red test a human has to look at." Second, with a single docker dependency, grouping cannot reduce the PR count at all, so the stated rationale may not survive contact with the mechanism — that is a product decision, not a review fix.

`tests/test_governance.py` validates dependabot key *placement* only, so nothing catches either case.

PR #156, opened 2026-09-10
fix: preserve historical artifacts and harden research execution

Nobody-Zhang:
## Summary

Completed Notebook cells now retain the exact artifact versions they produced, including figures, tables, downloads, environment metadata and lineage. Remote folding/scoring rejects invalid inputs and malformed or failed results, while failed history reads remain visible and retryable without replaying execution.

This integrates six staged commits from `next`:

| Area | Result |
| --- | --- |
| Historical artifacts | Project visible producer/capture relationships into `output_artifacts`; keep fixed and latest views separate across tabs, events, caches and asynchronous reads. Explicit version reads never fall back to latest. |
| Remote science | Validate sequences, numeric arguments, positions, exit status, JSON/base64, PDB coordinates/completeness and both existing CSV formats before recording successful provenance. Preserve the single-key error contract and avoid invented metadata. |
| History recovery | Track messages, steps and run state separately by conversation/open generation; preserve confirmed content, expose retryable partial/error states, merge live updates and clear replay gaps only after complete recovery. |
| Child collection | Apply one call-wide timeout budget and cancellation probes through Agent and Web runners, preserving child ownership, status and result identity. |
| Diagnostics | Display each check's detail/remedy and bounded facts, link only to known settings pages, retain prior results and invalidate them after configuration saves. |
| Database startup | Reject future schemas before application initialization writes, including WAL-visible versions and the formal-connection recheck; report explicit foreground/detached startup failures. |
| JSON tables | Use the full first-seen field union, preserve mixed JSON as text, and provide bounded large-text previews with local expansion and complete downloads. |

The source/dist pair, response shapes, browser contracts and bilingual documentation are synchronized. Chromium CI now installs the existing locked science extra for its real matplotlib fixture; the product scenes and assertions remain intact.

The npm installation guidance now points to the published [`@pku-yuangroup/openai4s-skills@0.2.0`](https://www.npmjs.com/package/@pku-yuangroup/openai4s-skills/v/0.2.0). That package was built from released `v0.2.0` with organization-name/documentation adaptation and unchanged Skill payloads: **603 Skills**, versus **604** in the current repository. The new single-cell recipe explicitly retains its GitHub installation path. Most of the additional Skill README changes are generated Install-section updates.

Retrosynthesis source candidates and license metadata are documented, but all six production datasets remain `not_frozen` pending maintainer admission and independent truth-file decisions. Dependabot configuration is a separate PR (#155); this PR only records its status. The local changes to `docs/next-version-progress.md` are excluded.

## Area

- [x] Web UI / frontend
- [x] Kernel / host RPC / runtime compatibility
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Feature
- [x] Test / harness
- [x] Documentation
- [x] Security-related change
- [x] Release / packaging guidance

## Verification

[Complete CI at implementation/documentation commit `1b56dc100c142f48c5f8d45631720e7eb6592052`](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34443430003): **25 jobs passed, 4 event-conditional skips, no failures**, first attempt. This includes Python 3.10/3.12/3.13/3.14, all three browser engines, container smoke, Linux interrupt/full sandbox, package installation, types, lint, docs and response contracts. The [Chromium job](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34443430003/job/102763034866) passed 653 frontend tests, the new history/artifact/JSON/diagnostics product scenes, admission-fault and sandbox-preview checks, and the 10/10 browser matrix. [Response capture](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34443430003/job/102763034845) covered 1,165 shapes and 212/212 routes without breaking changes.

Head `69d40ef12ccc625f937220fa48b15d97ff37a3e7` only updates the two TODO evidence documents after that run. Its [own CI run](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34444512496) was still running when this PR was prepared; the earlier green run is not presented as testing those final two documentation edits.

Local commands run successfully during implementation and final validation:

```text
uv run pytest tests/test_kernel.py tests/test_agent.py tests/test_gateway.py
uv run pytest tests/test_host_remote_science_service.py
uv run mypy
npm test --prefix frontend
node scripts/extract_webui_contract.mjs --check
node tests/browser_smoke.mjs
node tests/browser_sandbox_preview.mjs
node tools/skills-installer/check_package.mjs
python3 scripts/render_skill_install_sections.py --check
```

- Final local full suite: **8,767 passed / 33 skipped**. Explicit kernel/agent/gateway regression: **254 passed**; remote-science suite: **207 passed**. The final focused follow-up suite passed 112 tests; after the last documentation-only edit, 34 documentation/installer/crosswalk tests and commit hooks passed.
- Harness: **38 passed**. Installer self-tests: **16 passed**. Directory documentation and secret scans passed.
- Wheel/sdist resource verification and a fresh, repository-external `pip --no-index --no-deps` installation passed. The official import/CLI smoke found **13 modules / 604 Skills**. Final packaging ran in an isolated source snapshot with file hashes checked against the working tree.
- The public npm tarball matches the verified candidate hash. A fresh cache outside any checkout successfully listed 42 curated + 561 bioSkills and installed/uninstalled a recipe in a temporary directory. Anonymous registry access was verified.

Not run: live LLM, SSH/GPU prediction-quality or laboratory jobs; production dataset freezing. These require external resources or maintainer decisions and are not claims of the offline/browser gates. No Python release publication or tag/version change is included.

## Risk and Reviewer Focus

- **Artifact ownership:** `execution_sources.py`, `execution_views.py`, exact-version Gateway reads and frontend cache/Notebook/viewer paths. Check visible branch/frame identity, deduplicated captures and the absence of latest fallbacks.
- **Result authenticity:** `host/remote_science.py` and SDK compatibility. Check both CSV contracts, completeness checks, input rejection before SSH and provenance only after successful validation.
- **Recovery ordering:** message fetch/open/session and WS handlers. Check stale generations, unidentified streams, GET-only retry and complete recovery before gap clearing.
- **Cancellation and startup:** delegation/Agent/Web probes and migration/Store/CLI ordering. Check one deadline, subtree ownership, no lock-held waits, WAL visibility and rejection before application writes.
- **Product behavior:** bounded diagnostics/JSON rendering, safe text, download identity, and committed frontend resource consistency.

Code-owner review is required before merge, including security-focused ownership for the Store startup changes.

## Checklist

- [x] The staged implementation and integrated diff were reviewed, including independent reviews of functions by reviewers who did not write them.
- [ ] This PR is small and single-purpose. This is the requested `next` integration PR; six commits preserve the backend, frontend, documentation, CI and npm/follow-up stages. The large mechanical portion is generated Install documentation and built frontend assets.
- [x] No unrelated formatting, broad core rewrite or drive-by refactor is included.
- [x] Gateway, Store and SDK facade changes are local; kernel worker/manager protocol files were not changed.
- [x] Web features are in `frontend/`, with the rebuilt committed `openai4s/server/webui/dist/`; legacy `app.js` is unchanged.
- [x] Vendored renderer files and fixed fixtures were not reformatted; tests were not deleted to obtain a pass.
- [x] Offline tests require no live resources or real credentials; protocol and security-sensitive paths have focused failure coverage.
- [x] No new core runtime dependency, database table or schema migration is added; optional scientific dependencies retain their existing boundary.
- [x] No secrets, private data, unpublished roadmap or local machine paths are added to this PR description; source secret checks passed.
- [x] Documentation reflects tested behavior, remaining external decisions stay explicit, and affected English/Chinese pages are synchronized.

---

> [!NOTE]
> **Medium Risk**
> Changes span artifact versioning, provenance/metadata reads, and history reconciliation paths where incorrect fallbacks could mislead users; scope is mostly frontend and contracts with extensive tests but touches core notebook and viewer behavior.
> 
> **Overview**
> This PR tightens **immutable artifact handling** in the workbench: pinned versions get their own tabs and byte URLs, lineage/environment/metadata reads use explicit `version` query parameters with no silent latest fallback, and Notebook cells bind **confirmed** `output_artifacts` to the producing cell instead of guessing from live state.
> 
> **Conversation history** is loaded and recovered per open generation (messages, steps, run state tracked separately). Failed or partial reads surface a dedicated status bar with GET-only retry instead of an empty welcome state, and WebSocket gap handling waits for stable reconciliation.
> 
> **Customize Diagnostics** is reworked to run checks on demand with bounded timeouts, validate API responses, render per-check detail/remedy/facts safely, keep prior results across tab switches, and mark results stale after configuration writes.
> 
> **Artifact viewing** improves JSON tables (full key union, mixed arrays as raw text, large-text preview + local expand) and adds related copy, tests, and documented API/response-schema updates. **CI** installs the locked `science` extra on Chromium for the matplotlib browser fixture.
> 
> **Docs** document the published `@pku-yuangroup/openai4s-skills@0.2.0` install path, stricter remote fold/score validation behavior, and updated TODO/release-validation notes; `window` exports add `openKetcher` / `openViewer`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69d40ef12ccc625f937220fa48b15d97ff37a3e7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openai4s/storage/migrations.py
```diff
@@ -73,6 +73,45 @@ class MigrationError(RuntimeError):
...
+def preflight_schema(db_path: Path, *, target: int = SCHEMA_VERSION) -> int | None:
+    """Read committed schema state (including WAL) before application writes.
+
+    A missing database follows the normal new-database path. A corrupt or
+    unreadable database retains SQLite's error; it is not a future version.
+    """
+    path = Path(db_path)
+    if not path.exists():
+        return None
+    conn = sqlite3.connect(path.resolve().as_uri() + "?mode=ro", uri=True)
```
Nobody-Zhang, line 116: **Read-only preflight fails on a hot rollback journal (High)**

`preflight_schema` opens the DB with `?mode=ro` *before* the read-write connection. The Store keeps the rollback-journal default, so a process killed mid-commit (journal header synced, pages spilled) leaves a hot journal that only a read-write handle may replay; SQLite answers the read-only handle with `SQLITE_READONLY_ROLLBACK` → `sqlite3.OperationalError: attempt to write a readonly database`. Only `FutureSchemaError` is caught, so every `Store()` open (`serve`, `run`, `diagnostics`, tests) raises a traceback after an unclean shutdown and nothing ever performs the replay the pre-PR read-write open did.

Reproduced with a child that spills an UPDATE (`PRAGMA cache_size=-8`) and `os._exit`s: `preflight_schema` raises, `sqlite3.connect(path)` replays and reads `user_version` fine.

Fix applied on this branch: the preflight returns `None` on a read-only refusal and defers to the formal connection's `require_supported_schema` (which still runs before any DDL, so a future schema is still refused); regression tests cover both the replay and the future-schema-with-hot-journal case.

openai4s/cli/main.py
```diff
@@ -651,6 +672,11 @@ def _cmd_serve_detached(args, cfg) -> int:
 def cmd_serve(args) -> int:
     from openai4s.server import build_server, run_server
 
+    try:
+        preflight_schema(get_config(initialize_dirs=False).db_path)
```
Nobody-Zhang, line 676: **`FutureSchemaError` escapes as a traceback from `run` / `init` / `user`**

`cmd_serve` renders the refusal (`error: [future_schema] …`, exit 2), but `cmd_run` (`Agent(cfg=…)` → `build_dispatcher` → `get_store`), `cmd_init` (`_onboarding_service`) and `cmd_user` (`_team_store`) open the Store outside their `try` blocks or catch only `EnvironmentReadinessError` / `ValueError`. With `PRAGMA user_version=SCHEMA_VERSION+1` each of them prints a full Python traceback and exits 1 (verified for all three).

Fix applied: one `except FutureSchemaError` in `main()` renders the same one-line error / exit 2 for every present and future subcommand; test added in `tests/test_cli_contract.py`.

frontend/src/features/artifacts/ui.ts
```diff
@@ -296,14 +305,17 @@ export function renderViewer(): void {
...
 
+let viewerRequest = 0;
+
 function presentViewer(a: ArtifactRow): void {
+  viewerRequest += 1;
   viewerVersionState.value = a._exactVersion
     ? { status: "exact", artifact: a, versionId: String(a.version_id || "") }
     : { status: "latest", artifact: a, versionId: a.version_id || a.latest_version_id || null };
   dockArtifact.value = a;
   provMode.value = false;
```
Nobody-Zhang, line 322: **Conversation-strip tiles pin the head version, and the new isolation makes the pin real**

`tile.onclick = () => openViewer(a)` (L178): every strip row is a `_artifact_json` head whose `version_id` is the current latest (gateway stamps it on all rows), so `openViewer` always takes the `applyArtifactDeepLink` branch and opens an `_exactVersion` tab. On main that pin was soft — `syncArtifactVersion` still advanced every tab/dock row with the id. This PR deliberately skips pinned rows in `cache.ts`, `events.ts` and `load.ts`, so after the agent overwrites `results.csv` the strip-opened viewer keeps the old bytes with no re-render and no banner (only a ` · <vid>` suffix), and re-clicking the tile opens a second tab per overwrite. The Files grid already routes to `presentViewer`, and `openViewer`'s own doc comment says head rows must.

Fix applied: strip tiles call `presentViewer(a)`; test added in `ui.test.ts`.

frontend/src/islands/viewer.ts
```diff
@@ -162,16 +162,17 @@ function renderArtifactEditor(body: HTMLElement, a: ArtifactRow): void {
   const pop = el("div", "edit-ac hidden");
   body.appendChild(pop);
   bindEditorAutocomplete(ta, a);
-  artifactsFetch(`${API}/artifacts/${a.id}?_=${Date.now()}`)
-    .then((r) => r.text())
+  artifactsFetch(artUrl(a))
```
Nobody-Zhang: **Editor now reads the pinned version but saves the head — lost update**

The editor text is loaded via `artUrl(a)` (`/artifacts/versions/{vid}` for a pinned tab) while Save still POSTs `/artifacts/{id}/edit` with no base version, and the Edit button/menu item is offered on pinned tabs. Combined with the strip pin above this is routine: open `notes.md` pinned at v3 → agent writes v4 → Edit shows v3 → Save creates v5 from v3, silently superseding v4; `syncArtifactVersion` then skips the pinned row so the viewer still shows v3 after the “saved” toast.

Fix applied: Edit is hidden on `_exactVersion` tabs (menu item explains to open the latest version); the editor reads through the shared `fetchArtifactText` helper. Test added in `viewer.test.ts`.

frontend/src/features/send/send.ts
```diff
@@ -463,6 +466,10 @@ export async function send(text?: string | null, opts?: { execute?: boolean }):
       callLane("updateAnnotBadge");
     }
   } catch (e) {
+    // A transport failure is indeterminate; do not erase its optimistic bubble
+    // with a GET that may have raced the server's admission transaction.
+    if (e && Number.isInteger((e as { status?: number }).status) &&
+      (e as { status: number }).status >= 400) confirmHistorySubmission();
```
Nobody-Zhang: **A transport failure on `POST /message` never releases the history-submission token**

`beginHistorySubmission()` is confirmed only on a 2xx carrying `request_id` or when the caught error has `status >= 400`. A fetch `TypeError` (network drop, daemon restart, reset after headers) has no `status`, and `resetHistorySubmissions()` runs only when *switching* conversations. After one blip `historyUnconfirmed` stays 1 for the rest of the visit: every later `loadHistory` is `deferred` (“submission pending” banner whose Retry re-defers), `_replayGap` can never clear, `alignHistoryAfterTurn`'s follow-up is refused, and if a `processing` frame_update had set `running=true` and the terminal event is missed, the stopped `/status` is refused too → composer locked, Stop visible, until the user navigates away. Reproduced against the real modules.

Fix applied: a refusal releases immediately as before; an indeterminate transport failure releases after a bounded grace (`SUBMISSION_GRACE_MS`, 5 s) and then nudges `alignHistoryAfterTurn`. Test added in `first-send.test.ts`.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
-  callLane("closeAnnotPop");
-  callLane("updateAnnotBadge");
-  callLane("edacTeardown");
-  callLane("_molTeardown");
-  try {
-    const viewer = document.getElementById("dock-viewer");
-    if (viewer) viewer.innerHTML = "";
-  } catch {
-    /* no document */
+  if (switching) {
```
Nobody-Zhang, line 481: **Same-frame `resetHistory` reopen keeps notebook cells the server no longer lists**

`resetSessionScoped()` now runs only `if (switching)`, but `applySessionRevert`, `undoSessionRevert` and the `branch_projection_restored` / `branch_activation_state` resync reopen the *same* frame with `{ resetHistory: true }`. `loadExecutionLog` then does `mergeNotebookCells(serverCells, cells)` which seeds local keys first and never prunes, so cells removed by the revert stay rendered (and `kernels`, dock tabs, lineage, plan card from the previous branch persist). On main every open reset; only `activateSessionBranch` clears cells explicitly, which is why it alone still works.

Fix applied: the `resetHistory` branch calls `resetSessionScoped()` (transcript pieces it already handled are kept). Test added in `open.test.ts`.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
-  if (pid && pid !== project.value) {
-    project.value = pid;
-    _projArtFor.value = null;
+  if (deferred) {
+    result.messagesLoaded = false;
+    errors.messages = t(historyUnconfirmed.value > 0 ? "history.submissionPending" : "history.livePending");
+  }
+  if (!current(fid, gen)) return obsolete();
+  // Status failure never implies task completion or unlocks a running composer.
+  if (runState.status === "fulfilled" && unchanged() && (runState.value.running || !busy())) {
```
Nobody-Zhang, line 356: **Gap-replay / watchdog recovery leaves a phantom running cell and never reloads the execution log or artifacts**

`replay_begin` no longer clears `liveCells`/`_liveCell`, and `replay_end` now calls the GET-only `recoverConversation` instead of `openConversation`. When the cell's `notebook_cell_finished` and the terminal `frame_update` aged out of the replay buffer, the stopped read unlocks the composer and hides Cancel, but nothing retires the `live:true, status:'running'` cell (`liveCells.value = []` happens only in `startStream`/`turnDone`/switch) and neither `loadExecutionLog` nor `loadArtifacts` runs, so the finished cell's output and its artifacts don't appear until the next turn or navigation. The watchdog usually loses the race (`stale()` includes `!running`), so it does not rescue this.

Fix applied: on a stopped, unchanged read that still found live state, the commit finalizes live cells and triggers `loadExecutionLog` + `loadArtifacts`; the orphaned stream is finalized even when the messages GET failed. Tests added.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
+      });
+    });
+    if (!current(fid, gen)) return obsolete();
+    deferred = !unchanged() || busy();
+    if (!deferred) {
+      if (orphanedStream()) {
+        flushRender(streamAtRead as LiveStream, true);
+        liveStream.value = null;
+      }
+      if (host && stage) host.replaceChildren(stage);
```
Nobody-Zhang: **`host.replaceChildren(stage)` wipes the plan card and artifact tile strips on the GET-only paths**

`#messages` also hosts `.generated`/`.uploaded` tile strips (`renderConversationArtifacts`) and `#plan-card-live` (`renderPlanCard`, plus the legacy approval card). A full `openConversation` repaints them after the read; the three new GET-only paths (`turnDone → alignHistoryAfterTurn`, `replay_end → recoverConversation`, the Retry button) do not, and leave `planReady`/`planStatus` claiming a card exists so later `plan_progress` events update nothing visible. Trace: turn ends while a read is deferred → `turnDone` paints tiles + completed plan card → the microtask alignment replaces the children → both vanish until the next artifact event or navigation (tiles are order-dependent, the plan card deterministic).

Fix applied: only history items (`.msg`/`.step`), `#msgs-earlier`, the empty decoration and the retired stream wrap are replaced; other children are re-appended. Test added.

openai4s/server/execution_views.py
```diff
@@ -78,6 +79,16 @@ def execution_log(
             _with_dependency_defaults(cell)
             for cell in self._branch_cells(root_frame_id, branch_id)
         ]
+        output_bindings = output_artifact_bindings(
```
Nobody-Zhang, line 86: **Two versions of one filename from one cell make the Notebook show “Historical version could not be confirmed”**

`host.save_artifact` mints a version immediately (stamped with the live cell id, `host/data.py`), and the end-of-cell sweep mints another when the bytes changed since (same-cell reuse fires only on an equal checksum). `output_artifact_bindings` returns both rows; `notebookArtifactState` confirms only when exactly one `(artifact, version)` matches the filename, so the cell flips from confirmed (live) to unconfirmed (reload) with no download link and no inline ``. Trigger is mundane and prompted for: `df.to_csv('out.csv'); host.save_artifact('out.csv'); df2.to_csv('out.csv')` — ~10 bundled Skills use this loop.

Fix applied: the execution-log projection collapses to the newest version per `(cell, filename)` (`created_at`, then ordinal) while the sources export still lists every version; `artifact_lineage` also stops reading `version_meta` twice. Test added in `test_execution_view_service.py`.

openai4s/host/remote_science.py
```diff
@@ -326,13 +353,346 @@ def score_mutations(self, spec: dict) -> dict:
...
+
+def _csv_rows(text: str, field: str) -> tuple[list[str], list[dict[str, str]]]:
+    try:
+        rows = list(csv.reader(io.StringIO(text, newline=""), strict=True))
+    except csv.Error:
+        raise ValueError(f"malformed {field} CSV") from None
+    if len(rows) < 2 or not rows[0] or len(set(rows[0])) != len(rows[0]):
+        raise ValueError(f"empty or ambiguous {field} CSV")
+    headers = rows[0]
+    if any(len(row) != len(headers) for row in rows[1:]):
```
Nobody-Zhang, line 608: **A trailing blank line or a UTF-8 BOM makes a valid CSV “malformed”**

`csv.reader` yields `[]` for a trailing empty line, which `len(row) != len(headers)` reports as a malformed row; and `_decoded_text` decodes with `utf-8`, so a BOM lands in `rows[0][0]` (`'\ufeffmutation'`) and neither header set matches. Verified end-to-end: fold → `invalid_output: malformed pLDDT CSV row` / `missing or ambiguous pLDDT CSV columns`; score → the same for the mutation CSV. `scripts/fold_remote.sh` happens to be safe (`csv.writer`, CRLF), but the ESM wrapper has no in-tree reference and `print(df.to_csv())` is exactly what the provisioner specialist would write; on main the CSV bytes were accepted verbatim, so this is a tolerance regression with a misleading message.

Fix applied: trailing empty rows are dropped (a blank line *between* rows still fails), decoding uses `utf-8-sig`; tests added.

openai4s/host/remote_science.py
```diff
@@ -326,13 +353,346 @@ def score_mutations(self, spec: dict) -> dict:
...
+        if not isinstance(row, dict):
+            raise ValueError("top5 contains a non-object entry")
+        mutation = row.get("mutation")
+        if not isinstance(mutation, str) or mutation not in scores or mutation in seen:
+            raise ValueError("top5 mutation conflicts with validated CSV")
+        seen.add(mutation)
+        score_keys = [key for key in ("score", "esm_score") if key in row]
+        if len(score_keys) != 1:
+            raise ValueError("top5 has missing or ambiguous score fields")
+        if _finite(row[score_keys[0]], "top5 score") != scores[mutation]:
```
Nobody-Zhang: **`top5` scores are compared with exact float equality across two independent serializations**

`_finite(row[score]) != scores[mutation]` demands bit-equality between the summary JSON and the CSV. The natural wrapper pairing already fails: a float32 ESM score printed by pandas is `-1.2345678`, `json.dumps(float(t))` of the same tensor is `-1.2345677614212036`; a 4-dp summary differs in the 5th digit. Both were rejected as `top5 score conflicts with validated CSV`.

Fix applied: `math.isclose(rel_tol=1e-6, abs_tol=1e-4)` — scores are log-likelihood ratios of order 0.1–10, so this cannot hide a substantively different top5 while it stops honest rounding from being reported as fabrication. Tests added (also: `_finite` now requires ASCII decimals for text fields — `float()` accepted `1_0.5` and Unicode digits that no PDB reader parses the same way; a NaN inside the confidence JSON is now reported as non-finite rather than “invalid JSON”; every empty `positions` spelling shares one message).

frontend/src/features/execution/provenance.ts
```diff
@@ -152,24 +165,25 @@ export function renderProvenanceInto(v: HTMLElement, a: unknown): void {
...
   try {
     env = (snaps[key] ||
       (snaps[key] = await (a && a.id
-        ? api(`/artifacts/${a.id}/environment`)
+        ? api(artifactMetadataUrl(target, "environment"))
         : api("/kernel/environment")))) as EnvSnapshot;
     _envSnapById.value = snaps;
   } catch (e) {
-    if (provMode.value && provSub.value === "environment") {
+    if (provMode.value && provSub.value === "environment" && artifactMetadataCacheKey(asArtifact(dockArtifact.value)) === key) {
```
Nobody-Zhang, line 180: **Provenance → Environment now shows a *load failure* for every upload and legacy artifact**

`artifactMetadataTarget` promotes every docked row that has a head version to an exact-version request, and the gateway (deliberately, tested) answers `404 environment_snapshot_unavailable` when no snapshot was recorded — which is every user upload (`commit_artifact_upload` hard-codes `env_snapshot_id=None`), every pre-snapshot capture, every `host.save_artifact` provisional row and every capture whose freeze threw. The catch here renders that as `Failed to load environment: … [req-id]`, while the designed copy for this case (`prov.env.liveFallback` / `envSnapshotHonesty`) becomes unreachable. On main the same click showed the live fallback.

Fix applied (frontend only, the backend posture is kept): `environment_snapshot_unavailable` renders a dedicated “no environment was recorded for this version” state. Test added in `lineage.test.ts`.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
+  if (page.status === "rejected") errors.messages = readError(page.reason);
+  if (steps.status === "rejected") errors.steps = readError(steps.reason);
+  if (runState.status === "rejected") errors.runState = readError(runState.reason);
+  const cached = historyContent.value?.fid === fid ? historyContent.value : null;
+  // Without identities ordinary WS text cannot be joined to a REST tail.
+  // Keep the confirmed DOM and the live stream until a stopped, unchanged GET.
+  const stopped = runState.status === "fulfilled" && runState.value.running === false;
+  const orphanedStream = (): boolean => !!streamAtRead && liveStream.value === streamAtRead && stopped;
+  const busy = (): boolean => historyUnconfirmed.value > 0 || (!!liveStream.value && !orphanedStream());
+  let deferred = !unchanged() || busy() || (!!wasDeferred && !stopped);
```
Nobody-Zhang: **A `/status` failure keeps a previously deferred read deferred (discarding the good page), and a deferred first read is labelled “History could not be loaded.”**

`let deferred = … || (!!wasDeferred && !stopped)`: `stopped` is false when `/status` merely *failed*, so with nothing live the successful page is discarded and `errors.messages` becomes “Live messages are preserved; history will align after the turn stops” while no turn runs (bounded to while `/status` fails — one successful Retry heals it; Bugbot's report above). Separately, a deferred read with no cached content maps to `status: "error"` (L344), so opening a session whose turn is streaming shows “History could not be loaded.” + Retry although nothing failed.

Fix applied: when the run state could not be read, the client's own `running` decides whether to keep waiting; a deferred read reports `partial`; and the orphaned stream is finalized even when the messages GET failed (the composer unlocked while the old bubble kept its cursor). Tests added.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
+    }
   }
-  const found = (sessions.value as Array<{ id?: string; project_id?: string }>).find(
-    (x) => x && x.id === fid,
-  );
+  const complete = result.messagesLoaded && result.stepsLoaded && result.runStateLoaded;
+  historyLoad.value = { ...result, fid, generation: gen,
+    status: complete ? "loaded" : result.messagesLoaded || cached ? "partial" : "error",
+    errors, deferred };
+  if (complete && _replayGap.value === fid) _replayGap.value = null;
```
Nobody-Zhang, line 385: **An idle session can stall on “partly restored” after any seq-stamped event, with the replay gap kept**

`unchanged()` compares `_seqSeen[fid]`, and `connect.ts` advances it for *every* event carrying `seq` — the hub stamps every frame-scoped publish (`kernel_status`, `execution_queue`, post-turn `artifact_created` from background jobs). A reconnect on an idle session → `replay_end` → `recoverConversation`; a kernel idle-timeout event during the ~0.5 s read makes it `deferred`, the banner promises alignment “after the turn stops”, but the only re-alignment hook is `turnDone`, which never fires without a turn; `_replayGap` stays set until the user clicks Retry.

Fix applied: when a read ends deferred while nothing is live (`!liveStream && !running && historyUnconfirmed === 0`), `recoverConversation` re-reads itself a bounded number of times (2); the banner and Retry remain for sessions that keep moving. Test added.

frontend/src/features/csv/csv.ts
```diff
@@ -112,7 +112,10 @@ export function parseTable(
         const obj = j as Record<string, unknown> | null;
         j = (obj && (obj.rows || obj.data || obj.candidates || obj.items)) || [];
       }
-      if (Array.isArray(j) && j.length && typeof j[0] === "object") {
+      if (
+        Array.isArray(j) && j.length &&
```
Nobody-Zhang, line 123: **A homogeneous JSON array-of-arrays no longer renders as a table (regression vs main)**

main accepted `typeof j[0] === 'object'` and `sheetShape` iterated `Object.keys(row)` on arrays, so `[[1,2],[3,4]]` (`json.dump(df.values.tolist(), f)`) rendered as a two-column sheet with headers `0`,`1`. The new `every(!Array.isArray(row))` rule (and `sheetShape`'s matching skip) sends every array-of-arrays `.json` to the raw-source view; the new tests only cover a *mixed* array containing `[]`.

Fix applied: a homogeneous array of arrays is mapped to positional-column objects before the object check (a mix of shapes is still raw text); test added in `csv.test.ts`. Related: a `.json`-named file whose bytes are binary now gets the download tile instead of 300 k of replacement characters plus an Expand button (`renderers.ts`), using control-character density only — `looksBinary`'s base64-run heuristic is what long JSON strings trip.

Nobody-Zhang:
## Review (xhigh, recall mode) — fixes pushed as a5764cd7

15 inline findings above, all verified against the tree (most reproduced empirically) and all fixed in a5764cd7 with a regression test each. The three that matter most:

1. **`preflight_schema` opens read-only before the rw connection, so a hot rollback journal (daemon killed mid-commit) makes every `Store()` open raise `attempt to write a readonly database` and nothing ever replays it.** Now defers to the formal connection, which still refuses a future schema before any DDL.
2. **Strip tiles pinned the head-at-click and the new cache isolation froze them; the editor read pinned bytes but saved the head.** Strip → latest tab; Edit hidden on pinned tabs.
3. **A transport failure on `POST /message` left `historyUnconfirmed` > 0 for the whole visit** (every read deferred, composer could lock). Bounded 5 s grace, then realign.

Also fixed on the branch beyond the 15: `?version=` honoured on the id bytes route; ASCII-only decimals in `_finite`; non-finite confidence named as such; one message for every empty `positions`; no duplicate `version_meta` read; `artUrlBust`/inline fetches replaced by shared helpers; CLAUDE.md now says the browser gates need the science extra; contiguous-only reuse of cached earlier pages; bounded idle re-reads; array-of-arrays JSON tables; binary `.json` → download tile.

**Left as follow-ups (product calls or larger than this PR):** `delegate(wait=True)` multi-child wait has no cancel probe (pre-existing); a server-side base version for `/artifacts/{id}/edit`; carrying `output_artifacts` on `notebook_cell_finished` to drop the per-cell `loadExecutionLog` refetch and the client `figures` heuristic; the N+1 bindings query; a server seq snapshot for the three-read history join; the diagnostics `SETTINGS` map living client-side and `/onboarding/complete` missing from its invalidation allowlist; the remote stderr tail dropped from fold/score errors; `mean_score` overwritten to None; and (pre-existing on main) a session package exported with a delegated child's artifact cannot be re-imported.

Verified locally at a5764cd7: full offline suite, vitest 670/670, pre-commit + mypy, harness pr tier, directory READMEs, secret scan, webui contract, response contract 212/212, frozen shapes (1165, no breaking change), and `browser_smoke.mjs` against a token-gated daemon. Merge still needs a code-owner approval.

frontend/src/features/execution/provenance.ts
```diff
@@ -152,24 +165,25 @@ export function renderProvenanceInto(v: HTMLElement, a: unknown): void {
...
       body.innerHTML = "";
       body.appendChild(el("div", "dock-empty", t("prov.env.loadFailed", apiErrorText(e))));
     }
     return;
   }
   const docked = asArtifact(dockArtifact.value);
-  if (!provMode.value || provSub.value !== "environment" || (a && artifactCacheKey(docked) !== key))
+  if (!provMode.value || provSub.value !== "environment" || (a && artifactMetadataCacheKey(docked) !== key))
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in a5764cd7: the `a &&` guard in `renderProvEnvironment` is gone (`a` is always an object there), so the docked-key comparison is unconditional.

frontend/src/features/messages/open.ts
```diff
@@ -179,188 +185,293 @@ function resetSessionScoped(): void {
...
-  const previousFid = currentId.value;
-  if (previousFid && previousFid !== fid) unsub(previousFid);
-  resetNotebookCellCaches(previousFid, fid);
-  if (pid && pid !== project.value) {
-    project.value = pid;
-    _projArtFor.value = null;
+  if (deferred) {
+    result.messagesLoaded = false;
+    errors.messages = t(historyUnconfirmed.value > 0 ? "history.submissionPending" : "history.livePending");
+  }
```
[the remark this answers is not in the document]
Nobody-Zhang: Addressed in a5764cd7 (a different shape from the Bugbot autofix, which was not pushed): when the run-state GET fails, `wasDeferred` no longer keeps the read deferred unless this client still believes a turn is running (`runUnknown`), so a quiet retry commits the successful page and reports `errors.runState`; an orphaned stream is finalized on a stopped read even when the messages GET failed; and a read left deferred on an idle session re-reads itself a bounded number of times instead of waiting for a turn end that never comes. Tests: “a run-state failure with nothing live commits a previously deferred page”, “finalizes an orphaned stream after a stopped status even when messages fail”, “…re-reads itself, boundedly” in open.test.ts.

PR #160, opened 2026-09-15
fix: bound model calls and preserve workbench state

Nobody-Zhang:
## Summary

This integrates six reviewed fixes from `next` so model failures have bounded retries and resource use, artifact edits cannot silently overwrite a newer version, and stale workbench reads cannot replace current navigation or provenance state.

- **Model calls (P0-01/P0-02):** share a three-send budget, cancellation and total deadline across retries and narrowly defined stream compatibility fallback. Bound JSON/SSE input and the text queue; preserve real usage evidence, unknown usage and exactly-once accounting for drained old calls.
- **Artifact editing (P0-03):** add optional `expected_version_id` with lock-protected conflict detection. The new UI always supplies it, protects loading and immutable baselines, retains bounded drafts, and reconciles lost save responses through reads without automatically resending writes.
- **Files and navigation (P1-01/P1-03):** use one filtered result for cards, counts and pagination; separate navigation identity from request generations so stale success, failure, cleanup and creation responses cannot take over the active session.
- **Provenance and export (P1-02):** distinguish loading, failure and confirmed absence; validate response structure, retain exact version/owner identity, and require successful evidence reads before downloading.
- **P2 preparation only:** document snapshot retention/restoration and inbound connection budgets with future acceptance matrices. These documents enable no snapshot retention, migration changes, inbound deadlines or connection quotas.

The integration contains 12 commits. Database structure, version and dependency locks are unchanged. Workbench source and rebuilt `dist` are included together; compatibility facades were changed surgically.

[Sequential implementation and acceptance ledger](https://github.com/PKU-YuanGroup/OpenAI4S/blob/192f81e150512787bc88f272e4a1cac2d93113e5/docs/next-improvements-20260914.md)

## Area

- [x] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Test / harness
- [x] Documentation
- [x] Security-related change

## Verification

[Completed CI for exact head `192f81e150512787bc88f272e4a1cac2d93113e5`](https://github.com/PKU-YuanGroup/OpenAI4S/actions/runs/34889356928): **25 applicable jobs passed; 4 not-applicable jobs skipped**. PR-triggered checks are separate and must pass before merge.

- Full local offline suite: **8958 passed / 26 skipped** on the unchanged production/test content. Final-head CI passed Python 3.10, 3.12, 3.13 and 3.14.
- All-file pre-commit and strict mypy passed on a clean candidate. Additional independent review reran 175 frontend and 90 artifact backend tests successfully.
- Chromium, Firefox and WebKit each passed 800 frontend tests, source/dist parity and 14 browser scenarios. Focused kernel, agent, gateway and shared HTTP regressions passed.
- Response capture: **1167 shapes, 212/212 routes, no breaking drift**. Route contracts and all 38 harness scenarios passed.
- Docker runtime and the separate Linux persistent-interrupt and full filesystem/egress sandbox gates passed.
- Final wheel and sdist both passed artifact verification and were actually installed with `--no-deps` into separate fresh Python 3.12 environments outside the checkout: 13 modules / 604 Skills, only OpenAI4S installed.
- Bilingual directory coverage, source/package credential scans and Skills installer/package checks passed.

Commands exercised across the staged validation and final-head CI include:

```text
uv run mypy
npm test --prefix frontend
node tests/browser_matrix.mjs --browser=chromium
node tests/browser_matrix.mjs --browser=webkit
uv build --no-sources
```

Explicit, separate Ark validation covered text, streaming tools, cancellation, next-call recovery and artifact generation. The final run made eight requests: seven returned measured usage totaling 46,412 tokens; one cancelled unmetered browser request remains unknown. The metered-drain test accounted the old result exactly once. Three browsers then edited, filtered and exported the real synthetic artifact; six saved exports matched the same edited version.

Not executed: the proposed P2 snapshot/slow-connection matrices, because this PR delivers their design only. Ark live checks do not replace the Anthropic, Responses and Gemini protocol fixtures. The browser receipt proves stream arrival before Stop, not independently captured pre-click assistant paint. Artifact recovery completed after correcting an intermediate completion-bullet validation error; the ledger preserves that evidence.

## Risk and Reviewer Focus

Review `openai4s/llm/transport.py`, `client.py`, `usage.py`, `http_deadline.py` and `agent/runtime.py` for retry/deadline ownership, stream termination, cancellation and unknown usage. Unresolved DNS retains its legacy call slot; stdlib cannot forcibly interrupt DNS.

Review `server/artifacts.py`, the artifact edit route and `frontend/src/features/artifacts/editor.ts` for lock-scoped version comparison, compatibility when the expected version is omitted, draft bounds and lost-response recovery. Review session navigation and provenance/export guards for A→B→A races and fixed ownership/version identity. Deploy or roll back frontend source/dist with its backend counterpart.

## Checklist

- [x] Full integration diff independently reviewed; no remaining reproducible blockers.
- [ ] Small, single-purpose PR — this is the requested six-fix integration, organized and validated sequentially.
- [x] No unrelated formatting, wholesale facade rewrites, or vendored/fixture reformatting.
- [x] Workbench changes are in `frontend/`; rebuilt `dist` is committed; legacy UI receives no new features.
- [x] Relevant offline, contract, browser and runtime checks passed; no tests deleted without replacements.
- [x] New live validation is explicitly `live_llm`; default tests and CI need no API credentials.
- [x] Core remains stdlib-only; no new dependencies or unguarded optional science imports.
- [x] No real credentials or private data in the submitted diff or PR body; security-sensitive paths have synthetic regression coverage.
- [x] Docs and bilingual inventories match behavior and retain the evidence limits above.
- [x] Existing unrelated uncommitted document edits are excluded from this PR and the release archives.

---

> [!NOTE]
> **High Risk**
> Changes span LLM transport deadlines, usage/accounting, and conditional artifact writes plus navigation races—security- and data-integrity-sensitive paths with broad gateway and UI surface area despite heavy test coverage.
> 
> **Overview**
> This integration lands six reviewed `next` improvements: tighter model-call behavior, safer artifact editing, and workbench state that survives races and lost responses.
> 
> **Model calls** add `OPENAI4S_LLM_TOTAL_TIMEOUT` (default 600s) as one absolute deadline per logical call across retries, streaming, backpressure, and cancelled-call usage drain. Streaming fallback to a blocking POST is narrowed to explicit structured `stream` refusals; shared three-send / one-compatibility-POST budgets and input size limits are documented in `configuration.md`.
> 
> **Artifact editing** accepts optional `expected_version_id` on edit routes (409 on head mismatch; response includes `unchanged`). The new frontend editor checksum-verifies a fixed baseline, sends the expected version on save, keeps up to 10 / 8 MiB in-page drafts with a Files **Retained drafts** panel, and reconciles unknown or conflict outcomes via reads only—no automatic resend. Captured schemas add PUT/PATCH edit success shapes.
> 
> **Files & navigation** unify cards, counts, pagination, and filters under session/`_openGen` ownership so stale loads and WS updates cannot paint the wrong session; refresh keeps requested page capacity and defers errors/load-more to the current listing.
> 
> **Provenance/export** (per PR scope) tighten validation and export guards so metadata and downloads stay pinned to verified versions and owners.
> 
> **Docs-only P2 prep** adds bilingual pre-upgrade snapshot and inbound-connection design ledgers plus the T0–T9 acceptance record; no snapshot retention, migration, or gateway quotas are enabled. `plan-crosswalk.json` evidence digests refresh for two closed rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 192f81e150512787bc88f272e4a1cac2d93113e5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openai4s/llm/capabilities.py
```diff
@@ -1136,7 +1136,60 @@ def normalize_usage(
...
+                value = value.get(part, missing)
+            if value is not missing:
+                if type(value) is int and value >= 0:
+                    measured[key] = value
+                else:
+                    invalid.add(key)
+                break
+    if mapping.input_excludes_cache and "input_tokens" in measured:
+        if invalid & {"cache_read", "cache_write"}:
+            measured.pop("input_tokens")
```
Nobody-Zhang, line 1185: **An explicit `null` cache counter voids the measured `input_tokens`.**

Anthropic's usage schema declares `cache_creation_input_tokens` / `cache_read_input_tokens` as `integer | null`. A `null` reaches this strict loop as present-but-not-an-`int`, so the key is marked invalid and `input_excludes_cache` then pops `input_tokens` out of the evidence entirely. Verified against the real code:

```
normalize_usage({"input_tokens":100,"output_tokens":20,
                 "cache_creation_input_tokens":None,
                 "cache_read_input_tokens":None}, "claude")
  public  : input_tokens=100 ... total_tokens=120
  measured: {'output_tokens': 20, 'completion_tokens': 20}
  verifiable_token_usage -> None
```

The same payload with those keys simply *absent* measures 120 correctly. Downstream: `add_usage` writes `input_tokens=0` to the frame while the UI shows 100; `record_session_llm_usage` writes an `llm_input_tokens_unknown` row that the new `check_quota` gate turns into a permanent `QuotaExceeded` for the whole window (nothing clears those rows); and in an Auto Mode run `gateway.py:6272` raises `AutoBudgetDenied` + `fail_measurement`.

Not changed here, because `test_invalid_anthropic_cache_cannot_certify_zero_input` parametrises `cache=None` and pins exactly this treatment — and the fail-closed reading is defensible (if cache usage is unknown and the wire excludes it, the true input is unknown). Flagging so the trade-off is a decision rather than a discovery: as written, Auto Mode and team quotas are unusable against any Claude endpoint that emits the nullable form. `_token_value`, the public path, treats `null` as absent, so the two halves of this function disagree about the same payload.

openai4s/llm/capabilities.py
```diff
@@ -1136,7 +1136,60 @@ def normalize_usage(
         "completion_tokens": output_tokens,
         "total_tokens": total,
     }
-    return result
+    from .usage import UsageMetrics, usage_evidence
+
+    if isinstance(raw, UsageMetrics):
```
Nobody-Zhang: **Re-normalising an already-normalised usage folds the cache counters twice.**

The `isinstance(raw, UsageMetrics)` guard sits *after* `result` has been rebuilt, so the `input_excludes_cache` fold at 1124-1125 has already run a second time by the time the guard returns:

```
first : input_tokens=607  total_tokens=617   evidence: 607
second: input_tokens=1114 total_tokens=617   evidence: 607     # 1114 + 10 != 617
```

`tests/test_llm_usage_evidence.py:314` makes re-normalisation a supported contract, but only on the `chatgpt` mapping where `input_excludes_cache` is False, so the Anthropic case is untested.

**Fixed** by moving the guard above every derivation, and covered by `test_renormalizing_a_cache_folding_wire_does_not_fold_twice` (verified red without the fix).

The alias walk one block down had a related divergence: it `break`s on the first path that resolves to an unusable value, so `{"cache_read_input_tokens": null, "cache_read_tokens": 42}` was recorded invalid while `_token_value` used 42. It now falls through to the next alias the way `_token_value` does, and only marks the key invalid when no alias yields a usable value — a malformed *intermediate* (`prompt_tokens_details: "junk"`) still invalidates, which `test_ledger_cost_keeps_invalid_cache_evidence` pins.

openai4s/llm/usage.py
```diff
@@ -0,0 +1,95 @@
...
+
+class RawUsage(dict):
+    """Provider counters, with whether they describe the finished generation."""
+
+    def __init__(self, value: Any = None, *, final: bool = True) -> None:
+        super().__init__(value if isinstance(value, Mapping) else {})
+        self.final = final
+
+
+class UsageMetrics(dict):
```
Nobody-Zhang: **The metering evidence is an attribute on a `dict` subclass, so any plain copy fails *open*.**

```
verifiable_token_usage(n)        -> None    # evidence says final=False
verifiable_token_usage(dict(n))  -> 1540    # evidence gone, key-scan fallback
```

`dict(u)`, `{**u}`, a `json.dumps`/`loads` round-trip and every SQLite hop drop `.evidence` and `.final` silently, and the fallback then certifies counters the evidence had refused — an unmeasured reply becomes a measured one and enters the Auto Mode commit and the governance ledger.

The PR is already paying for this a boundary at a time (`agent/models.py:92`, `review.py:377`, `scientific_reviewer.py:394` each had to switch to `copy_usage`), while `openai4s/server/session_package.py:3200` still does `dict(item.get("usage") or {})`. Any boundary added later that forgets is a silent fail-open with no type error to catch it.

Not changed here — the fix is structural (carry the evidence as an explicit sibling field, e.g. `reply["usage_evidence"]`, so a dropped copy is visible to every serializer) and well outside this diff. Worth a follow-up, plus an audit of the remaining plain-copy sites.

openai4s/llm/providers/openai.py
```diff
@@ -210,12 +180,14 @@ def _on_event(evt: dict) -> None:
                 detail = error or evt.get("message") or str(evt)
             raise LLMError(f"OpenAI stream error: {detail}")
         if evt.get("usage"):
-            state["started"] = True
             state["usage"] = evt["usage"]
+            state["output_committed"] = True
         choices = evt.get("choices") or []
         if not choices:
+            if state["terminal"] and evt.get("usage"):
```
Nobody-Zhang, line 187: **Usage that does not arrive on or after the terminal chunk is discarded entirely.**

`usage_final` is set only in these two shapes, so a stream that delivers exact counters anywhere else returns `RawUsage(..., final=False)` and `measured_usage()` returns `{}` — even though `state["usage"]` holds the real numbers and the public dict displays them. Verified:

```
usage-before-finish: raw {'prompt_tokens': 11, 'completion_tokens': 3} final=False
                     measured {}   verifiable_token_usage None
```

On its own that was survivable. Combined with `gateway.py:6272`, which widened its gate from `extra and admission and token_admission` to `admission is not None`, it is not: the first Auto Mode turn against a proxy that ignores `stream_options` (the case the comment at line 132 calls *"harmless when unsupported"*) now raises `AutoBudgetDenied("adapter token usage is not verifiable")` and calls `fail_measurement(run_id)` — killing a run whose output and token counts were correct.

Not changed here: `test_initial_unmeasured_facade_reply_cannot_freeze_a_zero_budget` pins the widened gate with `extra=False`, so this is intended. But the operator-facing message does not name the real cause, and there is no fallback to a non-streaming request to obtain usage. A declared per-endpoint capability ("does this wire report terminal streamed usage?") would let absent evidence settle as `unknown` rather than refuse the run.

openai4s/llm/providers/anthropic.py
```diff
@@ -182,25 +208,38 @@ def _on_event(evt: dict) -> None:
                 block["signature"] = delta["signature"]
             return
         if event_type == "message_delta":
+            state["output_committed"] = True
             delta = evt.get("delta") or {}
             if delta.get("stop_reason") is not None:
                 state["finish"] = delta["stop_reason"]
             state["usage"].update(evt.get("usage") or {})
+            state["usage_final"] = (
```
Nobody-Zhang: **`usage_final` is assigned, not latched, inside a handler for a repeatable event.**

`message_delta` repeats. A second one carrying only `{"delta": {"stop_sequence": null}}` resets `usage_final` to `False` after an earlier delta legitimately set it — line 215 accumulates `state["usage"]` monotonically, so the counters survive while the bit that says they are trustworthy does not. The reply then returns `RawUsage(final=False)`: `add_frame_tokens` records 0/0, `verifiable_token_usage` is `None`, and `record_session_llm_usage` writes `llm_*_unknown` rows that exhaust the quota window.

Every sibling flag in this function is a latch (`|=` at 154, `= True` at 159/192/211/221); this one was the exception.

**Fixed** to `|=`, covered by `test_anthropic_usage_survives_a_later_message_delta` (verified red without the fix).

openai4s/llm/transport.py
```diff
@@ -80,6 +92,102 @@ def bind_call_context(fn, **context):
...
+            output_committed=err.output_committed if err is not None else False,
+            body=err.body if err is not None else None,
+            retryable=False,
+        )
+
+        failure.llm_not_started = not self.sent
+        return failure
+
+
+def streaming_refused(error: TransportError) -> bool:
```
Nobody-Zhang, line 176: **The gate misses the canonical structured refusal, so the compatibility path is unreachable for both real wire shapes.**

Requiring `error_code` to be one of three OpenAI-specific strings excludes:

| refusal body (HTTP 400) | `_error_code` | old result |
|---|---|---|
| `{"error":{"param":"stream","type":"invalid_request_error","code":null}}` — OpenAI's canonical unsupported-parameter body | `invalid_request_error` | **False** |
| `{"error":{"type":"invalid_request_error", ...}}` — every Anthropic body | `invalid_request_error` | **False** |

Anthropic never emits `code`, so `anthropic.py`'s whole `_StreamStartError` branch was dead — and `test_anthropic_explicit_stream_refusal_falls_back_to_blocking` only passes because it fabricates `error_code="streaming_not_supported"`, a code the Messages API does not produce. The README sentence added in this PR describes a branch that cannot fire.

**Fixed**: a body that names `param: "stream"` is itself the explicit structured refusal, whatever it calls the code. The narrowing is intact — still 400/422 only, still non-retryable, still no committed output, and a `param` naming anything else still refuses. Every row of `test_only_structured_stream_refusal_allows_one_json_attempt` still passes, plus two new tests (one positive, one negative), both verified red without the fix.

openai4s/config.py
Nobody-Zhang, line 1: **`timeout_s` is unvalidated, and this PR gave it a constructor that raises.**

`cfg.timeout_s` now becomes `HTTPExchangeDeadline(idle_timeout=...)` (`transport.py:297`), which rejects non-positive and non-finite values with a bare `ValueError` — while `LLMConfig` validates only the sibling `total_timeout_s` two lines below.

`OPENAI4S_LLM_TIMEOUT=0` (`.env.example:63` ships this as an editable knob; `0` is the natural "no idle timeout" idiom) therefore makes *every* LLM call die inside `_exchange()`. `post_json`/`post_sse` catch only `HTTPExchangeTimeout`, `URLError`, `OSError` and `HTTPException`, and `_retry_loop` catches only `TransportError`, so it escapes the whole typed-transport contract: `llm_failure_code` cannot classify it and `_friendly_error` has no copy for it. Before this PR the value was only handed to `urlopen`.

**Fixed**: validated in `__post_init__` with the same shape as its sibling, so a bad value is one clear config error at construction instead of a traceback per call. Covered by `test_request_timeout_is_refused_before_it_reaches_the_exchange`.

Related and *not* changed: the `max(cfg.timeout_s, 60.0)` idle floor was removed from all three stream adapters. The deleted comment said it existed "so a heavy-reasoning model that pauses between events isn't cut off". The default (120s) is unaffected, but any operator who tuned this knob downward now gets streams cut mid-thinking.

openai4s/config.py
```diff
@@ -133,13 +133,33 @@ class LLMConfig:
     model: str = ""
     # Secret: sourced from the environment (or the git-ignored .env). Empty
     # when unset; llm.chat then raises a clear error.
-    api_key: str = ""
+    api_key: str = field(default="", repr=False)
     # Deep-thinking models: keep a conservative default output cap.
     max_tokens: int = int(os.environ.get("OPENAI4S_LLM_MAX_TOKENS", "4096"))
     temperature: float = float(os.environ.get("OPENAI4S_LLM_TEMPERATURE", "0.7"))
     timeout_s: float = float(os.environ.get("OPENAI4S_LLM_TIMEOUT", "120"))
+    total_timeout_s: float = field(
```
Nobody-Zhang, line 165: **The range validation written for this field is unreachable from the environment variable.**

`default_factory` calls `float()`, which runs inside `__init__` — before `__post_init__` can turn a bad value into the careful message below it. So `OPENAI4S_LLM_TOTAL_TIMEOUT=` (blank; one keystroke from the `.env.example:66` line) or `=abc` aborts `Config()` — the daemon and CLI boot path — with:

```
ValueError: could not convert string to float: ''
```

instead of *"LLM total timeout must be a finite number from 1 to 3600 seconds"*. The good message is reachable only when a caller passes the value programmatically, which is not how this knob is normally set.

**Fixed**: the factory returns the raw environment value and `__post_init__` owns the parse (it already accepts a `str` via `float(self.total_timeout_s)`). Covered by `test_total_timeout_environment_value_reports_its_own_range`, verified red without the fix.

openai4s/agent/runtime.py
```diff
@@ -412,31 +423,54 @@ def report_abandoned_reply() -> None:
...
-            _LATE_ACCOUNTING.submit(sink, reply)
+            detached_call.detach()
+
+            def account_and_release(value):
+                try:
+                    sink(value)
+                finally:
+                    detached_call.settle()
+
+            _LATE_ACCOUNTING.submit(account_and_release, reply)
```
Nobody-Zhang: **A failed submit leaks a detached-call budget slot for the process lifetime.**

`report_abandoned_reply` latches `reported = True` and calls `detached_call.detach()` (line 447) *before* handing the settle to the drain. If `submit` raises — `threading.Thread(...).start()` → `RuntimeError: can't start new thread`, under exactly the process pressure this budget exists for — then both `settle_unreported()` call sites (570 and 528) short-circuit on `reported` and nothing ever returns the slot. Four of these in one session and `admit()` permanently refuses every later turn with *"this session already has cancelled model requests still billing"*, for calls nobody cancelled.

**Fixed**: settle before giving up, and log rather than re-raise — this also runs on the owning thread from `abandon()`, where raising would fail the user's turn for a metering failure, and `_LateAccounting._drain` already treats accounting as fail-soft. Covered by `test_a_late_accounting_submit_failure_still_releases_its_budget_slot`, verified red without the fix.

Related, not changed: the guard at line 426 now also fires for an ordinary non-cancelled provider error, so a plain 500 takes the *abandoned* path and charges the detached-call budget for a call that was never detached. It is released again by `account_and_release`, so it is transient rather than a leak — but the name and the refusal message both say "cancelled".

openai4s/server/scientific_review.py
```diff
@@ -840,12 +840,23 @@ def evaluate(
...
+                        if getattr(exc, "llm_not_started", False):
+                            budget.release(admission_id, started=False)
+                        else:
+                            budget.mark_unknown(admission_id)
                     except Exception:  # noqa: BLE001 - unknown is fail-closed
                         pass
                     if token_admission_id:
                         try:
-                            budget.mark_unknown(token_admission_id)
+                            known = verifiable_token_usage(getattr(exc, "usage", None))
```
Nobody-Zhang, line 907: **Scientific review got exact usage-based budget commits but still spends tokens outside the quota ledger.**

```
grep -rn "enforce_llm_quota|record_session_llm_usage|add_frame_tokens" openai4s/server/scientific_review.py
(no matches)
```

The sibling `ReviewService` gained all three in this very PR (`openai4s/server/reviews.py:444-459`), and the classic reviewer's pre-call gate is wired at `gateway.py:2751`. So a team member whose `llm_output_tokens` quota is exhausted is refused on ordinary turns *and* on classic review, but can still run unbounded scientific review rounds — and none of those tokens ever reach `usage_ledger`, so the quota the check reads can never fill from this path.

Not changed here: wiring a quota gate into this service is a product change rather than a review fix, and it needs a decision about what a mid-round denial does to an in-flight review. Filing it because this PR closed exactly this hole one file over, which makes the omission look deliberate when it reads like an oversight.

openai4s/storage/governance.py
```diff
@@ -552,8 +566,9 @@ def record_session_llm_usage(store: Any, root_frame_id: str, usage: Any) -> None
     With no ownership row it reads and never writes (INV-1). Metering must
     never break the call it meters, so every failure here is swallowed.
     """
-    if not usage:
-        return
+    from openai4s.llm.usage import measured_usage
+
+    counters = measured_usage(usage)
```
Nobody-Zhang: **Metering work moved above the `try` whose whole contract is that it swallows.**

The docstring two lines up states *"Metering must never break the call it meters, so every failure here is swallowed"* — and the `try:` starts on the next line, with both the lazy import and this call outside it. `measured_usage` does `dict(evidence.counters)` and `value.items()` on a caller-supplied object, so a usage whose `evidence.counters` is not a mapping (or a `Mapping` whose `items()` raises) propagates a `TypeError` out of `record_session_llm_usage` into `ReviewService`'s `finally` and the frame-accounting path, failing a turn that had already succeeded.

**Fixed**: both lines moved inside the `try`. Covered by `test_metering_a_hostile_usage_object_cannot_fail_the_call_it_meters`, verified red without the fix.

frontend/src/features/sessions/projects.ts
```diff
@@ -291,49 +291,42 @@ export function renderProjMenu(): void {
 }
 
 export function selectProject(id: string): void {
-  project.value = id;
+  beginProjectNavigation(id);
```
Nobody-Zhang: **A sidebar-only project switch strands the "Load earlier" latch.**

`selectProject` is the one caller that bumps the navigation generation while deliberately keeping the conversation open, and `beginNavigation()` resets `sessionsLoading` / `foldersLoading` / `_sessionsLoadingMore` but not `_msgEarlierLoading`.

Repro: open a long conversation → click "Load earlier" → switch the sidebar project while the request is in flight. `loadEarlierMessages`' `finally` (`messages.ts:94`) sees `gen !== _openGen.value` and skips `_msgEarlierLoading.value = false`; the button stays `disabled` showing "Loading…" (`messages.ts:56-57`) and every later click early-returns at `messages.ts:63` — permanently, until the user opens a different conversation. Every other `beginNavigation()` caller either follows with `openConversation` (which resets it at `open.ts:470`) or resets it inline; pre-PR `selectProject` did not touch `_openGen` at all.

**Fixed** in `beginNavigation()`, beside the three flags it already clears — the superseded request can no longer clear it itself, so its owner should. Covered by a new case in `load.navigation.test.ts`, verified red without the fix.

frontend/src/features/artifacts/editor.ts
```diff
@@ -0,0 +1,193 @@
...
+      if (this.store.drafts.get(this.key) !== this) return;
+      if (this.store.bytes + 2 * utf8(text) > EDITOR_MAX_BYTES) throw new EditorCapacityError();
+      this.store.drafts.delete(this.key);
+      this.baseline = Object.freeze({ sessionId: this.sessionId, artifactId: this.artifact.id, versionId: head.versionId, original: text });
+      this.text = text;
+      this.store.drafts.set(this.key, this);
+      this.phase = "ready";
+    } catch (error) {
+      this.phase = "error";
+      this.problem = error instanceof EditorCapacityError ? "capacity" : "load";
```
Nobody-Zhang, line 144: **A capacity failure during `load()` leaves a dead draft holding a slot that can never be retried or reopened.**

`open()` registers the editor in `drafts` before `load()` runs, so an `EditorCapacityError` here leaves it registered with `baseline === null` and `problem === "capacity"` — a state nothing can leave:

- it holds 1 of the 10 `EDITOR_MAX_DRAFTS` slots for the session;
- `open()` finds it by `sessionId + artifactId` and hands back the same dead object on every later attempt at that artifact;
- `load()` returns early at line 128 (`this.problem !== "load"`), and `editor-view.ts:62` hides `retry` for `"capacity"`;
- the only affordance offered is "Discard & reload", a confirm dialog about losing a draft that does not exist.

**Fixed**: a load that never produced a baseline releases its reservation — there is no draft to protect, and a later `open()` then builds a fresh editor. Covered by `a load that never produced a draft releases its reservation`, verified red without the fix.

Related, not changed: `ArtifactEditorStore.discard()` returns `true` unconditionally after a conditional delete, so it also reports success for the over-cap editor `open()` never registered.

frontend/src/features/execution/provenance.ts
```diff
@@ -71,38 +68,65 @@ function provRow(label: string, files: string[]): HTMLElement {
...
+let environmentRead: PendingRead<EnvSnapshot> | null = null;
+let messagesRead: PendingRead<{ messages: ChatMessage[] }> | null = null;
+function sameRead(left: ReadOwner | null, right: ReadOwner): boolean {
+  return !!left && left.key === right.key && left.request === right.request && left.frameId === right.frameId;
+}
+
+function readOwner(a: ArtifactRow): ReadOwner {
+  return { key: artifactMetadataCacheKey(a), request: _lineageReq.value, frameId: currentId.value };
+}
+function ownsRead(owner: ReadOwner): boolean {
```
Nobody-Zhang, line 86: **Reviewed and withdrawn — no change needed.** Recording it because it looks like a gap and is not.

`ReadOwner` carries `key` / `request` / `frameId` but not `_openGen`, and line 420 in this same file *does* consult `_openGen`, which reads like the guard being wired to one call site of several. It is deliberate: `selectProject` bumps the generation while the conversation stays open, so a read for the still-open artifact in the still-open frame remains valid, and `a sidebar-only navigation preserves the retained frame's in-flight read and retry` in `provenance.test.ts` pins exactly that. Adding the generation to `ownsRead` turns that test red.

Worth a one-line comment on `ReadOwner` saying the generation is excluded on purpose — the next reader will otherwise "fix" it too.

frontend/src/islands/viewer.ts
```diff
@@ -500,7 +448,34 @@ export function renderViewer(): void {
...
+      sessionId,
+      active,
+      close: () => { _editing.value = null; renderViewer(); },
+      reload: () => {
+        // A new load resolves the current head. Never reuse the stale baseline.
+        const current = dockArtifact.value;
+        if (current) dockArtifact.value = { ...current, version_id: undefined, latest_version_id: undefined };
+        renderViewer();
+      },
+      saved: (versionId) => {
```
Nobody-Zhang, line 469: **Saving an edit leaves an open Provenance panel on the previous version's lineage.**

The pre-PR save path was `if (provMode.value) callWindow("showProvenance", ...); else renderViewer();`. Extracting the editor kept only the `renderViewer()` half, and `renderViewer` → `renderProvenanceInto` reuses the cached payload: for a non-exact row `artifactMetadataCacheKey` is `[id, checksum || "unknown"]` and `dockArtifact` still holds the pre-save row, so `_lineageFor.value === artifactMetadataCacheKey(art)` still matches. Only `showProvenance` bumps `_lineageReq` and starts a fresh read.

Result: with the Provenance dock open, the panel keeps showing the previous version's producer / wrote / reads-inputs rows while the header already reflects the new version id.

**Fixed** by restoring the `provMode` branch. No unit coverage — `viewer.test.ts` has no harness for `renderViewer`'s editor callbacks; worth a scene in `tests/browser_editor.mjs`.

frontend/src/features/artifacts/editor.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+    this.emit();
```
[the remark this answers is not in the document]
Nobody-Zhang: Investigated and partly fixed in 0b336039 — but the stated mechanism doesn't hold, so the fix is different from the one implied here.

**The race doesn't exist.** There is no `await` between the `store.bytes + 2 * utf8(text)` check and the commit that sets `baseline`, so the Nth load always sees loads 1..N-1 already counted. Measured: three concurrent loads each wanting half the cap end with two `ready`, one refused, and `store.bytes` exactly equal to the cap. The 10-draft cap is also taken synchronously in `open()` before `load()` runs.

**The conclusion is right by a different route**, which I did reproduce: `bytes` counts committed drafts only, so ten simultaneous opens each stream up to `EDITOR_MAX_BYTES / 2` against a store reporting `0` — 40 MiB in flight under an 8 MiB cap — and a load opened against an already-full store still downloaded its whole allowance before being refused.

**Fixed:** `load()` now reads with the remaining committed budget as its limit, so a read nobody can afford stops at the first byte instead of after 4 MiB. An empty store still offers a whole draft's worth, so the common case is unchanged.

**Not fixed, deliberately:** I built the reservation this comment asks for — a `pending` counter claimed at the head read, released in a `finally` — and adversarial testing broke it three independent ways: a read that never settles holds its claim for the life of the page; "Discard & reload" during a load never releases it; and a version row that over-declares `size_bytes` locks out every other editor while it loads. All three surface as a `capacity` error the view offers no retry button for. That trades a bounded, transient over-read for an outage driven by unverified server metadata, so it isn't shipped.

Still true and now stated in the code rather than hidden: ten concurrent reads of honestly-tiny artifacts can stream past the cap if the server serves more than it declared, and `check()` reads with `EDITOR_MAX_BYTES` and no bound at all.

openai4s/host_dispatch.py
```diff
@@ -2152,7 +2154,13 @@ def _screen_tool_result(
         if not text or not text.strip():
             return result
         try:
-            verdict = scan_tool_result(text, source=src, cfg=self.cfg, use_llm=use_llm)
+            verdict = scan_tool_result(
+                text,
+                source=src,
+                cfg=self.cfg,
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in 9c592cb9 — and the cause is one level up from this line, so the fix is not here.

`HostDispatcher` is built **once per session from the boot config** and never sees a resolved one again. So this isn't a call site that was missed; it's every provider-reaching capability the dispatcher owns. Measured on a runner whose key exists only in settings (the documented Customize → Models install, which is why `_llm_cfg` exists at all):

```
dispatcher.cfg.llm.api_key == ""
host.llm                    -> MissingCredentialError
scan_tool_result(use_llm=1) -> verdict False, source "static"; the model was never called
```

So your impact statement is right — `OPENAI4S_SAFETY=llm` left only the static regex on untrusted web_fetch/MCP payloads — and the in-kernel `host.llm` port did not work at all on that install.

**Fix:** one line beside the per-turn delegation rewire in `_ensure_runtime`, whose docstring already stated the intent for its half ("Rewire per turn so delegated specialists inherit the currently selected model from the composer dropdown"). Delegated children have had `_dc.replace(self.cfg, llm=self._llm_cfg(st))` since that was written; the session's own ports never did. Placed there rather than at construction for the same reason that rewire is — it runs every turn, so a model changed mid-session reaches these ports without discarding the control plane.

After: the key resolves, `host.llm` reaches the provider, and this screener returns `source="llm"`. Mutation-verified — pinning the boot config back reddens the new test.

openai4s/host/llm.py
```diff
@@ -39,14 +60,98 @@ def _chat(self, *args, **kwargs) -> dict:
...
+        try:
+            parts = token_upper_bound_parts(
+                config.llm,
+                messages=spec.get("messages") or [],
+                max_tokens=spec.get("max_tokens"),
+            )
+        except Exception:  # noqa: BLE001 - an unpriceable call is not refused
+            parts = None
+        if parts is None:
+            return (0.0, 0.0)
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in e8bfbdfc. You are right, and I measured it: with `max_tokens: 0` on each item the previous commit reserved nothing and all 32 fan-out calls went through — 1280 tokens against a 100-token window, the whole overshoot back through one field the cell controls. After the fix, 1 call / 40 tokens.

Cause exactly as you describe: `token_upper_bound_parts` treats a present-but-zero `max_tokens` as no completion and declines to price the call, while every adapter resolves the cap as `max_tokens or cfg.max_tokens` and sends the configured one anyway. The projection now resolves the completion the way the wire does, so it prices what will actually be sent. Generous-limit and no-quota fan-outs are still untouched at 32/32, and the mutation (pricing the spec instead of the wire) reddens the new test.

On the other half of your summary — a wire the bound declines to certify, `responses` — I am deliberately leaving it reserving nothing, so a fan-out there is admitted exactly as it was before this gate existed. Substituting a number the bound module explicitly says cannot certify that wire would be the failure this PR has already corrected twice: an unverified value used as though it were verified. It is also not cell-controllable; it follows the operator’s provider choice. Stated in the code rather than left implicit.

openai4s/host/llm.py
```diff
@@ -41,12 +48,31 @@ def _chat(self, *args, **kwargs) -> dict:
...
-        response = self._chat(
-            spec.get("messages") or [],
-            config.llm,
-            max_tokens=spec.get("max_tokens"),
-            temperature=spec.get("temperature"),
-        )
+        # Before the request, not after: a refusal must not have spent anything.
+        # Outside the metering try below, because a swallowed gate is not a gate.
+        if self.quota_gate is not None:
+            self.quota_gate()
```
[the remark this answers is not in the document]
Nobody-Zhang: Fixed in c110ca4e, then hardened in e8bfbdfc — my first reply to this was going to decline it, on a premise that was wrong.

I said nothing here can price a call before making it. `token_upper_bound` does exactly that, and the Auto Budget has reserved against it all along. Reproduced your finding first: a 100-token `llm_input_tokens` window and 40 tokens a call, where a serialised caller is refused on its fourth — `host.llm([...]*32)` put all 32 through and recorded 1280 tokens, 12.8x the limit.

Fix: `LLMService` tracks in-flight promises and the gate is charged with what OTHER requests have promised, not with the caller own bound. Excluding the caller matters — these bounds carry a wire allowance and the transport retry ceiling, so one 40-token request prices at ~3.8k and charging it to itself let a 100-token window admit nothing at all. Excluded, a lone call is gated exactly as before, only siblings that start while it is in flight are held back, and a fan-out degrades to serial near a limit while nothing is throttled far from one.

Measured across three regimes: tight limit 32 calls/1280 tokens -> 1 call/40 tokens; generous limit all 32, untouched; no quota configured, all 32, inert (INV-1). `token_upper_bound_parts` splits the arithmetic per kind because a combined figure compared against either limit over-refuses; `token_upper_bound` is now its sum and was verified equivalent across 24000 input combinations, zero mismatches. The promise is in-process and released in a `finally` — a durable reservation would survive a crash and shrink the window with nothing able to clear it.

openai4s/host/llm.py
```diff
@@ -39,14 +60,108 @@ def _chat(self, *args, **kwargs) -> dict:
...
+        try:
+            parts = token_upper_bound_parts(
+                config.llm,
+                messages=spec.get("messages") or [],
+                max_tokens=completion,
+            )
+        except Exception:  # noqa: BLE001 - an unpriceable call is not refused
+            parts = None
+        if parts is None:
+            return (0.0, 0.0)
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in 286e8b64. Reproduced first: a deeply nested value under a message key the adapters drop gives `_projected` = (0.0, 0.0), and the batch put all 32 through — 1280 tokens against the 100-token window, outgoing payload 34 bytes, so 32 ordinary billed calls.

**Two layers**, because one is how the last two rounds went wrong — fix #2 own retry helper recursed, and so became this round trigger. Patching the newest shape is a losing game.

1. The pricer stops failing on cell-shaped input. `token_upper_bound_parts` counted nodes recursively and raised past ~500 levels; it is iterative now. Equivalent to the previous function across 2000 random shapes (zero mismatches), and depth 2000 prices at 87263 instead of raising.
2. A call nobody can price runs alone. Where a gate is installed, an unpriceable call takes a single slot for its duration instead of claiming nothing: the fan-out degrades to serial, the ledger catches up, the gate refuses on its own. No gate (CLI, tests, no owner row under INV-1) — nothing is serialised.

The layering is load-bearing: with the iterative count reverted, the bypass does not return to 32 calls, the slot holds it to 3 — what a serialised caller gets. Six vectors now all land on the baseline 1 call; generous-limit and no-quota fan-outs are untouched at 32/32.

**Two corrections to what I wrote last round.** The RecursionError was never `json.dumps` — the C encoder serialises depth 5000 fine; it was the pricer own `nodes()` walk. And my comment that non-finite floats are the only value a JSON frame can carry that `allow_nan=False` rejects, so normalising them closes the whole surface, was true about rejection and wrong about the surface: the bound can also raise, and a blanket `except` reads any raise as free. Your summary was right to call this residual risk.

Also fixed from the same review: `max_tokens: true` (the pricer `type(True) is int` is False while the wire `isinstance` is True — only bool sat in that gap), and the promise being released before the ledger row was written.

**Not fixed, measured and reported** — none is a hole in this gate, each is spend that does not pass through it: the in-flight promises are per-`LLMService` while the quota is per session owner, so N delegated dispatchers multiply it linearly (measured 1/2/4/8); `host.delegate` sub-agent turns never ask the gate; `host.compute`/`host.fold` reach no quota or ledger. The first needs the in-flight state keyed by the root frame governance resolves — a design change, and this same commit records what rushing shared mutable state into this path already cost me (a slot taken before the gate was asked and not returned on refusal, which hung the port).

openai4s/host/llm.py
```diff
@@ -39,14 +91,176 @@ def _chat(self, *args, **kwargs) -> dict:
...
+            # So a lone call is gated exactly as it was before, and siblings
+            # that start while it is in flight are the ones held back. Near a
+            # limit that degrades a fan-out to serial, which is the correct
+            # direction; far from one the promises are noise against the limit
+            # and nothing is throttled.
+            #
+            # The lock spans the check AND the claim, or two callers read the
+            # same pending total and both add to it.
+            self.quota_gate(
+                projected_input=self._inflight_input,
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in a85cd9cd. I reported this one myself last round as real, measured and unfixed — you are right that it should not stay that way.

The in-flight promises lived on the `LLMService` instance while the quota is billed to the session owner. Different keys, and `host.delegate` builds a dispatcher and so a service per child, so each child saw an empty in-flight total and was admitted against spend its siblings had already promised. Measured linear: 1/2/4/8 calls for 1/2/4/8 dispatchers.

**Fix:** the ledger moves into `storage/governance.py`, keyed by the resolved owner, project and kind — the same key `check_quota` reads. `reserve_session_llm_spend` checks and claims under one lock and returns the release, which `LLMService` runs in its `finally`. The per-owner solo slot for unpriceable calls moves with it.

A promise now carries a deadline and a stale one counts as lost. Releases run in a `finally`, so that is only for what a `finally` cannot cover (a killed thread) — but without it one lost release would shrink a TEAM window until the daemon restarted, and a claim outliving its operation is worse than the overshoot it stops. This path already demonstrated that once by hanging on a semaphore it forgot to return.

Regression test: two services on one owned session, both gating concurrently against a 100-token window. Mutation-verified with the defect itself — make nobody see anyone else promises and it fails with `[ok, ok]` and two provider calls.

Worth flagging that my first mutation of that test did NOT falsify it, and I did not treat that as the test being sound: it reset a dict outside the lock and simply raced. The second one says exactly what the defect said.

Still outside this gate and unchanged, as reported before: `host.delegate` sub-agent turns never ask it at all, and `host.compute`/`host.fold` reach no quota or ledger.

openai4s/agent/loop.py
```diff
@@ -938,12 +999,15 @@ def _account_abandoned_reply(reply: Mapping[str, Any]) -> None:
...
+                        self._record_frame_usage(usage)
                     if action_ledger is None:
                         return
+                    # Reached even with no counters, exactly as the Web loop's
+                    # ``account_abandoned_reply`` does: an unmeasured late reply
+                    # is recorded as *unknown* rather than as free, which is
+                    # what the team quota gate reads.
                     action_ledger.record_abandoned_usage(usage)
 
                 model: Any = ChatModel(
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in 9aa77f3d. This is the delegation gap I had been reporting as "outside this gate" — you narrowed it to one construction site, which is what it actually was.

`ChatModel` has taken a `quota_gate` since it was written and calls it before the send; the Web session passes one; the `ChatModel` built in `loop.py` for the CLI and for every delegated child never did. So a child ran its whole outer loop against a window nothing asked about, while compaction on the same run was gated. Mechanism present, wiring partial.

`_compaction_quota_gate` becomes `_session_quota_gate` and serves both sites. `enforce_session_llm_quota` resolves the frame to its session root first, which is what makes it correct for a child: a delegated frame has no owner row of its own and the quota belongs to the root owner. Inert without one (INV-1), so a single-user CLI run is unchanged — asserted by its own test.

**A check, not a reservation**, and the docstring says why: `ChatModel.complete` runs the provider call in a detached daemon thread whose reply can outlive the turn and still bill through `abandoned_reply`. A promise held across that could outlive its operation, which is the trade this PR has now refused three times. So concurrent children can still overshoot a nearly-exhausted window by the fan-out; what closes is starting at all against an exhausted one. Your impact statement is reduced rather than eliminated, and I would rather say that than imply otherwise.

Two of my own mistakes worth recording, both caught while verifying: the wiring guard first read `Agent.run` when the construction is in `_run_task` (red before any mutation), and once retargeted it SURVIVED its mutation — `quota_gate=self._session_quota_gate` appears twice in that function, so dropping either left the other matching. It counts both now, and removing either one reddens it.

openai4s/storage/governance.py
```diff
@@ -538,11 +637,367 @@ def check_quota(self, *, user_id: str, project_id: str | None, kind: str) -> Non
...
+#: admitted against the same unspent window. Measured linear, 1/2/4/8 calls for
+#: 1/2/4/8 dispatchers, up to the 48-child fanout cap.
+_INFLIGHT_LOCK = threading.Lock()
+_INFLIGHT: dict[tuple[str, str, str], list[list[float]]] = {}
+#: A promise this old is treated as lost rather than held. Releases run in a
+#: `finally`, so this is for the case a `finally` cannot cover -- a killed
+#: thread. Without it a single lost release would shrink a TEAM's window until
+#: the daemon restarted, which is the trade this codebase has already refused
+#: once: a claim outliving its operation is worse than the overshoot it stops.
+_INFLIGHT_TTL_S = 900.0
```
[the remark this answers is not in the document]
Nobody-Zhang: Confirmed and fixed in 4942733e — and this is a defect in the safety net I added in the very commit you reviewed, so thank you for catching it.

Verified your facts first: `total_timeout_s` defaults to 600 but validates 1..3600, and it is the WHOLE call deadline including retries (`CallState.deadline = monotonic() + total_timeout_s`). Against that, a flat 900s TTL expires a promise while its provider request is still running — the window then reads as unspent, the next fan-out is admitted against it, and the `finally` release removes an entry that is no longer there. Exactly as you describe, and default-600 installs are indeed unaffected.

The framing worth recording: a TTL shorter than the call it covers is not a safety net, it is the bug. That sweep exists to bound a LOST release, and instead it was discarding live promises.

**Fix:** the TTL is derived rather than fixed — `max(900, total_timeout_s + 120)`, passed down by `LLMService` from `config.llm.total_timeout_s` and forwarded by the dispatcher. A caller that states no deadline keeps the 900s floor rather than expiring immediately.

Three mutations verified: pinning 900 back reddens the new test at `0.0 != 500.0`, and dropping the service forwarding makes the measured `ttl_s` `None` where it is otherwise `3600.0`. The test asserts the property directly — a promise made with a 3600s deadline is still counted at +900s and +3500s, and is swept once the call it covers cannot still be running.

frontend/src/features/execution/provenance.ts
```diff
@@ -147,16 +179,24 @@ export function renderProvenanceInto(v: HTMLElement, a: unknown): void {
...
+    }
+    body.appendChild(notebookExportLink(art.root_frame_id));
     const list = (cells.value || []) as NotebookCell[];
     if (!list.length) body.appendChild(el("div", "dock-empty", t("prov.exec.noRecords")));
     list.forEach((e) => body.appendChild(cellNode(e)));
   } else if (provSub.value === "environment") {
     void renderProvEnvironment(body, art);
   } else if (provSub.value === "messages") {
-    void renderProvMessages(body);
+    void renderProvMessages(body, art);
```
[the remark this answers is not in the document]
Nobody-Zhang: Not applied — verified refuted, with a test added to pin the policy instead.

- The `exec` branch's check is not an ownership guard: `loadExecutionLog` discards its result unless `id === currentId.value` (`notebook/cells.ts:524`), so `cells.value` only ever holds the *current* session's cells. Without that branch the panel would render the viewer's own cells as the artifact's provenance. It is a data-availability guard.
- `GET /frames/{id}/messages` is guarded server-side, just not by the predicate you looked for. `_team_scope_guard` (`gateway.py:15574`, regex `/frames/([^/]+)(?:/.*)?`) runs before dispatch, resolves the root frame and answers **404** unless `session_visible_to`. `tests/test_team_session_ownership.py::test_cross_user_read_is_404_not_403` pins exactly that over a real socket.
- `_team_require_session_control` is the *control* predicate — its docstring says project visibility must not become write authority — and `tests/test_team_governance.py` pins this route as **not** a session-control mutation. I applied your suggestion as a mutation and measured it: a project member reading a teammate's project-visible session goes 200 → **403**, and the main session view (`messages/open.ts:257`) calls the same route, so it would break there too.
- Project members reading each other's project-visible sessions is the documented design (`may_control_session`: "Project visibility is intentionally broader than ownership. It can make a session readable"), so the panel is not disclosing anything the server does not already, deliberately, serve.
- The one factual observation is right: this PR did change the fetch to `root_frame_id`. That is the intended fix — an artifact's provenance panel must show the conversation that produced it, not the viewer's own.

Added `test_a_project_member_may_read_a_project_visible_session` in `947c9e84`: 200 with a non-empty positive control, then 404 once the owner sets the session private, verified by mutation.

PR #162, opened 2026-09-15
fix(windows): preserve Unicode paths and bridge Windows Ark CLI

difficulttopickaname:
## Summary

The Windows launcher can corrupt a package path containing Chinese characters under Windows PowerShell 5.1, causing installation or startup to use the wrong WSL path. It also cannot reuse an Ark CLI installed on Windows: the WSL service searches for `arkcli`, while the Windows executable is `arkcli.exe` and its location is not forwarded.

Decode captured WSL output as UTF-8 for the duration of each native invocation, restoring the previous console encoding afterward. Resolve the Windows Ark executable in the launcher, translate its path with `wslpath`, and forward it as one environment argument. An explicitly configured absolute Linux CLI path is also supported. Update the user guide and the README shipped inside the ZIP, and extend the native PowerShell regression coverage.

Targets `main` after the v0.3.0 preparation in #161 was merged. The diff contains four files; it preserves the release's version, frontend, and backend changes.

## Area

- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Test / harness
- [x] Documentation
- [x] Release / packaging

## Verification

Passed:

```text
powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File tests/test_windows_launcher.ps1
uv run pytest tests/test_release_docs.py tests/test_startup_no_implicit_install.py tests/test_startup_no_implicit_side_effects.py -q
python3 scripts/verify_release_tag.py v0.3.0
git diff --check
```

- Windows PowerShell 5.1: the added native-byte regression fails against the unpatched release launcher and passes against this commit. It also checks stderr/exit-code preservation, restoration of console encoding, CLI discovery, explicit overrides, and argv boundaries for paths with spaces.
- Real Ubuntu 24.04 / WSL2: a Unicode path returned by `wslpath` is corrupted before the patch and preserved afterward. With an installed Windows Ark CLI on the test process's PATH and no Ark-specific `WSLENV` entry, the old launcher forwards no CLI and the WSL lookup exits 127; the patched launcher forwards the translated executable path and its WSL existence check exits 0.
- No account credentials, provider requests, or inference calls are needed by the new regression tests.
- The three release/startup test modules pass (30 tests).
- These local checks passed before rebasing onto the merged main. All four changed files are byte-identical after the rebase.

Not run locally for this commit: the full Python offline suite and the full browser matrix. CI must validate the submitted head; the green release-branch baseline is not claimed as this commit's CI result.

Pending before release: the full v0.3.0 Windows ZIP rebuild and installed-package scientific workflow acceptance. Earlier candidate-package acceptance does not substitute for testing the final v0.3.0 artifact.

## Risk and Reviewer Focus

Review `Invoke-WslCaptureNative`, `Get-WslArkCliPath`, and `Get-WslBootstrapArgs`. Console encoding changes must stay scoped to the native call and restore on failure; CLI paths must stay single argv elements. A CLI must already be installed; the ZIP does not bundle or install Ark CLI. The launcher supports a native Windows `.exe` or an absolute WSL executable path.

## Checklist

- [x] The complete diff is reviewed, small, and focused.
- [x] No unrelated formatting or facade rewrites are included.
- [x] Relevant native tests and real WSL checks are listed above.
- [x] New default tests do not require WSL, credentials, network access, or a model.
- [x] No tests are deleted without replacements.
- [x] No core dependencies are added.
- [x] No secrets, private data, or machine-specific paths are included.
- [x] User-facing documentation matches the launcher behavior.


scripts/windows/openai4s.ps1
```diff
@@ -795,6 +819,7 @@ if ($BindHost.Contains(':')) {
 }
 
 $distro = Select-Distro
+$WslArkCli = Get-WslArkCliPath $distro
```
Nobody-Zhang: **An optional Ark CLI lookup now gates `stop`, `doctor`, `status` and `url`.**

`Get-WslArkCliPath` runs on every invocation, before the `Test-SandboxIndependentCli` management branch. Two consequences:

- An invalid `OPENAI4S_ARKCLI_PATH` makes `OpenAI4S.cmd stop` / `doctor` exit 1 with *"does not name an installed executable"*. Concrete case: set the variable, uninstall Ark CLI while a daemon is running, and the launcher can no longer stop that daemon. The script's own contract a few lines below is that management stays reachable when something else is broken.
- Every management command pays an extra `wsl.exe --exec wslpath` round trip for a value only a daemon process consumes.

Suggested fix: resolve it only when not taking the management path (e.g. `if (-not (Test-SandboxIndependentCli $Arguments)) { $WslArkCli = Get-WslArkCliPath $distro }`), and initialise `$WslArkCli = ''` with the other script-level settings.

docs/windows-wsl.md
```diff
@@ -317,6 +317,29 @@ carry a `managed-by-openai4s-windows-launcher` marker. Remove that marker before
 taking ownership of either file; later launches then preserve the complete
 file byte for byte.
 
+### Volcengine login
+
+Configure the model in Customize → Models. A CLI installed inside WSL is
```
Nobody-Zhang, line 340: **A Windows `arkcli.exe` on PATH silently shadows a CLI installed inside WSL, and the docs say the opposite.**

Implicit discovery forwards the `.exe` as `OPENAI4S_ARKCLI_PATH`, and `ArkCliBridge.executable()` (`openai4s/server/volcengine_arkcli.py:415`) checks that variable *before* `which("arkcli")`. So "a CLI installed inside WSL is discovered normally" only holds when no `arkcli.exe` exists on the Windows PATH.

Failure modes:
- A user who logged in with the WSL CLI and also has a standalone Windows `arkcli.exe` (another account, a stale login) gets the Windows identity and profiles without being told.
- With `[interop] enabled=false` in `/etc/wsl.conf`, the forwarded `.exe` cannot execute (`ENOEXEC` → *"Ark CLI could not be executed"*). A Volcengine login that worked before this PR stops working.

At minimum the docs should state the precedence and the escape hatch: an absolute Linux `OPENAI4S_ARKCLI_PATH` wins over discovery. Longer term, a fallback-only variable consulted after `which("arkcli")` would stop the override from happening at all.

scripts/windows/openai4s.ps1
```diff
@@ -543,6 +549,23 @@ function Assert-WslDataDir([string] $Value) {
     }
 }
 
+function Get-WslArkCliPath([string] $Distro) {
+    $configured = $env:OPENAI4S_ARKCLI_PATH
+    if ($configured -and $configured.StartsWith('/')) { return $configured }
+    $name = if ($configured) { $configured } else { 'arkcli.exe' }
+    $command = Get-Command $name -CommandType Application -ErrorAction SilentlyContinue | Select-Object -First 1
+    if ($command -and $command.Source.EndsWith('.exe', [StringComparison]::OrdinalIgnoreCase)) {
+        return (ConvertTo-WslPath $Distro $command.Source)
```
Nobody-Zhang: **An unconfigured `arkcli.exe` that WSL cannot translate aborts the whole launch, and the guidance blames the package folder.**

With `OPENAI4S_ARKCLI_PATH` unset, this calls `ConvertTo-WslPath`, whose failure path is `Stop-WithGuidance` → `exit 1`. If PATH contains a location `wslpath -a` cannot represent (a UNC tools share such as `\\fileserver\tools\arkcli.exe`), every `OpenAI4S.cmd` launch stops with:

```
WSL could not reach this folder: \\fileserver\tools\arkcli.exe
Unzip the package onto a local drive (for example C:\OpenAI4S).
```

The user never configured Ark CLI and the package is already on a local drive. An optional tool found by chance should be skipped with a note. Only an explicitly configured path should refuse, and its message should be about Ark CLI rather than the package.

scripts/windows/openai4s.ps1
```diff
@@ -190,14 +191,19 @@ function Invoke-WslCaptureNative([string[]] $WslArgs) {
...
     try {
         $ErrorActionPreference = 'Continue'
+        # WSL_UTF8 selects the bytes WSL writes; PowerShell 5.1 separately
+        # decodes native stdout with Console.OutputEncoding (often OEM).
+        [Console]::OutputEncoding = New-Object System.Text.UTF8Encoding($false)
         $output = @(& wsl.exe @WslArgs 2>&1)
         $code = $LASTEXITCODE
     } finally {
         $ErrorActionPreference = $previousPreference
+        [Console]::OutputEncoding = $previousEncoding
```
Nobody-Zhang: **In a host without a console, restoring the encoding throws after `ErrorActionPreference` is already back to `Stop`.**

On .NET Framework, the `[Console]::OutputEncoding` setter calls `SetConsoleOutputCP` and throws *"The handle is invalid"* when the process has no console (PowerShell ISE, a GUI parent using `CREATE_NO_WINDOW`). Inside the `try` block (EAP=`Continue`), the UTF-8 set fails without stopping the script. The `finally` block then restores `Stop` **first** and runs `[Console]::OutputEncoding = $previousEncoding`, which throws again and is now script-terminating. The launcher dies in `Get-WslDistros`, before it selects a distribution, where it previously worked.

Suggested fix: wrap both the set and the restore in `try { } catch { }`, restore only when the set succeeded, and restore the encoding before `$ErrorActionPreference`.

scripts/windows/openai4s.ps1
```diff
@@ -552,6 +575,7 @@ function Get-WslBootstrapArgs([string] $Distro, [string] $BootstrapLinux, [strin
     if ($WslDataDir) {
         $wslArgs += "OPENAI4S_DATA_DIR=$WslDataDir"
     }
+    if ($WslArkCli) { $wslArgs += "OPENAI4S_ARKCLI_PATH=$WslArkCli" }
```
Nobody-Zhang: **The forwarded `.exe` runs across the WSL interop boundary, where the bridge's child-env contract is not delivered, and nothing exercised that path.**

`ArkCliBridge` runs the CLI with `_command_env()`: `ARKCLI_CALLER_TYPE/NAME/SKILL_NAME` plus allowlisted `HTTP(S)_PROXY`. A Windows process started through interop only receives Linux variables listed in `WSLENV`, and `_child_env` drops `WSLENV`. `arkcli.exe` therefore sees neither the caller identity nor `OPENAI4S_WSL_PROXY`. (Compare `security/windows_dpapi.py`, which deliberately manages `WSLENV` and `start_new_session` for its interop child.)

The PR's WSL evidence is that "the WSL existence check exits 0". No real bridge call was made through the daemon: `auth whoami`, `auth login --no-browser --code …`, or `api apikey.list --params {"PageSize":100}`, whose JSON argv also has to survive the Linux-argv → Windows command-line conversion. Please run at least one `--params` call and one login round trip on real WSL before claiming the Windows login is reused.

scripts/windows/openai4s.ps1
```diff
@@ -190,14 +191,19 @@ function Invoke-WslCaptureNative([string[]] $WslArgs) {
...
     # to continue, then restore the script-wide fail-fast policy.
     $previousPreference = $ErrorActionPreference
+    $previousEncoding = [Console]::OutputEncoding
     $output = @()
     $code = 1
     try {
         $ErrorActionPreference = 'Continue'
+        # WSL_UTF8 selects the bytes WSL writes; PowerShell 5.1 separately
+        # decodes native stdout with Console.OutputEncoding (often OEM).
+        [Console]::OutputEncoding = New-Object System.Text.UTF8Encoding($false)
```
Nobody-Zhang: **The UTF-8 decoding fix covers only one of the two native-read paths.**

Only `Invoke-WslCaptureNative` switches the decoder. `Invoke-Bootstrap` (L634, `& wsl.exe @wslArgs 2>&1 | ForEach-Object { Write-Host ... }`) still decodes `WSL_UTF8` output with the OEM code page. Preflight, install and every non-rewritten CLI passthrough (`OpenAI4S.cmd doctor`, `setup`, …) therefore print localized `wsl.exe` diagnostics and Unicode paths as mojibake. The NAT-mode localhost-proxy warning, for example, shows up garbled on a cp936 console on every launch. `Invoke-AppCli` now also decodes the same command's output differently depending on whether the bind host was rewritten (capture path) or not (streaming path).

Suggest extracting the set/restore into a small pair of helpers and using them in both places.

scripts/windows/openai4s.ps1
```diff
@@ -543,6 +549,23 @@ function Assert-WslDataDir([string] $Value) {
     }
 }
 
+function Get-WslArkCliPath([string] $Distro) {
+    $configured = $env:OPENAI4S_ARKCLI_PATH
+    if ($configured -and $configured.StartsWith('/')) { return $configured }
+    $name = if ($configured) { $configured } else { 'arkcli.exe' }
+    $command = Get-Command $name -CommandType Application -ErrorAction SilentlyContinue | Select-Object -First 1
```
Nobody-Zhang: **A configured Windows path goes through `Get-Command`, which treats path characters as wildcards and picks the first match in PATH order.**

- `Get-Command` interprets `[`, `]`, `*` and `?` as wildcard characters. `OPENAI4S_ARKCLI_PATH='D:\Tools [x64]\arkcli.exe'` names an existing file, but `[x64]` is read as a character class, nothing matches, and the launch stops with *"does not name an installed executable"*.
- For a bare name (`OPENAI4S_ARKCLI_PATH=arkcli`), `Select-Object -First 1` takes the first Application in PATH order. If `%APPDATA%\npm` comes earlier, `arkcli.cmd` is selected and rejected even though `arkcli.exe` is also on PATH.

The guidance message also omits the rejected value, while the neighbouring port and Fake-IP checks print `Current value: …`.

tests/test_windows_launcher.ps1
```diff
@@ -84,4 +84,60 @@ Assert-AppPage 'http://172.20.0.2:8760/?token=synthetic'
...
+        if ([Console]::OutputEncoding.CodePage -ne 437) { throw 'console encoding was not restored' }
+        if ($ErrorActionPreference -ne 'Stop') { throw 'error preference was not restored' }
+    }
+    Copy-Item -LiteralPath (Join-Path $testRoot 'wsl.exe') -Destination (Join-Path $testRoot 'arkcli.exe')
+    function ConvertTo-WslPath($Distro, $WindowsPath) {
+        if ($WindowsPath -ne (Join-Path $testRoot 'arkcli.exe')) { throw 'wrong Windows CLI selected' }
+        return '/windows tools/arkcli.exe'
+    }
+    $env:OPENAI4S_ARKCLI_PATH = ''
+    $WslArkCli = Get-WslArkCliPath 'Ubuntu'
```
Nobody-Zhang, line 148: **Only the happy paths of `Get-WslArkCliPath` are covered.**

None of the branches that decide whether a launch is *refused* are exercised:
- configured-but-missing → guidance (the one fatal branch)
- a configured `.cmd` → rejected
- nothing on PATH → `''` (and not fatal)
- implicit discovery whose translation fails → skipped, not fatal (see the comment on L558)

The native quoting of the forwarded value through `Start-Bootstrap`/`ConvertTo-NativeArgument` is not covered either; only `Get-WslBootstrapArgs`' array form is.

tests/test_windows_launcher.ps1
```diff
@@ -84,4 +84,60 @@ Assert-AppPage 'http://172.20.0.2:8760/?token=synthetic'
...
 
+# Exercise actual native UTF-8 bytes under an OEM console encoding. This runs
+# in Windows CI without WSL and catches the failure a mocked string cannot.
+$testRoot = Join-Path ([IO.Path]::GetTempPath()) ('openai4s-native-' + [Guid]::NewGuid().ToString('N'))
+$savedPath = $env:PATH
+$savedEncoding = [Console]::OutputEncoding
+$savedArkCli = $env:OPENAI4S_ARKCLI_PATH
+try {
+    New-Item -ItemType Directory -Path $testRoot | Out-Null
+    Add-Type -OutputAssembly (Join-Path $testRoot 'wsl.exe') -OutputType ConsoleApplication -TypeDefinition @'
```
Nobody-Zhang: **`Add-Type -OutputType ConsoleApplication` is not supported on PowerShell 7.**

Since PowerShell 6, `ConsoleApplication`/`WindowsApplication` output types raise a terminating error. CI stays green only because it invokes `powershell.exe` explicitly. `pwsh -File tests/test_windows_launcher.ps1` passes every earlier contract and then fails on an error unrelated to the launcher, while the final line (`PowerShell $($PSVersionTable.PSVersion): launcher contracts passed`) reads as if the script were version-agnostic. Suggest skipping the native-bytes block, with a message, when `$PSVersionTable.PSEdition -ne 'Desktop'`.

tests/test_windows_launcher.ps1
```diff
@@ -84,4 +84,60 @@ Assert-AppPage 'http://172.20.0.2:8760/?token=synthetic'
...
+    if ((Get-WslArkCliPath 'Ubuntu') -ne '/opt/ark/bin/arkcli') { throw 'explicit WSL path lost precedence' }
+    $env:OPENAI4S_ARKCLI_PATH = Join-Path $testRoot 'arkcli.exe'
+    if ((Get-WslArkCliPath 'Ubuntu') -ne '/windows tools/arkcli.exe') { throw 'explicit Windows path was not translated' }
+} finally {
+    $env:PATH = $savedPath
+    $env:OPENAI4S_ARKCLI_PATH = $savedArkCli
+    [Console]::OutputEncoding = $savedEncoding
+    $resolvedTestRoot = [IO.Path]::GetFullPath($testRoot)
+    $tempRoot = [IO.Path]::GetFullPath([IO.Path]::GetTempPath()).TrimEnd('\') + '\'
+    if (-not $resolvedTestRoot.StartsWith($tempRoot, [StringComparison]::OrdinalIgnoreCase)) { throw 'unsafe test cleanup path' }
```
Nobody-Zhang: **A throw in `finally` replaces the assertion that actually failed.**

If the `try` block throws (for example *"native UTF-8 path was corrupted"*) and `Remove-Item` then fails under `$ErrorActionPreference = 'Stop'` (an AV scanner still holding the just-executed fixture `wsl.exe`), or the path guard throws, the run reports only the cleanup error and the real regression message is lost. Cleanup should be best-effort (`-ErrorAction SilentlyContinue`, skip rather than throw on the path guard), so the original exception propagates.

docs/windows-wsl.md
```diff
@@ -317,6 +317,29 @@ carry a `managed-by-openai4s-windows-launcher` marker. Remove that marker before
...
+$env:OPENAI4S_ARKCLI_PATH = 'C:\Tools\Ark\arkcli.exe'
+.\OpenAI4S.cmd
+```
+
+An absolute Linux path also works. No manual `WSLENV` setting is needed.
+The executable must already be installed; the ZIP does not bundle Ark CLI.
+Restart OpenAI4S after changing the path. A Windows npm `.cmd` launcher is
+not a WSL executable: use its native `arkcli.exe` or install the CLI in WSL.
+
+在「自定义 → 模型」中配置火山模型。启动器会发现 Windows PATH 中的
```
Nobody-Zhang: **The Chinese paragraph was placed in the English half of the document.**

`docs/windows-wsl.md` is organised as `## 中文` (sections 1–6) followed by `## English`. This Chinese paragraph sits under the English `### Volcengine login` heading, and the 中文 walkthrough gets no corresponding section. Chinese readers following sections 1–6, the package's main audience given the Tsinghua mirror defaults, will never find it. Suggest moving it into the 中文 half, for example as a subsection of `### 3. 启动 Windows 包`.

scripts/build_windows_zip.sh
```diff
@@ -220,6 +220,11 @@ OpenAI4S $VERSION - first launch on Windows
 
      openai4s serve --port 8760 --no-browser --detached
 
+   Volcengine login: the launcher discovers arkcli.exe on Windows PATH.
```
Nobody-Zhang: **The READ ME FIRST note reads as if it applied to the Ubuntu-terminal route it sits under.**

It directly follows `openai4s serve --port 8760 --no-browser --detached`, which runs from an Ubuntu terminal, and tells the reader to set `OPENAI4S_ARKCLI_PATH` to a *full Windows path*. Discovery and path translation happen only in `OpenAI4S.cmd`. A user who exports `OPENAI4S_ARKCLI_PATH='C:\Tools\Ark\arkcli.exe'` in Ubuntu and runs `openai4s serve` gets *"Ark CLI is not installed"*, because that is not a Linux file. Suggest scoping it: "When started through OpenAI4S.cmd, …; from a WSL terminal use an absolute Linux path."

openai4s/server/volcengine_arkcli.py
```diff
@@ -120,6 +120,16 @@ def _command_env(argv: Sequence[str]) -> dict[str, str]:
...
+    if os.name != "nt" and str(argv[0]).lower().endswith(".exe"):
+        # A Windows arkcli.exe reached through WSL interop receives only the
+        # Linux variables WSLENV names, and the allowlist never copies the
+        # daemon's own WSLENV. Name exactly the caller identity and the
+        # already-allowlisted proxy settings a Linux arkcli would see.
+        child_env["WSLENV"] = ":".join(
+            f"{key}/u"
+            for key in child_env
+            if key.startswith("ARKCLI_") or key.upper().endswith("_PROXY")
+        )
```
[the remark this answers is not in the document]
Nobody-Zhang: Switched the WSLENV suffix to `/w` in eba0a29e: Microsoft's flag table confirms `/w` is "only when running Win32 from WSL" and `/u` the opposite; `/p` deliberately avoided so proxy URLs are not path-translated.

tests/test_volcengine_connector.py
```diff
@@ -923,6 +924,51 @@ def test_the_offline_suite_never_resolves_a_real_arkcli(tmp_path, monkeypatch):
...
+    env = _command_env(("/mnt/c/Tools/Ark/arkcli.exe", "api", "apikey.list"))
+
+    names = env["WSLENV"].split(":")
+    assert {"ARKCLI_CALLER_TYPE/u", "ARKCLI_SKILL_NAME/u", "HTTPS_PROXY/u"} <= set(
+        names
+    )
+    assert all(
+        name.startswith("ARKCLI_") or name.upper().endswith("_PROXY/U")
+        for name in names
+    )
```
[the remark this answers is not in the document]
Nobody-Zhang: The test now expects `/w` on every WSLENV entry (and would fail on `/u`), in eba0a29e.

scripts/windows/openai4s.ps1
```diff
@@ -543,6 +573,49 @@ function Assert-WslDataDir([string] $Value) {
...
+            if (Test-Path -LiteralPath $configured -PathType Leaf) { $source = $configured }
+        } catch { }
+    } else {
+        # The first arkcli.exe in PATH order, not the first Application: an
+        # npm arkcli.cmd earlier on PATH would otherwise hide it.
+        $name = if ($configured) { $configured } else { 'arkcli.exe' }
+        $command = Get-Command $name -CommandType Application -All -ErrorAction SilentlyContinue |
+            Where-Object { $_.Source.EndsWith('.exe', [StringComparison]::OrdinalIgnoreCase) } |
+            Select-Object -First 1
+        if ($command) { $source = $command.Source }
```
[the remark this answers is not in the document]
Nobody-Zhang: A configured bare name containing `*`, `?`, `[` or `]` now skips Get-Command and is refused with the guidance; `arkcli*` and `*` added to the launcher test's refusal cases (eba0a29e). Implicit `arkcli.exe` discovery is unchanged.

PR #169, opened 2026-09-17
fix(release): validate evidence and resolve desktop UI regressions

Nobody-Zhang:
## Summary

Fix four release-readiness regressions:

- Reject a replacement release asset even when its mutable `SHA256SUMS` is updated. Upload and final publication now cross-check provenance, SBOM, sealed evidence, source identity, quality/build receipts, and the Windows package's exact Linux payload. The manual `--only publish` path is covered too.
- Generate Linux desktop entries with the required quoting and escaping so relocated paths containing spaces or punctuation launch correctly.
- Keep the latest project navigation when earlier project/session/folder requests finish later, including failed responses, same-project navigation, and a menu selection while a project route is still loading. Menu filtering preserves the current conversation’s in-flight history loading.
- Clear capability receipts when the selected model changes and discard results from its previous in-flight probe. Rebuild and include the production workbench assets.

## Area

- [x] Web UI / frontend
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [x] Test / harness
- [x] Security-related change
- [x] Release / packaging

## Verification

Tested on macOS arm64 with Python 3.13 and the locked dev/science environment. Regressions were reproduced before the fixes.

- Full offline suite: 9,292 passed / 30 skipped (564.81 s).
- Frontend: 88 files / 788 tests passed; TypeScript and Vite production build passed.
- Full pre-commit, mypy, directory README coverage, source-secret scan, and all 38 offline PR harness scenarios passed.
- Fresh wheel installed without dependencies outside the checkout: 13 module imports and all 604 Skills passed. Built wheel/sdist resource verification passed.
- Rebuilt wheel ran in an isolated daemon: Chromium workbench smoke and 30/30 Chromium/Firefox/WebKit matrix checks passed on the final bundle. Admission-response-loss recovery and sandbox-preview suites also passed. The first smoke attempt hit the 30-second cold Matplotlib font-cache timeout; unchanged rerun and the final-bundle run both passed.
- Both reported UI reproductions and the Back/menu race passed against the rebuilt production bundle; a delayed conversation-history response still completed after menu filtering. The navigation test delays a real API response; the model-receipt test explicitly uses fixture model responses.
- Eight Linux installer cases execute the generated desktop command and check CLI links, icons, uninstall, and retained user data.
- The new consistency gate rejects a real mismatched artifact set without changing it.

Commands include:

```text
uv run --no-sync pytest -n auto --maxprocesses=4 --dist loadfile
uv run --no-sync pre-commit run --all-files
uv run --no-sync python -m harness.cli run --tier pr --offline
uv run --no-sync python scripts/check_directory_readmes.py
uv run --no-sync python scripts/source_secret_scan.py
npm test --prefix frontend
uv build --wheel --no-sources --out-dir <temporary-dist>
uv build --sdist --no-sources --out-dir <temporary-dist>
python scripts/verify_release_artifacts.py <temporary-dist>
node tests/browser_smoke.mjs
node tests/browser_matrix.mjs
node tests/browser_sandbox_preview.mjs
```

Not run: native Linux/Windows runtime or Docker validation on this Mac, full Linux desktop science-stack rebuild, live-provider calls, or signing/publication. Browser and installer tests use isolated test data.

## Risk and Reviewer Focus

Review `scripts/release_consistency.py` and its upload/finalize call sites, the Linux installer's two escaping layers, and navigation/probe ownership across asynchronous calls. Manual recovery without a stage attestation establishes internal consistency, not independent producer authenticity.

Existing mismatched draft assets still need their original verified bytes restored, or fixes must be built through CI under a new immutable tag. This PR does not rewrite old receipts, move tags, replace published Python distributions, or publish a release.

## Checklist

- [x] Full diff reviewed; changes are limited to these four regressions and their validation/docs.
- [x] No broad rewrites of compatibility/composition facades; core engine/kernel/server facades were not changed.
- [x] Workbench source and rebuilt `dist/` are included together; legacy `app.js`, vendor assets, and captured fixtures were not reformatted.
- [x] Meaningful offline regressions added; no tests deleted and no live services/secrets required by the default suite.
- [x] No hard third-party core dependency added.
- [x] No credentials, private data, local absolute paths, or internal assignments included.
- [x] Security-sensitive release checks have synthetic mutation/refusal tests.
- [x] Docs describe actual guarantees; affected directory documentation is synchronized in English and Chinese.

---

> [!NOTE]
> **Medium Risk**
> Changes async navigation and wizard state ownership in the workbench; incorrect guards could drop valid UI updates, while the documented release checks affect maintainer publish workflows.
> 
> **Overview**
> Documents **release asset consistency checks** (provenance/SBOM/evidence/Windows–Linux payload alignment before upload and publish, plus recovery guidance) and **Linux `.desktop` `Exec=` quoting** for paths with spaces or punctuation.
> 
> **Onboarding** now treats capability receipts as tied to model identity (`profileId`, provider, model, base URL): changing identity clears the prior test receipt/decision, `choosePath` always cancels in-flight probes, Continue on an existing profile refreshes edited profile fields before advancing, and late probe results from a superseded model cannot overwrite the current test.
> 
> **Project navigation** bumps a filter generation on menu selection and gates `loadSessions` / `loadFolders` / `openProject` continuations on `_openGen` (and filter version) so stale project opens or list responses cannot replace the active session or sidebar data; new navigation tests cover superseded clicks and menu filtering while a conversation load stays owned.
> 
> Includes an updated **production Web UI bundle** (`index-Cn5c4S-o.js`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 929f5d9e6d38451888a02b89aba98bda4e7b861b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/release_pipeline.py
```diff
@@ -1823,6 +1824,7 @@ def step_upload(self) -> StepResult:
             return StepResult(
                 "upload", True, f"would upload {len(self.assets)} asset(s)"
             )
+        self._verify_consistency(self.assets)
```
Nobody-Zhang, line 1848: **The new gate rejects staging reruns in the same directory (regression)**

`step_assets` collects assets by the `.zip` suffix and version number. Both `openai4s-<v>-evidence.zip` and `-evidence-stopped.zip`, which the pipeline itself generates, match those criteria. On a rerun, they are therefore treated as distributions: they become provenance subjects and SBOM references, but no build receipt covers them.

Reproduction: `gh release upload` returns a 502 once → the run stops at `upload`, leaving two zip files in the directory → rerun in place → `provenance.intoto.json disagrees with the release assets: ... unexpected ['openai4s-0.2.0-evidence.zip'] ...; restore the verified assets`. Leaving only the stopped zip produces `sealed build receipts disagrees ... missing [...-evidence-stopped.zip]` instead. The same two runs succeed on `origin/main`. The error message also points the operator toward "tampered assets", although nobody has modified any assets.

The PR works around this in the test with `unlink()` (`tests/test_release_pipeline.py:1126`). `step_assets` should exclude these two self-generated filenames, and the test should again assert that leftover files do not block a rerun. The CI attach job uses a fresh directory each time and is unaffected; this affects the local/manual recovery path promised in the documentation.

scripts/release_consistency.py
```diff
@@ -0,0 +1,266 @@
...
+            or len(sources) != 1
+            or sources[0]["digest"].get("sha1") != source_sha
+        ):
+            raise ReceiptError(
+                "provenance source SHA differs from the sealed release report"
+            )
+        quality = _document(
+            archive.read("artifacts/quality-receipt.json"), "sealed quality receipt"
+        )
+        verify_receipt_document(quality, expected_sha=source_sha)
```
Nobody-Zhang: **Manual `--only publish` now depends on which checkout runs it (a regression that occurs after the version number has already been taken on PyPI)**

`verify_receipt_document` compares the sealed quality receipt against `manifest_digest()` / `SCHEMA_VERSION` from **the checkout currently running the script**. Staging runs at the frozen SHA, so these naturally agree. However, the manual recovery described in `release.yml:853-856` and the docs (running `--only publish` manually after PyPI publication succeeds but finalize fails) is typically performed from the current `main`. If any gate's argv has changed since the tag, it produces:

`release consistency verification failed: quality receipt was produced against a different gate manifest (85de3a8d36f3 != ffffffffffff)`

The same operation can publish on `origin/main`. The rejection itself is correct, but neither the error message nor the new documentation paragraph tells the operator to run from a checkout of the release tag. Suggested fix: catch `GateManifestError` and, only when the manifest digest / schema differs from this checkout, append guidance to "run from a checkout of release commit `<sha>` (tag `v<version>`)". Document this prerequisite in `docs/release-validation.md` as well.

frontend/src/features/sessions/load.ts
```diff
@@ -225,15 +228,20 @@ export async function loadFolders(): Promise<void> {
...
+    if (!current()) return;
     folders.value = [];
   }
 }
 
 export async function loadSessions(): Promise<void> {
-  const scope = project.value ? `&project_id=${encodeURIComponent(project.value)}` : "";
+  const pid = project.value;
+  const gen = _openGen.value;
+  const current = () => project.value === pid && _openGen.value === gen;
```
Nobody-Zhang: **List freshness is tied to the view generation, silently discarding refreshes within the same project**

`_openGen` represents "who owns the current view" (`openConversation`, `showDashboard`, `openProject`, and `clearConversationChrome` bump it), not "which project this list belongs to". Project P's conversation list remains valid as long as `project.value === P`; opening a conversation does not make it incorrect. Moreover, `openConversation` reloads only when `sessions.value` is empty (`messages/open.ts:508`), so nothing retries the discarded refresh.

Reproduction: `deleteSession(Y)` → while it is in `await loadSessions()`, the user clicks another row in the same project → `_openGen` changes → `if (!current()) return;` → the deleted Y remains in `sessions.value`. Then `deleteSession` takes `ss[0]` and may open the frame it just deleted (404). The same applies at the end of `send()`, in `commitTitle`, to the menu cancel's `void loadSessions()`, and to debounced WS refreshes. `loadMoreSessions` leaves `sessionPages` incremented without adding any rows. A scratch Vitest test fails on the PR head and passes on `origin/main`.

Changing both guards to compare only `project.value === pid` still passes all 10 navigation tests in this PR—the `_openGen` part is neither covered by those tests nor necessary for those behaviors.

tests/test_release_pipeline.py
```diff
@@ -1937,8 +2091,7 @@ def test_finalize_refuses_a_draft_rewritten_to_drop_its_distributions(assets):
     assert report["stopped_at"] == "publish"
     assert report["published"] is False
     detail = report["steps"][-1]["detail"]
-    assert "do not carry the same" in detail
-    assert "the draft is missing" in detail
+    assert "disagrees with the release assets" in detail
```
Nobody-Zhang, line 2579: **The reverse PyPI anchor (`absent_from_draft`) has lost its only regression coverage**

These two tests originally verified the direction in `step_publish` where the draft is missing a distribution that exists on PyPI. The evidence is now sealed over the complete asset set, so the evidence chain still references the removed file and the consistency gate rejects it first. That is a good rejection, but it proves nothing about the anchor.

Mutation check: change `if missing_on_pypi or absent_from_draft:` to `if missing_on_pypi:`. The entire `tests/test_release_pipeline.py` still passes on the PR head; on `origin/main`, the same mutation makes exactly these two tests fail.

The anchor is still necessary: the evidence chain is unsigned, and someone who can rewrite the draft can also issue new receipts and reseal it. A draft that is consistently resealed with only the sdist missing passes `verify_release_consistency` completely and is stopped only by this branch (`the draft is missing ['openai4s-0.2.0.tar.gz']`); with the mutation, it prints `v0.2.0 is public`. Add a test in this form (remove the sdist → `_receipt_dist` → `_write_checksums` → finalize, asserting "the draft is missing"), and update the docstrings of these two tests to identify the check that now rejects their inputs.

scripts/release_consistency.py
```diff
@@ -0,0 +1,266 @@
...
+        raise ReceiptError(f"{label} is not valid JSON: {error}") from error
+    if not isinstance(document, dict):
+        raise ReceiptError(f"{label} must be a JSON object")
+    return document
+
+
+def _match(actual: Any, expected: Mapping[str, str], label: str) -> None:
+    if not isinstance(actual, dict):
+        raise ReceiptError(f"{label} has no artifact digest map")
+    missing = sorted(set(expected) - actual.keys())
```
Nobody-Zhang, line 68: **Most checks in this gate can be deleted without failing the suite, including the direction that permits publication**

Deleting individual checks in `release_consistency.py` on the PR head leaves `tests/test_release_pipeline.py` green for 21 of the 26 checks. The most significant is this line: replace it with `missing = []`, add `OpenAI4S-0.2.0-extra-tool.zip` to a staged draft, refresh `SHA256SUMS`, and run `--only publish` without an attestation → `ok=True published=True`, while the suite still reports 109 passed. In other words, adding an asset with no supporting evidence has no test coverage at all; the existing tests cover only removal (`extra`) and replacement (`changed`).

Two other new tests do not pin down the checks they are named after:
- `test_finalize_checks_distribution_documents_after_checksum_refresh`: deleting both provenance/SBOM `_match` calls still passes—the later "sealed release report ... changed ['sbom.cdx.json']" message satisfies `assert document in detail`. It should assert `f"{document} disagrees with the release assets"`.
- `test_staging_rejects_a_windows_payload_from_another_linux_build`: the fixture payloads have different lengths, so only the size precheck is triggered. Disabling `if actual != digests[linux]` leaves the suite green. A same-length payload case is needed.

The PR checklist says "Security-sensitive release checks have synthetic mutation/refusal tests". Add a parameterized table with one tampering case per check, each asserting its own rejection message, and verify each check by deleting it in turn.

frontend/src/features/sessions/projects.ts
```diff
@@ -307,17 +312,23 @@ export async function openProject(id: string): Promise<void> {
   // parked on an await (an upload-created session about to open its
   // conversation, a resume watchdog) sees a stale token and stands down instead
   // of yanking the view back to where it started.
-  _openGen.value = (_openGen.value || 0) + 1;
+  const gen = (_openGen.value || 0) + 1;
+  const filterVersion = projectFilterVersion;
+  _openGen.value = gen;
   await loadProjects();
+  if (_openGen.value !== gen || projectFilterVersion !== filterVersion) return;
```
Nobody-Zhang: **`openProject` retires the current conversation's generation on entry, but leaves no one to take over when it yields**

Line 317 synchronously bumps `_openGen`, so every read owned by the currently open conversation X (history restoration, loading earlier messages, the resume watchdog, and a send being prepared) immediately considers itself stale and exits. Previously, `openProject` always reached `openConversation`, so the retired generation always had a new owner; now a menu filter can make it return early here (or at line 323): `_openGen` = g0+1, `currentId` is still X, and neither `openConversation` nor `newSession` has been called.

Reproduced with the real modules: `_msgEarlierLoading` remains true → the button stays on "Loading…" permanently and rejects every subsequent click (only `openConversation` or `resetSessionScoped` clears it); history that was still loading is never rendered, and `HistoryLoadStatus` does not show Retry because the generation no longer matches; the watchdog performs 0 polls in 10 seconds. Entry points: popstate returning to a project route, or clicking the menu after the `createProject` modal is closed during an await; the dashboard row-click variant returns after `showWorkspace()`, leaving a visible workspace with no conversation. The window requires a `/projects` or `/frames` request to still be pending, so this is a narrow race.

The comment here says "must not retire that conversation's reads", but the new test records `conversationGeneration` only **after** `openProject("A")` (lines 91–92), so it already sees the bumped value and cannot observe this. One possible fix: when yielding, if `_openGen.value === gen && currentId.value`, reopen the current conversation to reclaim the view (consistent with the watchdog reopening the same fid)—but this changes the URL/triggers a redraw and requires rewriting the assertions in these two menu cases. That is a product tradeoff, so I have not made the change.

scripts/release_pipeline.py
```diff
@@ -2095,8 +2097,19 @@ def _revalidate_draft_from_checksums(self) -> dict[str, str]:
                         f"digest ({actual[:12]} != {digest[:12]}); refusing to "
                         f"publish"
                     )
+            self._verify_consistency([Path(temp) / name for name in expected])
         return checked
 
+    def _verify_consistency(self, assets: Sequence[Path]) -> None:
```
Nobody-Zhang: **finalize accepts `--source-sha` / `--workflow-run-id` but compares neither**

`_verify_consistency` only forwards `version` and `required_kinds`. The `expected_sha` that `_verify` uses to validate the sealed quality receipt and build receipts is read from `release-report.json` inside **the evidence zip being validated itself**; the return value of `verify_build_receipts` is discarded, and the `workflow_run_id` recorded in the receipts is never compared against anything. By contrast, staging's `step_verify` enforces both (`expected_sha=self._frozen_sha()`, and `recorded_run != self.workflow_run_id → ReleaseError`).

Reproduction: finalize a draft whose evidence chain consistently points to commit `bbbb…`, run `666`, with `--only publish --source-sha aaaa… --workflow-run-id 7100` (without attestation) → `ok=True, published=True`. The hosted finalize job already passes `--workflow-run-id "$GITHUB_RUN_ID"`, so this anchor is available but unused.

With attestation, the digests are transitively pinned, so this is a hardening gap rather than a vulnerability; however, "flags are supplied but ignored" lacks parity with staging and contradicts the claim of "same cross-checks". I suggest adding optional `expected_sha` / `workflow_run_id` parameters to `verify_release_consistency`, enforced only when supplied; pass the original `self.source_sha` value rather than `_frozen_sha()` (the latter falls back to the operator's HEAD when no flag is supplied).

scripts/release_consistency.py
```diff
@@ -0,0 +1,266 @@
...
+) -> None:
+    """Require the complete staged evidence chain, including manual recovery.
+
+    The evidence is a pre-upload snapshot: its report covers distributions,
+    SBOM and provenance, but cannot contain its own hash or SHA256SUMS. Build
+    receipts are carried inside that archive, not as public sidecars.
+    """
+    try:
+        _verify(assets, version=version, required_kinds=required_kinds)
+    except (
```
Nobody-Zhang: **`zlib.error` is missing from this tuple and escapes the entire pipeline as a raw traceback**

`zlib.error`, `EOFError`, and `lzma.LZMAError` are not subclasses of `OSError`/`ValueError`/`BadZipFile`. `_verify_consistency` only catches `ReceiptError`, and `Pipeline.run` only catches `ReleaseError`.

Reproduction: the Windows zip's central directory is intact but its deflate stream is corrupted (or the evidence zip's `manifest.json` stream is corrupted) → `_digest(stream)` raises `zlib.error: Error -3 while decompressing data` → the process terminates with a traceback: no `--json` report, no failed `StepResult`, and `_seal_stopped_run` never runs. This still fails closed (nonzero exit), so it is a diagnosability issue rather than a bypass; however, it breaks the contract that "every rejection produces a line in the report", and this read path specifically handles bytes supplied by the draft.

Also, the tuple already extends to `KeyError`/`TypeError`/`AttributeError`/`RuntimeError`, yet a missing key leaves the operator seeing only text such as `'predicate'`. I suggest following `except ReceiptError: raise` with `except Exception` to consistently wrap failures in `ReceiptError`, including the exception type name in the message. (`EvidenceError` is a subclass of `RuntimeError`, and `UnicodeError` is a subclass of `ValueError`, so those two entries in the tuple are already redundant.)

scripts/release_consistency.py
```diff
@@ -0,0 +1,266 @@
...
+        if match is None:
+            continue
+        arch = "aarch64" if match[1] == "arm64" else "x86_64"
+        linux = f"OpenAI4S-{version}-linux-{arch}.tar.gz"
+        if linux not in files:
+            raise ReceiptError(f"{name} payload has no matching release asset {linux}")
+        expected_member = f"{path.stem}/payload/{linux}"
+        with zipfile.ZipFile(path) as archive:
+            members = [
+                info for info in archive.infolist() if info.filename.endswith(".tar.gz")
```
Nobody-Zhang: **The "exactly one payload" check does not match how the Windows launcher reads the archive**

This filters with the case-sensitive `endswith(".tar.gz")`, whereas the actual consumer of this zip, `scripts/windows/openai4s.ps1:243`, uses `Get-ChildItem payload -Filter '*.tar.gz' | Select-Object -First 1`—which is case-insensitive on Windows, with NTFS sorting by name, and trusts the `.sha256` **beside the selected file**.

Reproduction: in addition to the legitimate `<stem>/payload/OpenAI4S-0.2.0-linux-x86_64.tar.gz`, place `<stem>/payload/OpenAI4S-0.0.TAR.GZ` and its sidecar in the zip → this gate returns ACCEPTED; the launcher selects `OpenAI4S-0.0` first (`bootstrap.sh` accepts this directory name), and verification succeeds using the attacker's own sidecar. A "twin" member whose name differs from the legitimate filename only in case, and names with trailing dots, are also accepted. (A name such as `A.TAR.GZ` is rejected by `bootstrap.sh`'s directory-name rule and simply fails to install.)

This requires a repackaged zip with receipts, so the risk is low; nevertheless, this gate endorses a package that "actually installs something else". Key points for a fix: normalize member names using Windows extraction path semantics (treat `\` as a separator, collapse `.`/`..`, strip trailing dots/spaces from each segment, and casefold); allow only the payload and its sidecar under `payload/`, with exactly matching spelling; and retain the rule that "the entire archive must not contain a second tarball" (a `payload/` allowlist alone would miss aliases such as `payload\0.tar.gz`, `./payload/…`, and `wsl/../payload/…`). If a sidecar exists, its digest should equal the Linux asset's digest (compare as-is—`bootstrap.sh` only accepts lowercase hex).

frontend/src/features/onboarding/machine.ts
```diff
@@ -139,13 +139,25 @@ export function reduceWizard(state: WizardState, action: WizardAction): WizardSt
         surface: "wizard",
         complete: false,
       };
-    case "choosePath":
+    case "choosePath": {
+      const path = takePath(action.path);
+      // A receipt belongs to this exact model selection, not its display name
+      // or whether a saved cloud profile is now selected as an existing one.
+      const changed = !state.path ||
```
Nobody-Zhang, line 148: **A receipt can still be attributed to a model that was never probed**

The reducer now asserts that "A receipt belongs to this exact model selection", but `onTest` in `Wizard.tsx` still decides what to probe using `state.path?.profileId || status?.active_id`—the probe target and the model identified by the path can differ:

- **(a) An unsaved local selection**: click a local endpoint (`profileId: ""`) → Checklist → Test → click Test: this actually calls `probeModelProfile("mp-active")`, testing the currently active cloud profile, yet displays its capability badges under the local model and marks "test" as complete, while also wasting a provider request. The checklist's `goto` has no path guard, and the Test button is controlled only by `busy || testing`.
- **(b) Changing the model name after saving**: save local model m1 as P1 → return to the path step and change the model field to m2 without clicking Next (`onChoose({...state.path!, model})` retains `profileId: "P1"`, and the reducer clears the receipt as expected) → Checklist → Test: this probes P1, which is still m1 on the server, so m1's receipt appears beside `path.model = "m2"`.

The impact is limited to presentation (nothing uses the receipt as a gate), but this is precisely the kind of mismatch this PR aims to eliminate. I suggest: when a path exists, `onTest` should use only `state.path.profileId` (show `onboarding.test.needProfile` if it is empty); when no path exists, first `choosePath` for the active profile, then probe it; include the probed `profileId` in `testResult`, and have the reducer discard results that do not match `state.path.profileId`; clear `profileId` when editing the local model name (clicking Next will save a new profile). A→B→A edits made elsewhere remain invisible because `PathChoice` does not carry the `revision` already available on the server—that part requires a product-level decision.

frontend/src/features/onboarding/wizard-skip.test.ts
```diff
@@ -186,6 +202,38 @@ describe("first-run wizard: skipping while a connection test is waiting", () =>
     expect(wizard().step).toBe("readiness");
   });
 
+
+  it.each(["response", "error"])("does not apply the former model's late %s to the newly tested model", async (outcome) => {
```
Nobody-Zhang, line 226: **These two new cases already pass on the base branch, and still pass with the fix removed**

Every exit from the test step (Checklist, Next, Skip, Finish) already calls `leaveTest()`, and `onTest` obtains a new run id every time, so the old probe is invalidated before `choosePath` runs.

Verification: with `Wizard.tsx` + `machine.ts` reverted to `origin/main`, all 5/5 cases in `wizard-skip.test.ts` pass (while the 4 new cases in `machine.test.ts` and 1 in `wizard-integration` fail, showing that those two test files actually detect the regression); removing only `leaveTest();` from `choosePath` on the PR head still leaves wizard-skip and wizard-integration green at 7/7. In other words, `leaveTest()` in the wrapper is redundant and is not exercised by any test; the line added to the README for this test describes behavior that already held.

Keeping this as a characterization test for the "guarantee when switching models" is fine, but it should be labeled accurately. Any genuinely new guarantee that the tests can catch regressions in should live at the reducer layer (for example, have `testResult` carry the profile being probed and discard mismatched results).

scripts/build_linux_bundle.sh
```diff
@@ -376,54 +376,90 @@ Keywords=science;research;agent;python;r;notebook;llm;
...
+    for char in str(Path(app) / "OpenAI4S")
+)
+replacements = {
+    "Exec=@APPDIR@/OpenAI4S": "Exec="
+    + desktop_string('"' + argument.replace("%", "%%") + '"'),
+    "Icon=@ICON@": "Icon=" + desktop_string(icon),
+}
+template = Path(app) / "share/applications/openai4s.desktop.in"
+# Replace whole template lines once: placeholders inside a real path are data.
+rendered = "\n".join(
```
Nobody-Zhang: **The renderer either fails silently or fails destructively**

1. **Any drift in a template line is silently passed through.** `replacements.get(line, line)` matches entire lines exactly, without checking that a replacement actually occurred. The template (lines 365–372) and `uninstall.sh` still expand `$APP_NAME` / `$APP_NAME_LOWER`, while install.sh now uses a quoted heredoc and hardcodes `openai4s`/`OpenAI4S`. If the Exec line changes to something like `Exec=@APPDIR@/OpenAI4S %U`, the renderer exits with code 0 after writing a menu entry that launches the literal `@APPDIR@`, and install.sh still prints `Installed:`; `verify_linux_bundle._check_desktop_entry` uses substring checks, so such a bundle can still pass. The old `sed ... g` replaced every occurrence.
2. **A non-UTF-8 path zeros out an existing menu entry.** `sys.argv` is decoded with surrogateescape, while `write_text(encoding="utf-8")` uses strict encoding: if the bundle or `XDG_DATA_HOME` is in a directory with a legacy-encoded name (such as `/home/j\xfcrgen`), `write_text` first opens the target file with `"w"` (truncating the previously working entry to 0 bytes), then raises `UnicodeEncodeError`. `set -e` then aborts after the CLI link and icons have already been installed. The old `sed` was byte-transparent.

Suggestion: require each of the two keys to occur exactly once; otherwise, call `sys.exit(...)` without writing the file. Use `errors="surrogateescape"` for both reading the template and writing the target.

scripts/build_linux_bundle.sh
```diff
@@ -376,54 +376,90 @@ Keywords=science;research;agent;python;r;notebook;llm;
...
 fi
 
-sed -e "s|@APPDIR@|\$APPDIR|g" -e "s|@ICON@|\$ICON|g" \\
-  "\$APPDIR/share/applications/$APP_NAME_LOWER.desktop.in" \\
-  > "\$DATA_DIR/applications/$APP_NAME_LOWER.desktop"
-chmod 644 "\$DATA_DIR/applications/$APP_NAME_LOWER.desktop"
+# Exec has two escaping layers: the quoted command argument, then the desktop
+# file's string value. Use the bundled interpreter so no host Python or shell
+# substitution syntax participates in rendering a relocated path.
+"$APPDIR/runtime/bin/python3" -I - "$APPDIR" "$ICON" \
```
Nobody-Zhang, line 411: **install.sh now depends on executing the bundled interpreter, but no gate actually runs it**

The release `linux-app` job and the nightly bundle job both run only `build_linux_bundle.sh` + `verify_linux_bundle.py`. The latter only performs static checks on install.sh (it exists, is executable, contains `BASH_SOURCE`, has no `/home` paths, and contains the `@APPDIR@`/`@ICON@` placeholders—and its template checks use substring matching, while the installer requires whole-line matches). It already executes the runtime (`_check_runtime_executes`), but never runs install.sh. The smoke test in step 9 of the build script only runs `openai4s --help`.

The only test in this PR that "actually executes" it symlinks `runtime/bin/python3` to pytest's interpreter, and the app directory is also assembled manually—it tests the renderer text, not the runtime actually shipped, the template actually shipped, or the path after re-extraction. Any regression that occurs only in a real bundle (`-I -` under python-build-standalone, a `.pth` in the science stack printing or failing at startup, or a changed template line) will reach users with every gate green. The old sed installer had no such runtime dependency.

I suggest having `verify_linux_bundle.py` actually run install.sh against the extracted bundle (with temporary `HOME` / `XDG_*`), then parse the rendered `.desktop`. This needs validation against a real bundle on Linux; I cannot verify it on macOS, so I have not made the change myself.

docs/release-validation.md
```diff
@@ -160,6 +160,9 @@ integration. `Exec=` and `Icon=` in a `.desktop` entry are absolute paths, and a
 tarball does not know where it will be unpacked, so the entry ships as a
 template and `install.sh` substitutes the real location at install time.
 Shipping a pre-baked path would produce a menu entry that launches nothing.
+The installer encodes the executable path using the desktop-entry quoting and
```
Nobody-Zhang, line 163: **This sentence promises that paths containing `%` will launch; GLib and KIO actually reject these entries**

Encoding `%` as `%%` follows the specification and does prevent `%F` from being treated as a file argument. But real desktop implementations check whether the executable exists **before expanding field codes**:

- GLib ≥ 2.36 (GNOME and all environments based on `GDesktopAppInfo`): `g_desktop_app_info_load_from_keyfile` calls `g_shell_parse_argv` on the raw `Exec` string, then calls `g_find_program_for_path` on `argv[0]`. If it is not found, it executes `return FALSE`, and the entire entry never appears in the menu. At that point, `argv[0]` is still `/home/u/bundle %%F/OpenAI4S`—a path that does not exist. `g_key_file_get_string` only unescapes `\s \n \t \r \\`; it does not process `%%`.
- KIO (`desktopexecparser.cpp`): it likewise checks `isExecutable()` before `expandMacrosShellQuote`, and reports "Could not find the program" at launch.

This is not a regression (paths containing `%F` were already broken before this PR), and for GLib there is no encoding that lets a literal `%` pass the load-time check. But this text says such paths "must still launch"; `_desktop_launcher_argv` is a hand-written decoder that follows the specification's order and never calls a real launcher, so the `%F %%` parameterized case passes even though GLib will discard the entry. I suggest narrowing this sentence to spaces and shell punctuation, documenting this limitation, making install.sh warn when `$APPDIR` contains `%`, and clarifying in the decoder's docstring that it verifies the specification's order rather than acceptance by a real launcher.

scripts/release_pipeline.py
```diff
@@ -2095,8 +2097,19 @@ def _revalidate_draft_from_checksums(self) -> dict[str, str]:
                         f"digest ({actual[:12]} != {digest[:12]}); refusing to "
                         f"publish"
                     )
+            self._verify_consistency([Path(temp) / name for name in expected])
```
Nobody-Zhang: **finalize hashes every downloaded asset three times**

The loop above has just computed `checked[name] = sha256_file(local)` for each file. A few lines later, the gate called here hashes every file again (`release_consistency.py:164-166`), and `verify_build_receipts` hashes every distribution with a receipt a third time (`release_receipts.py:368`)—even though `_match(coverage, distributions, ...)` already established the result of that comparison four lines earlier.

Before this PR there was 1 pass; afterward there are 3, plus extracting and reading the Linux tarball from every Windows zip. According to `build_windows_zip.sh:281`, the Linux payload is about 380 MB: for a draft containing a Linux tarball, a Windows zip embedding it, and a DMG, roughly 2 GB is read three times. `step_upload` also adds another 2 passes to a run that already hashes every distribution 9 times.

Suggestion: add `verify_release_consistency(..., digests=None)` and have finalize pass `digests=checked` (hashes just computed in the same function for the same temporary files, weakening no guarantees); let `verify_build_receipts` accept `digests=` as well, so it no longer rereads bytes that `_match` has already compared. Also: the evidence zip's digest is computed but never used; it is written to a temporary directory only because `verify_build_receipts` accepts paths only.

PR #171, opened 2026-09-17
fix(contributors): keep bilingual avatar inventories in sync

CyrusAuyeung:
## Summary

Refreshing contributor avatars can add PNG files without updating `.github/contributors/README.md` and `README_zh.md`, causing the directory-documentation gate to fail. The manual inventory-update step was also noted in [#152](https://github.com/PKU-YuanGroup/OpenAI4S/pull/152).

The generator now synchronizes both marked inventory tables from the surviving PNG filenames returned by the avatar refresh. A successful refresh preserves the existing filename spelling when login casing changes, avoiding case-variant duplicates on case-sensitive filesystems. Failed downloads retain existing current-contributor avatars.

Both inventory documents are read and their markers validated before authentication, fetching or writing. Missing or malformed inventories stop the run with a diagnostic and exit code 1. All four READMEs use the same block renderer and writer: every changed document is prepared in a sibling temporary file before any destination is replaced, preserving file permissions and surrounding prose.

Regression fixtures track existing avatars in Git and stage the refresh before running the real directory gate. The bilingual documentation explains that staging is required for the gate to stop listing deleted tracked avatars. The generator itself does not modify the Git index.

## Area and Change Type

- Tests / harness / CI; Docs
- Bug fix; Test / harness; Documentation

## Verification

```text
uv run --no-project --with pytest python -m pytest --noconftest tests/test_update_contributors.py -o addopts= -q
# 28 passed on ordinary Windows NTFS
uvx black --check scripts/update_contributors.py tests/test_update_contributors.py
uvx isort --check-only scripts/update_contributors.py tests/test_update_contributors.py
uvx ruff check scripts/update_contributors.py tests/test_update_contributors.py
git diff --check
```

- The same 28 tests also passed with pytest's temporary directory under a fresh NTFS directory with per-directory case sensitivity enabled. Distinct differently cased filenames were verified both in the parent directory and inside the successful-refresh fixture.
- The new `OSError` and `KeyboardInterrupt` regressions fail on `65e2063` because both root READMEs have already changed when the last inventory staging write fails. They pass with this change: all four originals remain byte-identical and temporary files are removed.
- Preflight tests cover missing, missing-end, reversed and duplicate markers in either language and assert that authentication/fetching is never reached. Refresh tests cover additions, tracked removals, failed downloads, successful and failed casing changes, exact on-disk filenames and repeat-run stability.
- Network and image conversion are stubbed; these are script tests using the real generator, Git enumeration and directory checker.
- After merging `main` at `a6955de` into the PR branch, the 28 contributor tests and the three release-documentation tests changed by the base update passed together: **31 passed**. `git diff --check origin/main...HEAD` also passed on the merged head (`4772ddb`).

Not run locally: the full offline suite, mypy, full-repository directory checks or `pre-commit run --all-files`. Local validation used a Windows sparse checkout and the focused tests bypass the project-wide `conftest.py`. Linux/macOS execution, live GitHub requests and Pillow conversion were not exercised. The CI and Protocol fuzzing workflows for the merged head currently report `action_required`; upstream execution results are not yet available.

## Risk and Reviewer Focus

Review the preflight ordering, preservation of existing avatar spelling and the single preparation batch for all four READMEs. The existing directory-documentation requirements remain in place.

A staging failure leaves every README unchanged. Each `os.replace` is atomic, but replacing multiple paths is not an all-or-nothing transaction: an interruption during the replacement loop can leave a mixture of old and new documents. Avatars are refreshed before document staging, so this is not a transactional rollback of the entire refresh. These limits are documented rather than described as whole-tree atomicity.

## Checklist

- [x] Reviewed the complete diff; changes are limited to this fix.
- [x] Added offline regression coverage without deleting existing tests.
- [x] Kept the affected English and Chinese documentation in sync.
- [x] Added no dependencies, credentials or private data.


scripts/update_contributors.py
```diff
@@ -243,6 +291,7 @@ def main() -> int:
     have_png, written = write_avatars(people, token)
     block = render(people, have_png)
     changed = [p for p in READMES if os.path.exists(p) and update_readme(p, block)]
+    changed.extend(update_avatar_readmes())
```
Nobody-Zhang: **The marker validation runs after the PNGs and both root READMEs have already been rewritten, so a malformed inventory leaves exactly the half-updated tree this PR is meant to prevent.**

`main()` calls `write_avatars()` (writes new PNGs, prunes departed ones) and `update_readme()` (rewrites the wall in `README.md`/`README_zh.md`) before `update_avatar_readmes()` gets to check the `AVATAR-FILES` markers. I reproduced this with this PR's fixture: `README_zh.md` has no markers, the API returns `NewPerson`, and the checkout starts with `Departed.png`:

- `main()` raises `ValueError` with a traceback
- `Departed.png` has already been deleted and `NewPerson.png` written
- the root wall already links `.github/contributors/NewPerson.png`
- both avatar READMEs are untouched, so `check_directory_readmes.py` exits 1 (`bilingual structure mismatch` + ``does not mention `NewPerson.png` ``)

"Both marker blocks are validated before either inventory document is written" holds for 2 of the 5 outputs. The new marker test calls `update_avatar_readmes()` directly rather than `main()`, so it can't see this.

Suggest reading and validating both inventory documents at the top of `main()`, before `fetch_contributors()`, and returning 1 with a message (like the empty-API path) instead of raising. A `main()`-level test asserting that no PNG or root README changed would pin it.

scripts/update_contributors.py
```diff
@@ -233,6 +236,51 @@ def update_readme(path: str, block: str) -> bool:
     return True
 
 
+def update_avatar_readmes() -> list[str]:
+    """Document surviving PNGs, preserving their on-disk filename spelling."""
+    names = sorted(
```
Nobody-Zhang: **On a case-sensitive filesystem, a login whose casing drifted leaves two avatars for one person, and the inventory now documents the stale one as a separate contributor.**

I ran this on a case-sensitive APFS volume as a stand-in for the Linux runner. Setup: `existing.png` is committed, the API returns `Existing`, and the download succeeds:

```
on disk: ['Existing.png', 'README.md', 'README_zh.md', 'existing.png']
| `Existing.png` | Render-ready avatar for contributor `Existing`. |
| `existing.png` | Render-ready avatar for contributor `existing`. |
gate: 0
```

On that filesystem, `open("Existing.png", "wb")` creates a new file, and the prune in `write_avatars` keeps `existing.png` because it compares with `casefold()`. This function then lists every surviving file and takes the "contributor" name from the filename (`name[:-4]`). The duplicate passes the gate, gets committed, and is never pruned while the login is current.

A tree holding both `Existing.png` and `existing.png` also collides when checked out on macOS or Windows. On macOS the same refresh leaves only `existing.png`, so the committed inventory depends on which platform ran it.

The root cause is upstream of this function: the refresh never moves the old file to the current login casing. For example, on a successful download, remove any case-variant of `_avatar_name(login)` and then write the new file via a temp name + `os.replace`, so the new spelling sticks on case-insensitive filesystems too. The inventory then has nothing stale to document.

tests/test_update_contributors.py
```diff
@@ -150,3 +155,122 @@ def record(url, token):
...
+            ["Unavailable"],
+            ["Existing.png"],
+        ),
+    ],
+)
+def test_refresh_keeps_bilingual_inventory_and_gate_in_sync(
+    contributor_checkout, monkeypatch, existing, logins, failed, expected
+):
+    avatar_dir = contributor_checkout
+    for name in existing:
```
Nobody-Zhang, line 304: **The fixture's "real Git file inventory" never tracks the existing avatars, so the removal scenario passes only because nothing is committed.**

`check_directory_readmes._source_files()` uses `git ls-files --cached --others`. In the real repository every avatar is committed, and the `os.remove` in `write_avatars` doesn't touch the index. I reran the `Departed.png` scenario with a `git add -A` before `main()`, and the gate then reports:

```
directory docs: .github/contributors/README.md does not mention `Departed.png`
directory docs: .github/contributors/README_zh.md does not mention `Departed.png`
```

It stays red until someone runs `git rm` / `git add -A`. So "refresh, then run the gate" still fails whenever a contributor departs. The test's description says it runs the real gate against a Git inventory, but this is the state the real repository is actually in, and the test doesn't cover it.

Suggest committing (or at least `git add`ing) the `existing` files in the fixture. Then either stage the refresh in the test before asserting the gate, or document that the gate has to be run after staging.

tests/test_update_contributors.py
```diff
@@ -150,3 +155,122 @@ def record(url, token):
...
+        ([], ["NewPerson"], [], ["NewPerson.png"]),
+        (["Departed.png", "Departed.svg"], ["NewPerson"], [], ["NewPerson.png"]),
+        (
+            ["Existing.png"],
+            ["NewPerson", "Existing"],
+            ["Existing"],
+            ["Existing.png", "NewPerson.png"],
+        ),
+        (
+            ["existing.png"],
```
Nobody-Zhang, line 280: **The casing case only exercises the failed-download branch. The successful-download branch behaves differently per platform and has no test.**

With `existing=["existing.png"], logins=["Existing"], failed=[]`, macOS produces `["existing.png"]` and a case-sensitive filesystem produces `["Existing.png", "existing.png"]` (see the comment on `update_avatar_readmes`). No single `expected` value passes on both, which means the contract for that case was never defined. The PR description lists "filename casing" as covered, but the branch that actually produces a duplicate is untested. Worth adding once the intended behaviour is settled.

scripts/update_contributors.py
```diff
@@ -233,6 +236,51 @@ def update_readme(path: str, block: str) -> bool:
...
+        rows = [header, "| --- | --- |"]
+        for name in names:
+            detail = description.format(login=name[:-4])
+            rows.append(f"| `{name}` | {detail} |")
+        start = text.index(INVENTORY_START) + len(INVENTORY_START)
+        end = text.index(INVENTORY_END)
+        updated = text[:start] + "\n" + "\n".join(rows) + "\n" + text[end:]
+        if updated != text:
+            updates.append((path, updated))
+    # Validate both blocks before rewriting either document.
```
Nobody-Zhang: **Validating both blocks first protects against bad markers, but the two writes that follow are still not atomic.**

Each `open(path, "w")` truncates before it writes. If the second write is interrupted (ENOSPC, Ctrl-C, a killed automation run), `README.md` has the new inventory and `README_zh.md` is stale or truncated. The gate then reports a structure mismatch, and a truncated Chinese README loses its prose.

Writing each document to a sibling temp file and `os.replace`-ing them only after both writes succeed keeps the pair consistent. The root-README writer has the same pattern, so one shared helper would cover both.

scripts/update_contributors.py
```diff
@@ -233,6 +236,51 @@ def update_readme(path: str, block: str) -> bool:
...
+            "Render-ready avatar for contributor `{login}`.",
+        ),
+        (
+            "README_zh.md",
+            "| 文件 | 职责 |",
+            "贡献者 `{login}` 的可直接渲染头像。",
+        ),
+    ):
+        path = os.path.join(AVATAR_DIR, filename)
+        with open(path, encoding="utf-8") as fh:
```
Nobody-Zhang, line 325: **A missing avatar README crashes with `FileNotFoundError`, after the PNGs and the wall have already been written.**

`write_avatars()` calls `os.makedirs(AVATAR_DIR, exist_ok=True)`, so the script expects the directory might not exist yet, and `main()` guards the root READMEs with `os.path.exists`. This function opens both avatar READMEs unconditionally. On a checkout without them (a fresh directory, or `GITHUB_REPOSITORY` pointed at another repo), the run ends in a traceback and leaves the same half-updated tree as above.

The failure policies also differ: missing `CONTRIBUTORS` markers print a warning and are skipped, while missing `AVATAR-FILES` markers raise. Whichever policy you choose, check for the files up front together with the marker validation.

scripts/update_contributors.py
```diff
@@ -233,6 +236,51 @@ def update_readme(path: str, block: str) -> bool:
...
+            text.count(INVENTORY_START) != 1
+            or text.count(INVENTORY_END) != 1
+            or text.index(INVENTORY_START) > text.index(INVENTORY_END)
+        ):
+            raise ValueError(f"invalid avatar inventory markers in {path}")
+        rows = [header, "| --- | --- |"]
+        for name in names:
+            detail = description.format(login=name[:-4])
+            rows.append(f"| `{name}` | {detail} |")
+        start = text.index(INVENTORY_START) + len(INVENTORY_START)
```
Nobody-Zhang: **This reimplements the marker-block replacement that `update_readme()` already does, and uses a different failure policy.**

`update_readme()` splices the text between `START`/`END` with `re.sub` and returns `False` (after printing) when the markers are missing. This function does the same splice by slicing with `index()`, and raises instead. Having two splice implementations with two error semantics in one 300-line script means they will drift apart the next time either one changes (newline handling, marker validation, atomic writes).

Suggest generalizing `update_readme(path, block, start=START, end=END)` into a pure "render the updated text or raise" helper plus one shared writer, and using it for all four documents.

.github/contributors/README.md
```diff
@@ -5,11 +5,13 @@
...
 writes them from the GitHub contributors API and a maintained list of publicly
-recognized non-commit contributors. Nothing in the OpenAI4S runtime reads this
-directory.
+recognized non-commit contributors. The same script updates the marked file
+tables in both directory READMEs from the PNGs left on disk after refresh and
+pruning. Nothing in the OpenAI4S runtime reads this directory.
 
 ## Files
 
+<!-- AVATAR-FILES:START -->
```
Nobody-Zhang, line 21: **Consider fixing this at the gate instead of with generator-maintained markers.**

`EXCLUDED_PREFIXES` in `check_directory_readmes.py` exists for trees that are "third-party, generated, or byte-exact fixtures", and the parent README describes the boundary. `.github/contributors/` is a generated tree: one script writes and prunes every PNG in it.

The marker approach needs two header literals, two row templates and marker validation to stay in sync with these READMEs. It also brings the partial-write, stale-index and platform-dependent-inventory problems raised on this PR. Excluding the directory, or teaching the gate that a directory's `*.png` files are owned by a generator, removes that whole class of failure.

The cost is losing per-file documentation, and that's a maintainer call, so this is an option rather than a required change.

scripts/update_contributors.py
```diff
@@ -233,6 +236,51 @@ def update_readme(path: str, block: str) -> bool:
     return True
 
 
+def update_avatar_readmes() -> list[str]:
+    """Document surviving PNGs, preserving their on-disk filename spelling."""
+    names = sorted(
+        name
```
Nobody-Zhang: **This lists the directory again, but `write_avatars()` has already computed the surviving set.**

After pruning, `write_avatars()` builds `on_disk = set(os.listdir(AVATAR_DIR))`, and the wall is derived from it. The inventory works out the same "which PNGs survived" fact separately, with its own filter (`endswith(".png")` + `isfile`). If either filter changes later (extension case, symlinks), the wall and the inventory can disagree about which files exist. Returning the surviving PNG names from `write_avatars()` and passing them in keeps both on one source. Minor: `text.index(INVENTORY_START/END)` is also recomputed after validation.

tests/README_zh.md
```diff
@@ -427,7 +427,7 @@ OpenAI4S 的离线正确性门禁。`uv run pytest` 用确定性 fake 跑完这
...
-| [`test_update_contributors.py`](test_update_contributors.py) | Community Contributors 生成器会把已公开署名的非 commit 贡献者保留在按提交数排序的 API 用户之后，按登录名大小写无关地去重，并拒绝同时出现在排除名单里的署名登录名；API 返回空时仍会失败即拒绝，不会把贡献者墙改写成只剩维护名单。头像临时刷新失败时会保留该现有贡献者已提交的 PNG，同时继续清理已离开的身份，并如实报告本次刷新了 0 张而不是给出一个看起来正常的数字。断言直接写出署名登录名字面量——从被测常量反推出来的期望，在该常量为空时同样会通过。另外钉住两处链接完整性回归：登录名大小写与已提交文件名不一致时，文件会被保留、并回退到远端头像，而不是为一个刚被清理掉的文件写出本地 `<img src>`；以及未鉴权的 `github.com/<login>.png` 回退路径不会把仓库 token 带过那次跨主机跳转。 |
+| [`test_update_contributors.py`](test_update_contributors.py) | Community Contributors 生成器会把已公开署名的非 commit 贡献者保留在按提交数排序的 API 用户之后，按登录名大小写无关地去重，并拒绝同时出现在排除名单里的署名登录名；API 返回空时仍会失败即拒绝，不会把贡献者墙改写成只剩维护名单。头像临时刷新失败时会保留该现有贡献者已提交的 PNG，同时继续清理已离开的身份，并如实报告本次刷新了 0 张而不是给出一个看起来正常的数字。断言直接写出署名登录名字面量——从被测常量反推出来的期望，在该常量为空时同样会通过。另外钉住两处链接完整性回归：登录名大小写与已提交文件名不一致时，文件会被保留、并回退到远端头像，而不是为一个刚被清理掉的文件写出本地 `<img src>`；以及未鉴权的 `github.com/<login>.png` 回退路径不会把仓库 token 带过那次跨主机跳转。 新增的离线刷新场景在临时 Git 仓库中运行真实目录 README 门禁，检查新增、移除、下载失败及文件名大小写变化后的双语头像清单，并验证重复运行结果稳定、非法清单标记会被拒绝。 |
```
Nobody-Zhang: **Nit: stray half-width space after `。`.**

The new text reads `…跨主机跳转。 新增的离线刷新场景…`. No other row in this table puts a space after a full-width period.

```suggestion
| [`test_update_contributors.py`](test_update_contributors.py) | Community Contributors 生成器会把已公开署名的非 commit 贡献者保留在按提交数排序的 API 用户之后，按登录名大小写无关地去重，并拒绝同时出现在排除名单里的署名登录名；API 返回空时仍会失败即拒绝，不会把贡献者墙改写成只剩维护名单。头像临时刷新失败时会保留该现有贡献者已提交的 PNG，同时继续清理已离开的身份，并如实报告本次刷新了 0 张而不是给出一个看起来正常的数字。断言直接写出署名登录名字面量——从被测常量反推出来的期望，在该常量为空时同样会通过。另外钉住两处链接完整性回归：登录名大小写与已提交文件名不一致时，文件会被保留、并回退到远端头像，而不是为一个刚被清理掉的文件写出本地 `<img src>`；以及未鉴权的 `github.com/<login>.png` 回退路径不会把仓库 token 带过那次跨主机跳转。新增的离线刷新场景在临时 Git 仓库中运行真实目录 README 门禁，检查新增、移除、下载失败及文件名大小写变化后的双语头像清单，并验证重复运行结果稳定、非法清单标记会被拒绝。 |
```

Nobody-Zhang:
**Replayed my fixes on top of `ff8ca010` instead of merging them.**

Your batched staging is the stronger structure, so I reset onto it and re-applied my changes by hand. `main()` still makes exactly one `_write_texts` call with all four documents, and `test_main_stages_all_four_readmes_before_replacing_any` passes unchanged.

On top of that, `f049081b`:

- **`read_documents()` validates the two root walls as well as the inventories**, before anything is fetched or written. A root README carrying a second `CONTRIBUTORS` pair was skipped by `render_readme_update` while its sibling updated — two walls disagreeing about who contributed, with the run still exiting 0. Reproduced before the change: `README.md` untouched, `README_zh.md` updated, exit code 0.
- **`render()` links each avatar under the name that survived the prune.** With `existing.png` committed and the API returning `Existing`, the refresh rewrote that file while the wall still pointed at `https://github.com/Existing.png` (`0 linked locally`), so the refreshed file was never referenced by anything. On a case-sensitive APFS volume it now reports `1 linked locally` and links `.github/contributors/existing.png`.
- **Staged writes carry a known prefix, and both the next write and the prune sweep what a killed run left.** A leftover temp file is not an image, so the prune ignored it and `check_directory_readmes.py` stayed red on a file nobody documented. The prune needs it too: a run where no document changes never reaches the writer.
- **`_target_name()` makes the committed-spelling choice testable without a case-sensitive filesystem.** The parametrized casing case passes on macOS either way, so it cannot catch a regression there; the unit assertion can.

**One finding I dropped on purpose.** I had made a *missing* inventory document non-fatal: refresh the avatars and the wall, report the missing file, exit non-zero. Your tests pin the opposite — a missing or malformed inventory in either language stops the run before authentication, network access or writes — and that is what is on the branch now. The bilingual pair is already incomplete when a document is missing, so refusing does not save the directory gate either way, but failing closed is the simpler contract. Noting it only so the alternative is on the record.

Docs: the staging instruction is now `git add -A`, since a refresh rewrites the root walls too, not only the avatar directory.

Verification: `tests/test_update_contributors.py` 33 passed (including your four staging tests), the full offline suite green, `check_directory_readmes.py` green, regenerating the inventories in this checkout is a no-op, and each of the four behaviours above was mutation-checked — revert the fix and the matching test fails.

CI has still not run on any of these commits: the fork's workflow runs sit at `action_required` and need a maintainer to approve them.

PR #175, opened 2026-09-19
fix: recover interrupted streams and repeated-action stops

### LINKED ISSUE #174: [Bug] 长流式生成中途上游静默 ≥600s 触发 TransportError 后回合不可恢复；重复动作熔断（no_progress）后点击“继续”无出口
### Area

Server / gateway / API

### Current behavior

您好！我们在自部署的 OpenAI4S 上遇到了两类会话中断问题。经过只读排查（未改动源码与部署），已定位证据链，整理如下，希望能帮助改进产品。感谢开源这份优秀项目。

### 现象 A：长流式生成中途，上游长时间无输出（≥600s），回合被读超时掐断且无法恢复

证据：

- 数据库 `messages` 表中，该 frame 的失败回合时间线：末条正常消息 → 错误消息间隔为 614–739 秒，恰为 600s 超时 + 开销，可反推配置确实生效；若只有约 380s 间隔则说明是 120s×3 的重试路径，本例不是。
- 日志中的完整 Traceback（已脱敏，不含 key/token）：

  ```
  Traceback (most recent call last):
    File ".../transport.py", line 441, in __next__
      raise TransportError(f"event stream read error: {e}") from e
  openai4s...TransportError: event stream read error
  ```

- 源码 `transport.py:438–447` 的重试判据：`retryable = not committed`。即只要流已吐出过事件（committed），就按设计禁止重放（避免重复输出/工具副作用），于是这个回合直接死亡，用户侧表现为一次不可恢复的"无法连接到 LLM 服务"。
- 上游侧实证（排除本地因素）：
  - 对 `z***.cat` 与 `m***.l***.xyz` 分别从 Windows 与 WSL 两端做 TCP+TLS+HEAD 探测，全部 0.1–0.7 秒成功（HTTP 200）。即连接始终健康，是上游在长生成中途"静默卡死"。
  - 对照实验：通过本地网关 REST API 向同一服务、同一上游发送一个极短请求（提示词为"只回复两个字符 OK，不使用任何工具"），HTTP 200，assistant 正常回复"OK"，短请求完全正常，无任何 TransportError。（顺带观察到：这个两字回复耗时约 60 秒，说明该中转整体响应也相当慢。）

### 现象 B：模型连续输出相同内容触发 same_action_streak 熔断，用户点击"继续"无法恢复

证据：

- 日志中可见 `no_progress` 相关错误与 `same_action_streak` 计数（gateway `gateway.py:10171` 附近的投影逻辑）。
- `host_call_log` 中该回合 252 次工具调用全部成功（ok=1），说明熔断并非工具失败所致，纯粹是重复内容计数。
- 对比连续两次失败回合中同位置工具调用的 `args_preview`，逐字节相同。

### 实际行为 vs 期望行为

实际行为：

- A：committed 后的流式读超时 → 报"无法连接到 LLM 服务"，回合不可恢复（设计上为避免重复输出而禁重试）。
- B：熔断后仅提示"编辑提示或显式继续"，而"显式继续"实际会把同一动作重放并再次熔断，没有真正的恢复出口。

### 环境与缓解（供其他用户参考）

- 临时缓解：将 `OPENAI4S_LLM_TIMEOUT` 调低（如 120–180s）可更早暴露问题、减少无效等待（committed 后反正不可重试，早超时早止损）；或将长任务拆到更短上下文的会话。
- 根因在第三方中转/上游在超长上下文长生成中的静默行为，本地网络/超时配置已排除。

再次感谢！

### Expected behavior

期望行为（建议，供参考）：

- A：committed 状态下的流式静默超时，宜以"可恢复的停止 + 上游静默超时"明确文案呈现，而不是连接错误样式的报错；可考虑比总超时更短的 idle-chunk 看门狗，或在 committed 后提供透明的回合续跑（从 checkpoint 重发该模型调用，而不重复已落盘的工具结果）。这与 open PR #160（fix: bound model calls and preserve workbench state）中"给模型调用加预算/总时限"的方向互补。
- B：熔断时注入一条 system 提示，点名被重复的动作并强制模型换一个替代方案（或自动带该提示续跑一次），而不是留下死胡同。

### Steps to reproduce

现象 A：

复现步骤（可操作）：

1. 在包含超长上下文与密集工具链的会话中发起一个会产生长输出的任务（如批量生成多份文档、连续多步工具调用）。
2. 生成进行到中途（此前已有事件流出，即流已 committed），上游停止吐出任何 chunk。
3. 等待约 600 秒（`OPENAI4S_LLM_TIMEOUT=600`）后，界面出现"无法连接到 LLM 服务"错误气泡，回合终止。

现象 B：

复现步骤：

1. 给模型一个它会陷入循环的任务（我们的一次实际案例：工作区内 pip 无法联网安装 `python-docx`，模型改为手工拼 zip 生成 docx，成功后开始反复"自写 kernel-runner 脚本"）。
2. 模型连续多个回合输出逐字节相同的文本与相同参数的工具调用，直至 `same_action_streak` 达到阈值熔断。
3. 界面出现"已停止重复动作。请编辑提示或显式继续"。
4. 用户点击"继续"→ 模型重放同一动作 → 再次立即熔断。反复点击均如此，会话表现为"无法继续"。

### Version or commit

OpenAI4S 0.3.0（`OpenAI4S-0.3.0-linux-x86_64-1ba855e...acf356`，即 v0.3.0 release，2026-09-17 发布）

### Environment

- OS：Windows 11 宿主 + WSL2 Ubuntu 24.04（服务部署在 WSL 内，`127.0.0.1:8760`）
- Python：应用自带 runtime（bin/python3）
- 关键配置：`OPENAI4S_LLM_TIMEOUT=600`（应用根 `.env`，已确认生效，详见下文时间线推算）
- 模型路由：第三方中转 `z***.cat/v1` 与 `m***.l***.xyz/v1`（模型 `glm-5.3-flash`）

### Logs or screenshots

```text

```

### Checks already tried

_No response_

### Severity

P1 - important bug

### Public disclosure check

- [x] I have removed secrets, private data, local absolute paths, and unreleased research details.

Nobody-Zhang:
## Summary

Fixes #174.

Reported and diagnosed by @Pandasama2025 in #174. Their reproduction steps and analysis informed this fix. This PR credits them as a co-author and adds them to the maintained contributor list and both README avatar walls.

After an upstream stream stalled after its first event, the turn was reported as a connection failure and lacked recovery context. A `no_progress` stop also left the next turn without an explanation of the repeated action.

This change classifies interrupted streams, preserves received public prose, and provides an explicit **Continue** action. Continuation receives the same bounded recovery guidance from live history or the Action Ledger after reopening the session. The transport still refuses automatic replay after committed output; incomplete code and tool arguments are never executed.

The work is split into four reviewable commits:

1. `c8d9ae6f` — classify stream timeouts/interruption and verify the committed-output retry veto.
2. `c0faf4e5` — preserve partial prose and reconstruct stopped-turn recovery context, with backend regression coverage.
3. `1bf4c79b` — add the guarded Continue action for live/restored failures, frontend tests, bilingual inventories, and rebuilt dist.
4. `132d8616` — credit @Pandasama2025 with a co-author trailer, maintained recognition entry, and bilingual contributor walls/inventories.

## Area

- [x] Web UI / frontend
- [x] Server / gateway / API
- [x] Tests / harness / CI
- [x] Docs

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Test / harness
- [x] Documentation
- [ ] Security-related change
- [ ] Release / packaging

## Verification

**This is a draft: the local full suite is not green.** Full-suite results were **9,232 passed, 138 skipped, 19 failed**. Re-running those 19 cases against the unchanged base reproduced 18 failures. The remaining artifact snapshot test passed individually on both the base and this change. All newly added regression tests passed.

The baseline failures include CRLF-dependent script/hash checks, missing `uv` in WSL, interpreter-path assertions, an existing task-mode assertion, and two HTTP timeouts. The response-schema check started the full suite and then stopped as designed when pytest failed; it did **not** produce a successful schema-verification result.

Other results:

- Contributor credit follow-up: `python -m pytest tests/test_update_contributors.py` — **33 passed**; all-file pre-commit and bilingual directory coverage passed.
- Focused kernel/agent/gateway, progress, transport and ledger coverage: **346 passed, 2 failed**; both failures reproduced on the unchanged base.
- Frontend: **88 files / 798 tests passed**. A temporary Vitest setup selected English for existing locale-dependent assertions; LF inputs were used for frozen extraction checks.
- Offline harness: **38 scenarios passed** (26 contract, 12 production paths).
- Response route contract: **212/212**.
- Both frontend TypeScript configurations, Vite build, mypy, all-file pre-commit, bilingual directory coverage, source secret scan, wheel/sdist build and artifact verification passed.

Ran (machine-specific interpreter prefixes omitted; Python tests ran in the configured locked environment / WSL equivalent):

```text
python scripts/capture_response_schemas.py --check
python scripts/capture_response_contract.py --check
python -m harness.cli run --tier pr --offline
python -m mypy
python -m pre_commit run --all-files
python scripts/check_directory_readmes.py
node frontend/node_modules/vitest/vitest.mjs run --config .build/vitest.config.mts
# From frontend/:
node node_modules/typescript/bin/tsc --noEmit -p tsconfig.json
node node_modules/typescript/bin/tsc --noEmit -p tsconfig.node.json
node node_modules/vite/bin/vite.js build
# From the repository root:
uv build --out-dir .build/issue174-dist
python scripts/verify_release_artifacts.py .build/issue174-dist
git diff --check
```

Manual browser verification used live Ark `doubao-seed-2.0-lite` responses with controlled local fault injection:

- After the first real SSE content event, inject a read timeout. Verify the partial prose and Continue action survive refresh, then continue with the real provider to completion.
- Repeat one real `list_skills({})` reply three times to trigger `no_progress`. Refresh and continue. The real provider switches to distinct collection offsets `0/150/300/450` and completes through `finalize_response`.
- A natural 600-second provider stall was **not** observed; the timeout path was exercised by injection. Credentials were supplied only through the temporary process environment.

Not run:

```text
python -m harness.smoke.linux_bwrap_interrupt
python -m harness.smoke.linux_sandbox
node tests/browser_smoke.mjs
```

Reason:

```text
Local verification used Windows/WSL and a real browser against an isolated daemon.
The full Docker/bubblewrap and cross-engine browser matrix were not exercised locally.
The focused browser flows above cover the changed stopped-turn behavior.
```

## Risk and Reviewer Focus

- `openai4s/llm/transport.py` and `models.py`: typed stream classification must preserve the existing committed-output retry veto.
- `openai4s/agent/recovery.py`, `engine.py`, and `ledger.py`: live and reconstructed recovery guidance must match, without replaying actions or copying raw arguments/provider errors into the system note.
- `openai4s/server/agent_run.py` and the surgical `gateway.py` changes: retain public prose once while discarding incomplete code drafts and preserving terminal-event ordering.
- `frontend/src/features/messages/failure.ts` and `send/turn.ts`: continuation is explicit and guards the active frame, running state, existing composer text, and stale failure rows.
- Recovery guidance influences the provider's next choice; it does not guarantee the provider will find a successful alternative. The no-progress circuit remains active.

### Diff Readiness

      `host_dispatch.py`, `store.py`, `worker.py`, `manager.py`.
      source. `OPENAI4S_WEBUI=legacy` is an escape hatch, not a place to add
      features.
      reformatted.

### Tests

      a documented smoke test.

### Core Dependency Policy

      stdlib web server.
      in-tree use site. No science imports were added.
- [x] New dependencies, if any, are documented and justified. None were added.

      verification notes.
      guarantees.

### Docs

      appropriate.
      affected.

---

> [!NOTE]
> **Medium Risk**
> Touches LLM transport retry semantics, gateway turn failure handling, and model context via recovery notes; UI continuation sends a fixed prompt through global `send`.
> 
> **Overview**
> Fixes interrupted LLM streams and **no_progress** stops so users can recover without losing partial work or context.
> 
> **Backend:** Mid-stream SSE read failures are classified as `llm_stream_timeout` or `llm_stream_interrupted` (`StreamTimeoutError` / `StreamReadError`) while keeping the existing **`output_committed` retry veto**. New `recovery.py` emits a bounded `[Stopped turn recovery]` system note for stream interruption and no-progress stops; the engine, Action Ledger reconstruction, and gateway append the same guidance on live failure and session reopen. Partial public prose is retained via `finish_interrupted()` when a turn dies mid-stream; incomplete code drafts are cleared.
> 
> **Frontend:** Shared `failure.ts` drives failure hints and a guarded **Continue** button for those three failure codes (active frame, empty composer, latest failure row only). `turnDone` attaches failure UI on failed live turns; stored history uses the same module. User-facing copy and tests cover live + restored paths.
> 
> Dist assets were rebuilt from `frontend/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bf4c79be56470a05f912f500863723abb96d627. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openai4s/agent/engine.py
```diff
@@ -188,6 +189,17 @@ def _finish(
...
+            note = recovery_message(
+                stop_reason,
+                progress_reason=progress_reason,
+                tool_names=(
+                    (call.name for call in state.last_reply.tool_calls)
+                    if state.last_reply is not None
+                    else ()
+                ),
+            )
+            if note is not None and (not state.messages or state.messages[-1] != note):
```
Nobody-Zhang, line 201: **[correctness · high] The recovery note lands between `tool_use` and `tool_result` on the Anthropic and Gemini wires.**

The note is appended as a `system` message directly after the tool results. `_anthropic_messages` / `_gemini_contents` (`openai4s/llm/messages.py`) emit a mid-timeline system message *before* flushing `pending_results`, so the continuation request becomes:

```
assistant  [tool_use list_dir]
user       "[system] [Stopped turn recovery] …"
user       [tool_result]
user       "Continue…"
```

Anthropic rejects that with a 400 (a `tool_use` must be followed immediately by its `tool_result`), and since `ledger.py` rebuilds the same note at the same position, every later turn of the session fails too. I reproduced the ordering by feeding this PR's history shape to `_anthropic_messages`. The OpenAI chat wire is unaffected, which is why the Ark verification passed.

Root-cause fix is in the wire assembly: hold a mid-timeline system message until the pending tool results are flushed. A wire-level regression test for both assemblers should go with it.

openai4s/llm/transport.py
```diff
@@ -438,7 +440,10 @@ def dispatch() -> None:
         except Exception as e:  # noqa: BLE001 - normalize transport read failures
             # A mid-stream read failure after events were delivered must not be
             # replayed: the caller already saw partial output.
-            raise TransportError(
+            error_type = (
```
Nobody-Zhang, line 460: **[correctness · medium] Local `on_event` exceptions are now reported as an upstream stream interruption.**

This `except Exception` also covers `dispatch()` → `on_event(event)`. A `KeyError` in a provider's delta assembler, or a sink/persistence failure, becomes `StreamReadError` → `llm_stream_interrupted`: the user is told the upstream dropped the stream, gets **Continue**, and the history gains a note saying the stream was cut. Continue then reproduces the local bug every time.

Suggest classifying only real read failures (`OSError`, which includes `TimeoutError`, and `http.client.HTTPException`) as `StreamReadError`, and leaving everything else a plain `TransportError` as before.

openai4s/agent/recovery.py
```diff
@@ -0,0 +1,56 @@
...
+            "The previous model response was interrupted before it completed. "
+            "Its partial tool arguments and code were not executed. "
+            "Earlier completed tool calls and cells remain recorded in this history. "
+            "On continuation, inspect those results and continue only the unfinished work. "
+            "Do not replay completed actions, resubmit jobs, or recreate existing artifacts "
+            "merely because the model stream was interrupted."
+        )
+    else:
+        return None
+    return {"role": "system", "content": "[Stopped turn recovery]\n" + detail}
```
Nobody-Zhang, line 56: **[correctness · medium] On the OpenAI Responses wire this positional note becomes a standing instruction.**

`_responses_input` (`openai4s/llm/messages.py:278`) hoists *every* `system` message into top-level `instructions`, wherever it sits. "The previous turn stopped because… Do not repeat the same tool calls" then applies to every later request in the session, and each further stop stacks another copy — a fresh unrelated request is framed as a continuation. The Anthropic/Gemini assemblers already mark mid-timeline system content as `[system] …` timeline text; the Responses assembler has no such branch. Not fixed in this pass (wire behaviour outside this diff) — worth a follow-up.

openai4s/llm/models.py
```diff
@@ -84,6 +84,14 @@ def to_dict(self) -> dict:
         }
 
 
+class StreamReadError(TransportError):
```
Nobody-Zhang, line 87: **[correctness · low] The retry-budget rewrap in `_retry_loop` drops this subclass.**

`transport.py:239` re-raises a plain `TransportError(...)` when the budget is exhausted. For an *uncommitted* stream timeout (retryable) that path loses `StreamTimeoutError`, so `llm_failure_code` returns `None`: generic connection-failure text, no recovery note, no Continue — the #174 symptom again on the uncommitted path. `type(err)(...)` there preserves the classification.

frontend/src/features/messages/failure.ts
```diff
@@ -0,0 +1,79 @@
...
+  box.textContent = bits.join(" ");
+  box.dataset.requestId = failure.request_id ? String(failure.request_id).slice(0, 96) : "";
+  box.dataset.failureCode = failure.code ? String(failure.code).slice(0, 64) : "";
+  if (failure.output_committed) box.dataset.committed = "1";
+  if (recoverable) {
+    const frame = currentId.value;
+    const button = el("button", "turn-continue", COPY[LANG].continue) as HTMLButtonElement;
+    button.type = "button";
+    button.onclick = async () => {
+      if (button.disabled || running.value || !frame || currentId.value !== frame) return;
```
Nobody-Zhang: **[correctness · medium] Stale Continue buttons stay enabled and fail silently.**

Every recoverable failure row keeps a clickable button. When the row is no longer last (a second timeout after continuing gives two buttons), when the frame changed, or when the composer holds a draft, the handler just returns — at most a focus change. To the user that is a broken button. Suggest removing/disabling the button once its row is not the last one, and a `hint(...)` explaining why a non-empty composer blocks Continue. Not applied in this pass because it needs a `dist/` rebuild.

frontend/src/features/messages/failure.ts
```diff
@@ -0,0 +1,79 @@
...
+    llm_rate_limited: "turn.failure.llmRateLimited",
+    llm_upstream_overloaded: "turn.failure.llmUpstreamOverloaded",
+  } as Record<string, string>)[String(code || "")];
+  return key ? t(key) : "";
+}
+
+export function failureMeta(failure: Failure): HTMLElement {
+  const box = el("div", "msg-failure-meta");
+  const bits: string[] = [];
+  const cause = failureCodeHint(failure.code);
```
Nobody-Zhang: **[simplification · low] The same explanation is rendered three times for one stop.**

The backend tail (`_friendly_error` / the `no_progress` `err_text`), this `COPY` sentence and the `failureHint` toast all say the same cause plus the same 'continue' advice, as three separately maintained bilingual strings. A `no_progress` stop shows two near-identical 'Stopped repeating actions…' sentences back to back in one bubble. Consider letting the meta row carry only the button + support id for recoverable codes.

frontend/src/features/messages/messages.css
```diff
@@ -75,3 +75,7 @@
     animation: spin 1s linear infinite;
   }
 }
+.msg-failure-meta .turn-continue {
```
Nobody-Zhang, line 78: **[conventions · low] The Continue control has no themed styling.**

The global rule is only `button{font:inherit;cursor:pointer}`, so this renders as a native UA button inside the themed failure row (light grey in dark theme), and `cursor: pointer` here duplicates the global rule. Reuse one of the existing workbench button classes.

openai4s/agent/engine.py
Nobody-Zhang: Fixed in 4e3186f2. `_anthropic_messages` and `_gemini_contents` now hold a mid-timeline system message while tool results are pending and emit it right after the flush, so the pair stays adjacent: `assistant[tool_use] → user[tool_result] → user["[system] …"] → user["Continue"]`. Fixed in the assemblers rather than in `engine._finish`, because the ledger rebuilds the same note at the same position and a permission-resolution note can land there too.

Tests: `tests/test_llm_system_placement.py` — `…never_splits_the_anthropic_tool_pair`, `…never_splits_the_gemini_call_pair`, `…trailing_system_note_after_tool_results_is_still_delivered`. All three fail against the pre-fix `messages.py`.

openai4s/llm/transport.py
Nobody-Zhang: Fixed in 4e3186f2. The handler error is now typed where it happens: `dispatch()` wraps `on_event(event)` and raises a plain `TransportError("LLM event handler failed: <ExcType>", retryable=False, output_committed=True)`, so only a failed *read* reaches the `StreamReadError` / `StreamTimeoutError` branch. I went with this instead of the `isinstance(OSError, HTTPException)` filter suggested above: a `BrokenPipeError` raised by the sink is an `OSError` too and would still have been misclassified. The message carries the exception type only, not its text.

Test: `test_sse_event_handler_failure_is_not_an_upstream_interruption` (fails pre-fix).

openai4s/agent/recovery.py
Nobody-Zhang: Fixed in 0fbc0583. `_responses_input` now follows the rule the other two assemblers already had: only *leading* system messages become `instructions`; a later one is emitted as a `[system] …` user item at its position, and is held while a tool batch is still being answered so it never sits between a `function_call` and its `function_call_output`. This also takes compaction handoff notes out of standing instructions on this wire.

Tests: `test_the_responses_wire_keeps_a_mid_timeline_note_out_of_instructions`, `test_the_responses_wire_does_not_put_a_note_inside_a_tool_batch`.

openai4s/llm/models.py
Nobody-Zhang: Fixed in 4e3186f2: the budget-exhausted rewrap is now `raise type(err)(…)`, so a `StreamTimeoutError` keeps its class and `llm_failure_code` still returns `llm_stream_timeout` on the uncommitted path.

Test: `test_sse_timeout_keeps_its_code_when_the_retry_budget_runs_out` (fails pre-fix).

frontend/src/features/messages/failure.ts
Nobody-Zhang: Fixed in 0fbc0583.

- Stale rows: `messages.css` hides the button once a later row exists (`.msg:has(~ .msg) .turn-continue`), which covers live, reopened and older-page rendering without a hook in each path. The click handler keeps its own check and now *removes* the button when it finds itself stale, in case the rule did not apply.
- Draft in the composer: Continue still refuses, but now says why via `hint(…)` (new `draft` copy, en/zh) instead of only moving focus.

`stopped.test.ts` asserts both; `dist/` rebuilt (the pre-change build reproduced the committed bundle byte-for-byte, so the dist diff is this change only). Checked in a browser against the real `style.css` + built CSS: old row `display:none`, newest row visible. A click-through against a live daemon is still unverified — the access-token gate needs a human login.

frontend/src/features/messages/failure.ts
Nobody-Zhang: Fixed in 0fbc0583 for the in-bubble duplication: for recoverable codes `failureMeta` no longer adds the cause sentence, since the row's own text (the gateway tail, live and on reopen) already states it — the meta row carries the support id and the button. `failureCodeHint` is unchanged, so the composer hint still names the cause, as it does for every other failure code. `stopped.test.ts` asserts the meta row does not repeat it.

frontend/src/features/messages/messages.css
Nobody-Zhang: Fixed in 0fbc0583: the button now uses the existing `outline-btn small` classes (same as the history-retry control) and the duplicate `cursor` is gone. One thing the restyle exposed: `style.css` dims the whole `.msg-failure-meta` row to `opacity:.72`, which made the button read as disabled, so a meta row that contains the button is restored to `opacity:1` with `color: var(--text-400)` for the text. `scripts/check_css_tokens.py` passes.

